PR #238, opened 2015-01-17
Fix examples using deprecated method

Sitebase:
Replace deprecated createClient method in examples with new connect method.


mcollina:
Thank so much!

PR #239, opened 2015-01-18
Don't load "crypto" just for "randomBytes"

ForbesLindesay:
This is fine in node, but in the browser it loads the enormous crypto dependency when all that is needed is a random ID.  The `crypto.randomBytes` function is also not supported on older browsers, so it would seem more sensible to use the simpler `Math.random()` unless true cryptographic security is required for this ID.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2,7 +2,6 @@
  * Module dependencies
  */
 var events        = require('events')
-  , crypto        = require('crypto')
   , Store         = require('./store')
   , eos           = require('end-of-stream')
   , mqttPacket    = require('mqtt-packet')
@@ -26,7 +25,7 @@ var defaultConnectOptions = {
 };
 
 var defaultId = function() {
-  return 'mqttjs_' + crypto.randomBytes(8).toString('hex');
+  return 'mqttjs_' + Math.random().toString(16).substr(2, 8);
 };
 
 function nop() {}


mcollina:
Wooow, thanks! This should be a huge space-saving thing for browsers.
Are you using MQTT in production somewhere?

ForbesLindesay:
No, but actively considering it.  He other big (but much more work) saving would be to omit the requirement for Buffer to be loaded.  Unfortunately that would be much more difficult.  I've been working on a project to try and do data flow analysis to figure out which methods of Buffer are actually used and then remove all the dead code, which may well do the trick, but it's a tall order and I haven't actually got very far.

The other thing I would need to actually support this in production would be a fallback to long-polling (which I'm guessing would also require some additional work on the server)

mcollina:
I think you can add long-polling in by leveraging streams, i.e. this client can run on _any_ node binary stream. I think something like [primus](http://npm.im/primus) can do the trick, or basically anything than gives you a node binary stream. You just have to call the [MqttClient](https://github.com/mqttjs/MQTT.js#client) directly, passing a function that returns a stream (for automatic reconnect). As for the server, what are you using? If it's custom-built, you can just use [mqtt-connection](http://npm.im/mqtt-connection) on top of anything, or if it is [Mosca](http://npm.im/mosca) you might require some more refactoring (that I'll be happy to merge), but I think it will be worth it.

The way this library is built is to leverage node `Buffer` API. [mqtt-packet](http://npm.im/mqtt-packet), also uses `Buffer` (and [bl](http://npm.im/bl)) for parsing. So, removing `Buffer` is out of scope. However, you can use MQTT.js on node, and [paho.js](https://eclipse.org/paho/clients/js/) in the Browser. That is built around the standard websocket API (good), and array buffers (good), but it is less pluggable (you cannot plug in a custom transport stream).

PR #240, opened 2015-01-22
Fixed bugs caused by parsing the protocol from the URL

256dpi:
- Nodes `url.parse` return `mqtts:` instead of the expected `mqtts`.
- The `wss` protocol configuration was missing.
- Added tests.


mcollina:
Thanks a million!

PR #241, opened 2015-01-22
`tls.connect` first expects the host property

256dpi:
@mcollina Found another one.

Maybe it would be better to have real servers to test against, kinda like an acceptance test?

I'm working on a `mqtt-server` module that builds upon the new mqtt-connection. Wanna release that one under the mqttjs organization, and use it for tests?

### DIFF
--- a/lib/connect/tls.js
+++ b/lib/connect/tls.js
@@ -3,7 +3,7 @@ var tls = require("tls")
 
 function buildBuilder(mqttClient, opts) {
   opts.port     = opts.port || 8883;
-  opts.hostname = opts.hostname || opts.host || 'localhost';
+  opts.host     = opts.hostname || opts.host || 'localhost';
 
   opts.rejectUnauthorized = !(opts.rejectUnauthorized === false);
 


256dpi:
I created a repository here: https://github.com/256dpi/mqtt-server

mcollina:
I think we should move the whole connect argument parsing to the individual client test files, these already include real servers. Would you like to submit a PR on that?

How do you feel in helping maintaining this project? Check out [CONTRIBUTING.md](https://github.com/mqttjs/MQTT.js/blob/master/CONTRIBUTING.md) and let me know if you agree with it.

Regarding using a full-featured broker for tests, I don't think it is a good idea. It would be a chicken-egg problem, which one is right? Regarding having a 'mqtt-server' inside the mqttjs org, it would be a good idea, even to simplify the tests here. However I would split it in two modules, a [create-stream-server](https://www.npmjs.com/create-stream-server) thing that just creates all the TCP, TLS and WS servers, and then the mqtt-server could just be built on its top, just by wrapping the callback and providing some alias for 'mqtt' and 'mqtts'.

I am working on a big refactoring of [mosca](http://npm.im/mosca), splitting out the core mqtt broker handling from all the rest (which is still private), which will be based directly upon [mqtt-packet](http://npm.im/mqtt-packet) to achieve more speed. However this is far from being completed, so nevermind :D.

256dpi:
Would be cool to help contributing as I'm dependent on the libraries in my side project.

The 'create-stream-server' module is good idea. Maybe wen can transfer the mqtt-server repo to mqttjs and work with PRs from there. I have also other ideas in my head, which I would like to discuss.

mcollina:
Added you to the repo, add yourself in the README! :)

I have created the two repos under the main org. Have fun! :dancers: 

Ping me when you have something to review!

256dpi:
Cool thanks! I'll let you know!

PR #246, opened 2015-02-04
Use port for protocol when none is provided

### LINKED ISSUE #245: Websockets in browser should use default ports for http and https
I noticed that when the port isn't specified in the URL being passed to the [`buildBuilderBrowser`](https://github.com/mqttjs/MQTT.js/blob/master/lib/connect/ws.js#L27) function, it's defaulting to using whatever port the current page is being loaded with. Wouldn't it be more intuitive to use whatever the default port for `ws` and `wss` be? (I'm assuming that would be the same as `http` and `https`, so 80 and 443)

RangerMauve:
Fixes #245

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -29,13 +29,7 @@ function buildBuilderBrowser(mqttClient, opts) {
   }
 
   if (!opts.port) {
-    if(opts.protocol === 'wss'){
-      opts.port = 443;
-    } else if(opts.protocol === 'ws') {
-      opts.port = 80;
-    } else {
-      opts.port = parsed.port;
-    }
+    opts.port = parsed.port;
   }
 
   var host = opts.hostname || opts.host
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -26,10 +26,17 @@ function buildBuilderBrowser(mqttClient, opts) {
 
   if (!opts.host) {
     opts.host = parsed.hostname;
+    if (!opts.port) {
+      opts.port = parsed.port;
+    }
   }
 
   if (!opts.port) {
-    opts.port = parsed.port;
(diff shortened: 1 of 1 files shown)


RangerMauve:
Not really sure why this failed. Travis CI seems to just be saying

```
Error: shasum check failed for /home/travis/tmp/npm-2022/1423012627457-0.11922413692809641/tmp.tgz
```

mcollina:
Thanks for contributing!!

I think we are missing some functionality here, let's recap:

1. if nothing is provided, it should use the host/port combination of the current page
2. if only a hostname is provided, it should use that hostname, and the port based on the protocol
3. if both a hostname and port are provided, it should use those

I think that the current code fails on one, because line https://github.com/RangerMauve/MQTT.js/blob/patch-2/lib/connect/ws.js#L37 can never be reached, as protocol is always set (see https://github.com/RangerMauve/MQTT.js/blob/patch-2/lib/connect/ws.js#L19-L25).

Can you try adding a unit test for this? The in-browser tests live in https://github.com/mqttjs/MQTT.js/tree/master/test/browser, and you can run them with `npm run browser-test`.

RangerMauve:
Good point. I'll look into it further later tonight.

RangerMauve:
I'm running the tests now, and it's failing on `MqttClient :: specifying nothing :: should connect` because it's trying to connect to `http://localhost:61017/` and I assume that it's because there is no MQTT broker set up on that port.

Any idea on what I should do? Change the test?

mcollina:
Try starting a MQTT broker using https://github.com/defunctzombie/zuul/wiki/zuul.yml#server-optional
It should support websockets!

mcollina:
Ahum, it passes the current sets of tests, and I do not see how that can be tested anyway. Merging.

PR #250, opened 2015-02-12
added switch to secure protocol when crt and key are present

itavy:
Signed-off-by: Octavian Ionescu <itavyg@gmail.com>


mcollina:
This is really good! Thank so much! How are you using MQTT.js? We are always looking for new committers who wants to help maintaining this :)

PR #254, opened 2015-02-26
Handle paths properly

lericson:
This became an issue with for example `fubar` which does not allow WebSocket connections on any path but `/mqtt`. :stuck_out_tongue_winking_eye:


mcollina:
Thanks!

PR #255, opened 2015-02-26
fix and test for issue 252

mdickens:
Test seems to be effective at catching problem without fix, but did not see offline event as expected to fail the test. Instead when it fails the test just times out.


mdickens:
Hi,

I see the build failed, Its all passing when i run it, please let me know what next steps are.


  225 passing (24s)
  2 pending

Issue #252, opened 2015-02-15
client goes offline on publish on mac


meilon:
Hi,
I'm developing an app on my MacBook Pro and I have a problem with MQTT.js 1.0.10 on it. The client should receive a message on one QoS 2 topic, do some stuff with it and publish it on another QoS 2 topic.

After the message reaches the client at first nothing happens on the console. Then, after the keepalive timeout occurs, the client reconnects and publishes the message. The general error event doesn't get called and the err object in the publish callback doesn't contain anything.

This happens with mosquitto 1.3.5, iot.eclipse.org and test.mosquitto.org without TLS on my MacBook Pro with current OSX 10.10.2 and node 0.12.0. Because the same script works on a Raspberry Pi with note 0.10.26 I'm guessing it's an issue with my Mac or it's generally Mac reladed. Node is installed via brew

Here the code I'm testing with:
```
var mqtt = require('mqtt');

var mqttOptions = {
    clientId: 'delayed-demo', 
    clean: false, 
    keepalive: 10
}

mqttClient = mqtt.connect('mqtt://test.mosquitto.org', mqttOptions);
mqttClient.on('connect', function() {
	console.log(new Date().toTimeString() + ' [MQTT] connected!');
	mqttClient.subscribe('/poke', { qos: 2}, function(err, granted) {
		if (err) {
			console.log(new Date().toTimeString() + ' [MQTT] ERROR: ' + err);
		} else {
			console.log(new Date().toTimeString() + ' [MQTT] Subscribed to topic ' + granted[0].topic + ' with QoS ' + granted[0].qos);
		}
	});
});

mqttClient.on('offline', function() {
    console.log(new Date().toTimeString() + ' [MQTT] now offline!');
});

mqttClient.on('error', function(error) {
    console.error(new Date().toTimeString() + ' [MQTT] general error: ' + error);
});


mqttClient.on('message', function (topic, message) {
	console.log(new Date().toTimeString() + ' [MQTT] Topic: ' + topic + ', message: ' + message);
    
    mqttClient.publish("/ping", "pong!", {qos:2, retain:false}, function(err) {
        if (err) console.log(new Date().toTimeString() + ' [MQTT] publish error: ' + err);
        console.log(new Date().toTimeString() + ' [MQTT] Published alarm to "/ping"');
    });
});
```
And here's the console output:
```
19:23:50 GMT+0100 (CET) [MQTT] connected!
[...]

PR #256, opened 2015-02-26
fix and test for issue 252 (retry)

mdickens:
2nd try first request build failed... added one last change . Test seems to be effective at catching problem without fix, but did not see offline event as expected to fail the test. Instead when it fails the test just times out.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -752,7 +752,7 @@ MqttClient.prototype._handleAck = function(packet) {
  * @api private
  */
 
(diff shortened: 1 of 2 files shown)


meilon:
I just tested it by manually changing the 3 lines in the client.js by hand, it worked in my project. This should definitely fix issue #252. Haven't tested the test, though.

mcollina:
Thank so much! A new release is coming :)

PR #261, opened 2015-03-11
Created an abstract test for the Store.

mcollina:
This is needed to test https://github.com/mcollina/mqtt-level-store.

Also some updates in the README.

Any opinion against this?


itavy:
hi,

it seems ok. as a side comment i've seen that in the test there is a mix between synchronous and asynchronous tests. is there a reason for this mix or shall we stick to only one type of tests ?

mcollina:
They all seem async to me, can you please point me to the sync ones?

itavy:
for this pull yes, but there are tests which are not async:

in test/client.js
line 107 
line 115

in test/mqtt.js
all tests except last 3

mcollina:
They seem all legit to me. If you would prefer to change them, submit a PR and we will discuss there.

Merging this one, thanks for reviewing @itavy!

PR #262, opened 2015-03-12
Changed all the tests to async

itavy:


256dpi:
Thanks for the work, but can you elaborate what the benefit of that approach is?

itavy:
hi,

i think that all tests should use same pattern (either sync or async) ;
since it is an evented, some things can be tested only in async

256dpi:
IMHO it is unnecessary. @mcollina What do you think?

mcollina:
I think it is necessary, but not how it is implemented.

`mqtt.connect()` creates a client that continuously try to connect at those address. The right approach is to actually make sure either:

1. the `'client'` event in the server is emitted, and the client closes
2. it never connects to anything (a small timeout is ok), i.e. the `'connect'` event in the client is never emitted.

After each of those cases the client needs to be closed properly.
Both these checks are asynchronous, and in the current code causes spurious failures.

timjacobi:
I don't think it is a good idea to use a callback with test code that is not async. This takes away semantics i.e. when someone sees the *done* callback in a test they tend to assume that the test contains async logic.

mcollina:
I don't think we reached an agreement on this. Just closing for now.

PR #272, opened 2015-04-22
outgoingStore reading stream modified

erwinkendo:
Add proper management of stored messages to support persistent Store objects (e.g. mqtt-level-store
). The stream reading mechanism was changed to avoid overhead when reading stream. A reading condition was added to avoid reading all the stream in case a reconnection is taking place but wasn't successful.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -219,10 +219,23 @@ MqttClient.prototype._setupStream = function () {
   // Echo connection errors
   parser.on('error', this.emit.bind(this, 'error'));
 
-  this.outgoingStore
+  var out = this.outgoingStore
     .createStream()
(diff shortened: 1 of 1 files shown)


mcollina:
:+1: for me, but you need to get travis happy, the code style validations are failing :).

erwinkendo:
:fearful: You are right, I have modified it a little so Travis CLI doesn't produce any error (jshint doesn't work very well in windows with the current " /* " and " /*.js " syntaxes BTW).
The time results for some Mocha test are a little bit higher that with the previous implementation (~50ms max) in some, I guess this is due to the ```once.('readable')``` event that the rest of the storage is read only after a msg callback is issued.

PR #273, opened 2015-04-23
outgoingStore reading stream modified (fix)

erwinkendo:
Add proper management of stored messages to support persistent Store objects.
Changed stream reading mechanism to avoid overhead when reading stream.
Added reading condition to avoid reading all the stream in case a reconnection is taking place and is not done.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -229,28 +229,22 @@ MqttClient.prototype._setupStream = function () {
         if (!packet) {
           return;
         }
-        // Avoid unnecesary stream read operations
-        if (!that.disconnecting && !that.reconnectTimer) {
+        // Avoid unnecesary stream read operations when disconnected
+        if (!that.disconnecting && !that.reconnectTimer && (0 < that.options.reconnectPeriod)) {
           outStore.read(0);
           // Ensure that the next message will only be read after callback is issued
           that.outgoing[packet.messageId] = function () {
             storeDeliver();
           };
           that._sendPacket(packet);
+        } else {
+          outStore.close();
         }
       }
       storeDeliver();
     })
     .on('error', this.emit.bind(this, 'error'));
 
-
-  // pump(outStore);
-  // this.outgoingStore
-  //   .createStream()
-  //   .on('data', function (packet) {
-  //     that._sendPacket(packet);
-  //   })
-  //   .on('error', this.emit.bind(this, 'error'));
 };
 
 MqttClient.prototype._handlePacket = function (packet, done) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -175,6 +175,7 @@ MqttClient.prototype._setupStream = function () {
     writable = new Writable(),
     parser = mqttPacket.parser(this.options),
     completeParse = null,
+    outStore = null,
     packets = [];
 
   this._clearReconnect();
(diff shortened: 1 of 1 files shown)


erwinkendo:
:astonished: @mcollina sorry! I will fix this fix pull request, although I don't understand why my change affects the Websocket Client pinging, and only in node 0.10

mcollina:
It seems a spurious failure on Travis. Can you please remove that blank commit?

erwinkendo:
:open_mouth: Ok done, commit removed. :smile:

lib/client.js
```diff
@@ -219,12 +220,37 @@ MqttClient.prototype._setupStream = function () {
...
+        }
+        // Avoid unnecesary stream read operations
+        if (!that.disconnecting && !that.reconnectTimer) {
+          outStore.read(0);
+          // Ensure that the next message will only be read after callback is issued
+          that.outgoing[packet.messageId] = function () {
+            storeDeliver();
+          };
+          that._sendPacket(packet);
+        }
```
mcollina: this does not close the stream if the client gets disconnected, it might leak memory.

lib/client.js
```diff
@@ -219,12 +220,37 @@ MqttClient.prototype._setupStream = function () {
...
     .on('error', this.emit.bind(this, 'error'));
+
+
+  // pump(outStore);
+  // this.outgoingStore
+  //   .createStream()
+  //   .on('data', function (packet) {
+  //     that._sendPacket(packet);
+  //   })
+  //   .on('error', this.emit.bind(this, 'error'));
```
mcollina: can you please remove this code?

mcollina:
Would you like to add a unit test for this?

lib/client.js
erwinkendo: Good point, maybe bind

lib/client.js
erwinkendo: Sorry, my bad. I will.

erwinkendo:
A unit test sounds good, I think the best way would be to test it against level-store package in different scenarios like issuing disconnect then reconnect, boot with stored messages and so on. What do you have in mind?

mcollina:
I think you can reproduce all the possible scenarios with https://github.com/mqttjs/MQTT.js/blob/master/lib/store.js, without calling the `close` method.
Yes, several connecting/disconnecting scenarios would be good!

lib/client.js
erwinkendo: I am currently using ```(!that.disconnecting && !that.reconnectTimer && (0 < that.options.reconnectPeriod))```to check if the client disconnected but sometimes (unexpectedly) this condition becomes true (presumably ```!that.reconnectTimer``` fault) even when the client is not connected. Checking the ```this.connected```but this flag is not ```true```when the ```_setupStream``` is created when the client is first created. 
I will continue to use this condition because it reduces the Store reading calls, but any other suggestions are welcomed.

mcollina: This is perfect. However I noticed that the stream does not get destroyed properly if a disconnect happens while we are still sending the messages. My bad, I did not put in this feature in the first place. Would you mind adding it while you are at it?

erwinkendo: Sure, currently I am issuing ```outStore.close();``` in the else condition inside ```storeDeliver ```.
Would this be enough?

mcollina:
Ping, any update on this one?

erwinkendo:
Sure, I haven't finished the unit test but the changes to the code are finished. I wrote you 10 days ago about the way I am closing the stream in case a disconnection takes place, once I know your opinion I will update the pull request :smile:

mcollina:
Hi @erwinkendo I completely lost your question :/, can you please repeat it?

erwinkendo:
Sure @mcollina , I was wondering if the way I am closing the stream is ok.
I am issuing outStore.close(); in the else condition inside storeDeliver.
Would this be enough?

lib/client.js
mcollina: yes it is :D.

mcollina:
Thanks, that's perfect, I'm merging!!

PR #274, opened 2015-04-26
Updated Client.publish API description

BrainCrumbz:
Updated API description according to current rev of https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js
NOTE: API description on npm package page is also outdated.


mcollina:
Thanks! This will be updated on NPM when we publish a new version!

PR #286, opened 2015-05-19
Setup the ping timer after a reconnect

mcollina:
The problem lies in the fact that the `'connect'` event is not emitted anymore at every reconnection, as it was in MQTT.js < 1.0.0. We should set up the ping timer even in the reconnect case, which is just what this PR does.

Please review, it should be straightforward anyway.


mcollina:
Never mind, the `'connect'` event is in fact emitted.

PR #289, opened 2015-05-25
pass the option 'rejectUnauthorized' for self-sign certificate when using wss://URL

ozboy:
pass the option 'rejectUnauthorized' for self-sign certificate when using wss://URL

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -7,8 +7,8 @@ function buildBuilder (client, opts) {
   var wsOpt = {};
   wsOpt['protocol'] = 'mqttv3.1';
   if(opts.protocol == 'wss')
-      if(opts.hasOwnProperty('rejectUnauthorized'))
-          wsOpt['rejectUnauthorized'] = opts.rejectUnauthorized;
+    if(opts.hasOwnProperty('rejectUnauthorized'))
+      wsOpt['rejectUnauthorized'] = opts.rejectUnauthorized;
   var host = opts.hostname || 'localhost',
     port = String(opts.port || 80),
     path = opts.path || '/',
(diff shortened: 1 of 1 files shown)


lib/connect/ws.js
```diff
@@ -4,13 +4,16 @@ var websocket = require('websocket-stream'),
   _URL = require('url');
 
 function buildBuilder (client, opts) {
+  var wsOpt = {};
+  wsOpt['protocol'] = 'mqttv3.1';
+  if(opts.protocol == 'wss')
+      if(opts.hasOwnProperty('rejectUnauthorized'))
+          wsOpt['rejectUnauthorized'] = opts.rejectUnauthorized;
```
mcollina: I fear this might cause some issue in the browser. Can you please check?
Otherwise I would prefer to add it only if it is not a browser.

ozboy:
I have used browserify and test on IE, Firefox and Chrome.

```javascript
<html>
<head lang="en">
    <meta charset="UTF-8">
    <title>test mqtt wss with broker self-sign certificate</title>
</head>
<body>
<script src="./browserMqtt.js"></script>
<script>
    //var client = mqtt.connect('wss://waoz.cloudapp.net:443/Mosca/');
    var client = mqtt.connect('wss://waoz.cloudapp.net:443/Mosca/', {"rejectUnauthorized": false});
    client.on('connect', function ()
    {
        console.log('connect');
        client.subscribe('mqtt/test');
        client.publish('mqtt/test', 'hello browser!');
    });
    client.on("message", function(topic, payload) {
        alert([topic, payload].join(": "));
        client.end();
    });
</script>
</body>
</html>
```

There is no difference with original rule on browser because I did not modify the function *buildBuilderBrowser* in lib/connect/ws.js.
It still need to be trusted self-sign certificate manually on browser before connecting to the broker successfully.

mcollina:
Thanks!

PR #296, opened 2015-06-10
Clear connackTimer before setting a new one

### LINKED ISSUE #295: Reconnect timer clears even though it hasn't reconnected.
I've been testing out the reconnection logic, and it seems that it'll try to reconnect every second, and then it will clear stop trying to reconnect almost exactly 30 seconds later, after exactly 30 attempts. To reproduce this, I just unplug the ethernet cable from a machine and watch the console.

I've got a keepalive set to 22 seconds if that matters.

Taking a look at the code flow the following appears to occur: 
 1. Socket disconnects
 1. stream emits `close`
 1. Reconnect timer is started via `#_setupReconnect()`
 1. `#_reconnect()` is called
 1. `#_setupStream()` is called
 1. `reconnectTimer` gets cleared and a new stream is set up
 1. Stream gets created again
 1. stream emits `close`
 1. repeat

For some reason, it appears that step 8 isn't happening every now and then, and I'm not really sure why.

Another comment is that using `setInterval` for the reconnect timer doesn't make sense if you're going to be clearing it immediately anyway.

*Edit: I'm using the a websocket connection in a Chrome browser*

RangerMauve:
Fixes #295

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -247,8 +247,8 @@ MqttClient.prototype._setupStream = function () {
       storeDeliver();
     })
     .on('error', this.emit.bind(this, 'error'));
-  
-  clearInterval(this.connackTimer);
+
+  clearTimeout(this.connackTimer);
   this.connackTimer = setTimeout(function () {
     that._cleanUp(true);
   }, this.options.connectTimeout);
--- a/lib/client.js
+++ b/lib/client.js
@@ -247,7 +247,8 @@ MqttClient.prototype._setupStream = function () {
       storeDeliver();
     })
     .on('error', this.emit.bind(this, 'error'));
-
+  
+  clearInterval(this.connackTimer);
   this.connackTimer = setTimeout(function () {
     that._cleanUp(true);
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks so much!! Can you please add a unit test for this?
Il giorno mer 10 giu 2015 alle 21:33 RangerMauve <notifications@github.com>
ha scritto:

RangerMauve:
Honestly, I'm not really sure how I would approach making a unit test for something like this. Would just making sure that the previous `connackTimer` was changed upon making a new stream be sufficient?

RangerMauve:
Or should I add something in [the reconnecting tests](https://github.com/RangerMauve/MQTT.js/blob/patch-3/test/client.js#L126)?

mcollina:
Can you please fix the linting errors? https://travis-ci.org/mqttjs/MQTT.js/jobs/66288162

RangerMauve:
Of course. Lemme just get to a computer first. :P

PR #301, opened 2015-06-17
bad behavior when offline

chiarishow:
Hi,

after watching Matteo in many conferences about mqtt I decided to give it a try but I found a strange behavior when the client is offline.

So that's the case:
1) the client is offline and every second it tries to reconnect.
2) some messages are stored in the outgoingStore with qos 1 (no problem with qos 0 or 2).
3) after 30 minutes the client comes online again.
4) the first message in the outgoingStore is sent 1800 times (the other messages 1 time only). That happens because every time the client tries to reconnect it pushes the first message in the queue. 

So my solution is to fill the queue only when the client is actually connected.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -111,7 +111,7 @@ function MqttClient (streamBuilder, options) {
   // Mark connected on connect
   this.on('connect', function () {
     this.connected = true;
-    var outStore = null,
+    var outStore = null;
     outStore = this.outgoingStore.createStream();
 
     // Control of stored messages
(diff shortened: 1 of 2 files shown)


mcollina:
This is really really good! Thanks for fixing it. Can you please add a unit test, so that we don't hit this bug again?

chiarishow:
Ok, unit test added.

Btw I noticed that the client must successfully connect and send a qos 1 message in order to cause the problem. But hopefully now is fixed. 

Thank you.

PR #326, opened 2015-10-01
Use svg version of travis badge

eiriksm:
Hey thanks for making this available.

Just noticed you are using the png version of the build status badge, while travis actually offers a svg version. On retina-style screens this looks a lot better with svg :)

So here is a very minor PR changing the badge to the svg version. If you think it is a good idea, feel free. If not, no biggie.

Thanks again, have a nice day!


mcollina:
Thanks! I didn't even know about it!

Issue #317, opened 2015-09-09
CLI --key & --cert Fails


philmaker1:
This following lines of code in the project appear to be referencing a value which does not yet exist.

https://github.com/mqttjs/MQTT.js/blob/master/bin/pub.js#L52
https://github.com/mqttjs/MQTT.js/blob/master/bin/pub.js#L56
https://github.com/mqttjs/MQTT.js/blob/master/bin/sub.js#L52
https://github.com/mqttjs/MQTT.js/blob/master/bin/sub.js#L56

Philip

mcollina:
Hi @philmaker1! Would you mind sending a PR to fix them?

Thanks!

mcollina:
Fixed long ago.

PR #327, opened 2015-10-02
Allow key, cert, ca to be passed into cli. Don't break if no message is specified during pub.

janjongboom:
Fix for #317

### DIFF
--- a/bin/pub.js
+++ b/bin/pub.js
@@ -19,7 +19,7 @@ function send(args) {
 
 function start(args) {
   args = minimist(args, {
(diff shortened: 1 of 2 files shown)


mcollina:
Can you please also add these to subscribe?
And also update the docs https://github.com/mqttjs/MQTT.js/tree/master/doc.

janjongboom:
@mcollina Yeah, they are added as well for subscribe.

The key, cert and ca options are already described in docs.

mcollina:
Oh my bad :). Thanks for sending this PR! :)

PR #328, opened 2015-10-02
Prevent unecessary 'drain' listeners on the write stream when buffering publishes in qos 0 with no callback

platy:
Moving the default of the publish callback to nop. The defaulting is still done for the qos 1/2 callbacks, but under qos 0, the callback is left undefined so as to not add nop event handlers to the 'drain' event on the stream when the stream is buffering.

I don't believe this effects any situation except for when:
- publishing with no callback in qos 0
- when the stream is not ready to write to
in this situation it doesn't add a 'drain' listener to the write stream

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -330,8 +330,6 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
     return this;
   }
 
-  callback = callback || nop;
-
   packet = {
     cmd: 'publish',
     topic: topic,
@@ -344,8 +342,9 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
   switch (opts.qos) {
     case 1:
     case 2:
+
       // Add to callbacks
-      this.outgoing[packet.messageId] = callback;
+      this.outgoing[packet.messageId] = callback || nop;
       this._sendPacket(packet);
       break;
     default:


platy:
I saw this when using the mqtt client under a higher load (with a lot of messages being sent in pulses) and noticed this EventEmitter warning:

````
(node) warning: possible EventEmitter memory leak detected. 11 drain listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at TLSSocket.addListener (events.js:179:15)
    at TLSSocket.Readable.on (_stream_readable.js:671:33)
    at TLSSocket.once (events.js:204:8)
    at sendPacket (/usr/share/vertex/node_modules/mqtt/lib/client.js:33:21)
    at MqttClient._sendPacket (/usr/share/vertex/node_modules/mqtt/lib/client.js:601:7)
    at MqttClient.publish (/usr/share/vertex/node_modules/mqtt/lib/client.js:352:12)
````

I understand that it means that the client is trying to send packets faster than the socket can write them, and so maybe I need to increase the limit because I have many packets to send and sometimes the socket will be slow. But I see that that is probably not necessary as I don't need a callback.

mcollina:
Good point! Which node version are you using?
I usually use that callback for QoS 0 for that purpose - to avoid writing too fast.
See https://github.com/mqttjs/MQTT.js/blob/master/benchmarks/bombing.js.

platy:
I'm pretty sure that I wasn't sending the publishes on the same tick as the publishes are each triggered by other events. The problem came up whilst sharing an LTE connection with a lot of other traffic. So I figured that the buffering is inevitable in slow periods and that the stream buffer is the best place for the packets to back up as I have no way to push back.

The Node version on that device was 0.12(.7?)

mcollina:
I think a complete fix should be to increase the max number of listeners in the stream as well, something like 1000 or so, as the same problem would happen for QoS 1 and 2.
Would you mind sending a PR for that as well?

platy:
Sure, that sounds pretty straight forward.

Thanks for the quick replies!

PR #329, opened 2015-10-02
Increase max listeners on stream to 1000

platy:
10 listeners is quite easily not enough on an intermittent connection when each publish callback requires a drain listener.


mcollina:
Thanks!

PR #330, opened 2015-10-06
Simplify rejectUnauthorized defaulting by using `!== false`

platy:
I believe this is much easier to read and it complies with the check-style rules. It's important that a reader can understand this line and that the default setting is true.


mcollina:
Thanks!! :)

PR #334, opened 2015-10-16
Session Present flag is needed while handling connect event

kokeksibir:
After making connection client needs to know if there exists its stored session in server side. Flag is avaliable in packet, however it should be passed to event as argument. For future needs argument type is object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,9 @@ version 1.3 and 1.4 works fine without those.
 `function(connack) {}`
 
 Emitted on successful (re)connection (i.e. connack rc=0). 
-* `connack` received connack packet 
+* `connack` received connack packet. When `clean` connection option is `false` and server has a previous session 
+for `clientId` connection option, then `connack.sessionPresent` flag is `true`. When that is the case, 
+you may rely on stored session and prefer not to send subscribe commands for the client.
 
 #### Event `'reconnect'`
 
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -667,7 +667,7 @@ MqttClient.prototype._handleConnack = function (packet) {
   clearTimeout(this.connackTimer);
 
   if (0 === rc) {
-    this.emit('connect');
+    this.emit('connect', null, {sessionPresent: packet.sessionPresent});
```
mcollina: I would do:

```js
this.emit('connect', packet);
```

so we avoid a useless object allocation.

mcollina:
Thanks for the PR! I've added a small comment.

Would you mind:

1. Adding a unit test for this
2. Updating the README

kokeksibir:
Ok, I think it's done now.

README.md
```diff
@@ -212,9 +212,10 @@ version 1.3 and 1.4 works fine without those.
 
 #### Event `'connect'`
 
-`function() {}`
+`function(connack) {}`
 
-Emitted on successful (re)connection (i.e. connack rc=0).
+Emitted on successful (re)connection (i.e. connack rc=0). 
+* `connack` received connack packet
```
mcollina: You might possibly add the meaning of sessionPresent here, so it's more clear why this is important.

PR #338, opened 2015-10-26
Fixes TLS Client example code.

### LINKED ISSUE #249: error in examples
hi,

1.in the file "/examples/tls client/mqttclient.js" the options for key and certificate are "keyPath" and "certPath" but in the file "/lib/connect/tls.js" when you build the secure socket, the tls constructor will look for "key" and "cert" keys to locate the files.

as is stated in the example:
var KEY = __dirname + '/tls-key.pem';
var CERT = __dirname + '/tls-cert.pem';
var TRUSTED_CA_LIST = [__dirname + '/crt.ca.cg.pem'];

but you pass them directly to tls constructor, and it will throw error since tls constructor expects a buffer and not a string, it should be 

var KEY = fs.readFileSync(__dirname + '/tls-key.pem');
var CERT = fs.readFileSync(__dirname + '/tls-cert.pem');
var TRUSTED_CA_LIST = [fs.readFileSync(__dirname + '/crt.ca.cg.pem')];

2.in the mqtt constructor, if you provide the key and cert options it will not build a secure client unless you specify one of the secure protocols.

Best regards,
Octavian

jmeas:
Resolves #249

### DIFF
--- a/examples/tls client/mqttclient.js
+++ b/examples/tls client/mqttclient.js
@@ -14,9 +14,9 @@ The line to add on the file should be as follows:
 
 var mqtt = require('mqtt');
 var fs = require('fs');
-var KEY = __dirname + '/tls-key.pem';
-var CERT = __dirname + '/tls-cert.pem';
-var TRUSTED_CA_LIST = [__dirname + '/crt.ca.cg.pem'];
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks @jmeas! :)

PR #341, opened 2015-10-29
Adding this.options.queue as an option

sandro-k:
I have added the ability to supply the `queue` of packages that is used while the client is offline to the constructor options. The test shows how one could use this to create a stream of messages that will be published right after the connection is made. I am aware that I could use

```
mqtt.connect();
mqtt.publish();
```
to publish even if the connection is not established, but this way I can queue messages without having access to a `MqttClient` instance. 

This functionality is needed to store packages published by the `<mqtt-publish>` before the `<mqtt-connection>` element can know If it might get the `MqttClient` injected or create it itself.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -95,7 +95,7 @@ function MqttClient (streamBuilder, options) {
   // Are we disconnecting?
   this.disconnecting = false;
   // Packet queue
-  this.queue = [];
+  this.queue = this.options.queue || [];
   // Are we intentionally disconnecting?
   this.disconnecting = false;
   // connack timer
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -374,6 +374,36 @@ module.exports = function (server, config) {
       });
     });
 
+    it('should publish a message (offline) from queue', function (done) {
+      var nop,
+        topic = 'test',
+        payload = 'test',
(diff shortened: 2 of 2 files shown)


mcollina:
This is nice @sandro-k. While we are at it, we should refactor this to fix #343 as well.

What do you think?

sandro-k:
Sure, would the expected behavior be that the packet is added to the `queue` and not jump out of the `_sendPacket ` method?

mcollina:
Actually we should ditch `queue` alltogether, and get packets in the store.

sandro-k:
Ok, we can do that. But I kind of like the simplicity of `queue` just being an Array. Every environment is capable to just create an array and store `packets` within the the array without the need to create a different object. If we remove `queue`, the user has to be able to reach `Store` to create it before the user creates the `MqttClient`.

mcollina:
We can still accept a `queue` array in the constructor. For QoS 1+ publishes, we need to put them also in the store.

mcollina:
This has saw very little activity in the past while. Closing for now.

sandro-k:
The current implementation of `<mqtt-publish>` actually needs this change to work as expected. I might reopen this in the future.

Issue #344, opened 2015-11-03
Missing Packet publishing with QoS 2


xadh00m:
Hi,

I´ve made a NodeJS project with MQTT.js 1.5.0 which publishes to a topic 10 times per second to a local HiveMQ 2.3.1 for testing. I always get the following error after usually less then 100 messages published:

```
Error: missing packet
    at Store.get (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/lib/store.js:94:8)
    at MqttClient._handlePubrel (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/lib/client.js:833:22)
    at MqttClient._handlePacket (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/lib/client.js:271:12)
    at process (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/lib/client.js:218:12)
    at Writable.writable._write (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/lib/client.js:228:5)
    at doWrite (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:237:10)
    at writeOrBuffer (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:227:5)
    at Writable.write (/dev/projects/focus/debug-topic-publisher/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:194:11)
    at Socket.ondata (_stream_readable.js:528:20)
    at emitOne (events.js:77:13)
```
It seems that the packet has been removed from store already. Is this a known issue? Any hints on how I could approach/fix this error?

Regards,
Mirko

mcollina:
There are a couple of options in the MQTT spec regarding how messages are stored for QoS 2. Maybe the way it is implemented here is not working well with HiveMQ.

Have you checked with Mosquitto as well?

Probably the best way to fix this is to add some `console.log` where things are deleted: https://github.com/mqttjs/MQTT.js/blob/master/lib/store.js#L72

jdiamond:
I tried reproducing this using the [public HiveMQ test broker](http://www.hivemq.com/try-out/) with the following script:

```
#!/usr/bin/env node

var mqtt = require('mqtt');

var client = mqtt.connect('mqtt://broker.hivemq.com');

client.on('connect', function() {
    var count = 0;

    (function publish() {
[...]

Issue #343, opened 2015-10-31
Async queue


oroce:
Hey,

I have a strange setup, my raspberry pi connect to the server through mqtt, but connection sometimes gets dropped, that isn't very problematic since mqtt.js is capable of queueing up the messages.

Sometimes the connection can be back in 1 or 2 minutes but there's a change when there's no connection for 10 minutes but in the meantime watchdog comes in and reboot the device since there was no internet for 5 minutes. And here comes the problem: `mqtt.js` only queues the events in memory  but cannot persist those events to the disk.

Is there somebody who is facing with the same problem? Or do you have recommendation for this case?

First I thought `mqtt-level-store` is for this specific use case but then realized it's more for handling backpressure of messages on the client.
Then I did some experimenting here: https://github.com/oroce/mqttfun

Basically I implemented async queue option, where I needed to change two things:
- ability to pass custom `queue` to the constructor/connect method (there's already a PR for that: https://github.com/mqttjs/MQTT.js/pull/341)
- make `queue.shift` async: https://github.com/oroce/MQTT.js/compare/master...oroce:async-queue

And then I wrote 3 queue adapters:
- memory: https://github.com/oroce/mqttfun/blob/master/queues/memory.js (basically async `[].push` and `[].shift`)
- redis: https://github.com/oroce/mqttfun/blob/master/queues/redis.js
- leveldb: https://github.com/oroce/mqttfun/blob/master/queues/level.js

What do you think? Should I solve this outside of `mqtt.js`, or should be async queue a thing?

mcollina:
You are wrong about the Store API. Its purpose is to store both incoming and outgoing messages and confirmations.

For outgoing, see https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L46-L53.

oroce:
@mcollina: That's okay, my problem happens when `this.connection !== true`: https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L586-L588

mcollina:
If there is no connection, then the packet goes in the queue. There might be a bug in this code, but it was written to handle your use case.

oroce:
Oh no, that does everything as it is intended to do. My problem is that I want a queue which can survive the restart of the client (basically I'm restarting the device itself). For this I need a NON inmemory queue.

mcollina:
mqtt-level-store should be on disk.

oroce:
[...]

PR #345, opened 2015-11-06
v2 dev

### LINKED ISSUE #166: Restore subscribtions after reconnect in deferred open scenario
Hi,
   I am using MQTT(client ) and Mosquitto(broker). I have clients which will be subscribed to the broker always. If the the broker restarts or goes down the client is not able to re-connect or throw any error. Is this also related to #112 and #148. I am using mqtt 0.3.8. My client is created as
client = mqtt.createClient(PORT,MQTT_SERVER, { encoding : 'binary', username : MQTT_USERNAME, password : MQTT_PASSWORD, clientId : 'backend'});

### LINKED ISSUE #348: mqtt.js 1.5.0 CLI get stuck
This is just something I would like to point out.

Running this command, the mqtt.js CLI 'freezes' with no response:

     mqtt publish -h '192.168.1.12' -t 'anyone' -m 'hello world' -u 'megatron' -p 'ciao123'

Because mistakenly using the lower case -p instead of -P to specify the password (mqtt.js acts as 'ciao123' was the port number..).

I don't know if this is worth a 'fix', but can't be useful some kind of hint/error in this scenario?

### LINKED ISSUE #382: Default value of keepalive should be at least 1 minute
As title: any opinions?

### LINKED ISSUE #431: Connection is broken query
Hi,

I am looking at understanding the capability of mqtt.js a little more when a clients connection to a broker is broken.

I've read through the documentation and I am still a little unsure what happens to outgoing messages when the client fails to connect, are messages automatically queued as default or do I need to set something explicitly to enable this queuing. I can see that QoS: 0 is default, and therefore should queue QoS zero messages?

Is there a limit to what can be queued, is it stored in memory or written to disk?

Is there a way to log what has been written to the queue and when it is removed from the queue when the connection is re-established?

Thank you in advance!

Jon

mcollina:
This PR is up for discussion to release a new major version of MQTT.js.
Thanks to the update to mqtt-packet v4, it's 3x faster in sending packets.

* [x] Update to [mqtt-packet](http://npm.im/mqtt-packet) v4
* [x] Drop support for node v0.8 and v0.10.
* [x] Drop deprecated API, in particular `createClient`, `createServer` and so on
* [x] Use [readable-stream](http://npm.im/readable-stream) v2.x.x

Please review my changes :).

Fixes #382.
Fixes #351.
Fixes #431.
Fixes #348.
Fixes #166.

Standardized, hurray!


mcollina:
@jdiamond, would you like to have a look at #343, #344 and #166. These should be along the lines of what you have contributed recently.

mcollina:
@jdiamond would you mind testing/reviewing this?

ralphtheninja:
@mcollina When v2 is released, mind writing something about what has changed api wise?

wuhkuh:
These commits still utilize util.inherits which is deprecated and discouraged (see #458) - would a change be possible before this one gets pulled?

mcollina:
Please, send a PR against this branch that replaces them with the inherits module.

roccomuso:
@wuhkuh @mcollina but shouldn't be kept somehow the backwards compatibility with older node.js versions? those not ES6-compliant..

itavy:
that can be solved very easy with babel. transpile it to use at least 0.12.

mcollina:
@roccomuso what do you mean? We bump the dependency not because of ES6, but because of streams 3, and the complexities of keeping the codebase in shape for that. v1 is still there to use, and if we want to backport patches, we can! Would you like to take care of the v1.x.x of MQTT.js?

roccomuso:
@mcollina I'm referring to the usage of <code>util.inherits()</code> that is discouraged but required for older node.js version, that's why @itavy proposed the use of babel.

wuhkuh:
@roccomuso Aren't we dropping support for older node.js versions?

> Drop support for node v0.8 and v0.10.

mcollina:
@roccomuso we are not using any ES6 for inhering. Giving that `util.inherits` its discouraged, we switched to the npm module http://npm.im/inherits

roccomuso:
@mcollina uh great then. Now I realized what you meant with **replaces them with the inherits module.** ;)

PR #349, opened 2015-11-15
Port validation.

roccomuso:
Validate that the port is a number.

### DIFF
--- a/bin/pub.js
+++ b/bin/pub.js
@@ -65,6 +65,13 @@ function start(args) {
     args.protocol = 'mqtts'
   }
 
+  if (args.port){
+    if (typeof args.port !== 'number') {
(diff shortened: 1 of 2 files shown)


mcollina:
That should already be handled by https://github.com/mqttjs/MQTT.js/blob/master/bin/pub.js#L23.
What's happening in minimist when `port`  is not a number? Is passed straight as a string?

roccomuso:
if you pass a string, like: <pre>-p 1883abc</pre>, it doesn't work, and if you do a <pre>console.log(typeof args.port)</pre> it prints <pre>string</pre>. So it seems to be passed straight as a string.

mcollina:
Thanks!! :)

PR #351, opened 2015-11-16
Handling wrong authentication.

roccomuso:
Done: emit an error on console if credentials are wrong, (both for pub and sub).

Todo: gotta be handled the 'No Authorizations' case.
I did some attempts on sub.js, editing the client.on('connect') callback this way:

      client.on('connect', function() {
        client.subscribe(args.topic, { qos: args.qos }, function(err, granted){
          if (err){
            console.warn('Error: ', err);
            client.end();
          }else console.log('Granted: ', granted);
        });

and this is my output on a topic that i'm not authorized to subscribe to:

    Granted: [ { topic: 'priv8', qos: 128 } ]

Weird, isn't it? The broker i'm using is Mosca. I'm implementing on Mosca the following functions: authenticate, authorizePublish, authorizeSubscribe.

### DIFF
--- a/bin/sub.js
+++ b/bin/sub.js
@@ -89,8 +89,16 @@ function start(args) {
   var client = mqtt.connect(args);
 
   client.on('connect', function() {
-    client.subscribe(args.topic, { qos: args.qos });
-  });
+    client.subscribe(args.topic, { qos: args.qos }, function(err, granted){
+      // Handle 'No authorization'
+      if (err){
+        console.warn('Error: ', err);
+        client.end();
+      }else if (granted.length && granted[0].hasOwnProperty('qos') && granted[0].qos > 2) {
+        console.warn('Error: client not authorized to subscribe to %s', granted[0].topic);
+        client.end();
+      }
+    });
 
   client.on('message', function(topic, payload) {
     if (args.verbose) {
@@ -105,7 +113,6 @@ function start(args) {
     console.warn(err);
     client.end();
   });
-  // TODO: alert impossibility to subscribe to a topic, 'No authorization'.
   
 }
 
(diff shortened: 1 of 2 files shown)


bin/sub.js
```diff
@@ -99,6 +99,14 @@ function start(args) {
       console.log(payload.toString())
     }
   });
+  
+  client.on('error', function(err){
+    // Handle 'No Authentication'
+    console.warn(err);
+    client.end();
+  });
+  // TODO: alert impossibility to subscribe to a topic, 'No authorization'.
```
mcollina: While you are at it, we should fix this TODO as well :).

roccomuso:
Is 'no authorization' error handled somewhere else? Do we need to catch any 'particular' mqtt packet?

mcollina:
If a subscription fails, you would find something different from 0,1,2 as the qos in the first argument of the `subscribe` callback: https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L801-L805.

roccomuso:
Subscription 'no authorization' alert handled: ab40be81c01cf0e85f09ad17efc48cce22fea67b

Any tips for the publish?

mcollina:
Not really, an unathorized publish will disconnect the client, there is no feedback from the broker.

mcollina:
Is this done?

roccomuso:
Not yet. How do we handle this, authorized or not, a client will be either disconnected, am i right?

mcollina:
Yes. But not really. in one case you'd get a 'bad' disconnect triggered from the server, on the other it will actually disconnect cleanly.
It's probably impossible to tell with QoS 0, as it disconnects straight away. But it should be possible to do something for QoS 1 & 2.

mcollina:
Merged with #345

PR #358, opened 2015-12-10
Fixed the ping rescheduling

### LINKED ISSUE #357: Mqtt connection is closed frequently because mqtt broker thinks the client is offline.
I look over the recent changes that have rescheduling for ping request. 

I thinks this is the cause of my problem. 

According to below part of mqtt specification, PINGREQ seems not to be related to data from server. 

############################################
The PINGREQ Packet is sent from a Client to the Server. It can be used to:
Indicate to the Server that the Client is alive in the absence of any other Control Packets being sent from the Client to the Server.
Request that the Server responds to confirm that it is alive.
Exercise the network to indicate that the Network Connection is active.
############################################

Therefore, I think that the client must send ping request regardless the data from server. So, please reconsider this change.

Thanks.

### LINKED ISSUE #359: in v1.6.0, the client keeps disconnecting.
After I updated to v1.6.0, I see that the client keeps disconnecting every 30 seconds.

4rzael:
Fixes #357
Fixes #359 

As said in the documentation, the ping should be rescheduled when sending packets, not on packet receive.

> It is the responsibility of the Client to ensure that the interval between Control Packets being sent does not exceed the Keep Alive value. In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet [MQTT-3.1.2-23]

## How to reproduce the error fixed by this PR

Here is an emitter client. It sends packet every seconds. There wasn't any problem with this one.
```js
var mqtt    = require('./');
var client  = mqtt.connect('mqtt://test.mosquitto.org', {keepalive: 1});

client.on('message', function (topic, message) {
    console.log(message.toString());
});

client.on('disconnect', function () {
    console.log('DISCONNECTION');
});

var i = 0;
setInterval(function () {
    client.publish('test4mqttjs', 'Hello mqtt' + i);
    console.log(i++ % 2 ? 'tac' : 'tic');
}, 1000);
```

Here is the listener. Here was the problem.

```js
var mqtt    = require('./');
var client  = mqtt.connect('mqtt://test.mosquitto.org', {keepalive: 3});

client.subscribe('test4mqttjs');

client.on('message', function (topic, message) {
    console.log(message.toString());
});

client.on('offline', function () {
    console.log('DISCONNECTION');
});
```

It should not disconnect, but with MQTT.js v1.6.0, it did when the emitter was up because of the ping rescheduler. See more infos in the associated issue.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -285,8 +285,6 @@ MqttClient.prototype._handlePacket = function (packet, done) {
       // or just log it
       break;
   }
-  // When a packet is received, reschedule the ping timer
-  this._shiftPingInterval();
 };
 
 MqttClient.prototype._checkDisconnecting = function (callback) {
@@ -593,6 +591,9 @@ MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
     return this.queue.push({ packet: packet, cb: cb });
   }
+
+  // When sending a packet, reschedule the ping timer
+  this._shiftPingInterval();
 
   switch (packet.qos) {
     case 2:
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -644,38 +644,27 @@ module.exports = function (server, config) {
       });
       setTimeout(done, 1000);
     });
-    it('should defer the next ping when a control packet is received', function (done) {
-      var client = connect({keepalive: 0.1}),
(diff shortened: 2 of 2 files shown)


mcollina:
Thanks!! :+1:

PR #361, opened 2015-12-19
ensure publish callbacks actually called after reconnecting

### LINKED ISSUE #360: Publish callbacks after reconnect not getting called back?
I'm publishing messages with qos 1 (about 2000/s) over a single MQTT.js client connection and am passing in a callback function to `publish` so that I know when I can continue publishing.

Occasionally, the connection goes down and MQTT.js reconnects, but I think I'm losing the publish callbacks for any messages that were inflight when the connection went down.

In `publish`, the callback is stored in the `outgoing` object [here](https://github.com/mqttjs/MQTT.js/blob/13c665c8593851d7470cd9162312d48d1feb7ac7/lib/client.js#L351).

In `_handleAck`, it gets retrieved [here](https://github.com/mqttjs/MQTT.js/blob/13c665c8593851d7470cd9162312d48d1feb7ac7/lib/client.js#L781) and then called back later on by the `outgoingStore.del` method.

But when the client reconnects, that callback is replaced with the `storeDeliver` function [here](https://github.com/mqttjs/MQTT.js/blob/13c665c8593851d7470cd9162312d48d1feb7ac7/lib/client.js#L129) before republishing the inflight packets. When `_handleAck` gets the puback packet, it calls `storeDeliver`, not my original callback.

Would you accept a PR that invokes the original callback passed in to `publish` and then calls `storeDeliver`?

jdiamond:
fixes #360

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -118,15 +118,23 @@ function MqttClient (streamBuilder, options) {
     // Control of stored messages
     outStore.once('readable', function () {
       function storeDeliver () {
-        var packet = outStore.read(1);
+        var packet = outStore.read(1),
+          cb;
         if (!packet) {
           return;
         }
         // Avoid unnecesary stream read operations when disconnected
         if (!that.disconnecting && !that.reconnectTimer && (0 < that.options.reconnectPeriod)) {
           outStore.read(0);
-          // Ensure that the next message will only be read after callback is issued
-          that.outgoing[packet.messageId] = storeDeliver;
+          cb = that.outgoing[packet.messageId];
+          that.outgoing[packet.messageId] = function () {
+            // Ensure that the original callback passed in to publish gets invoked
+            if (cb) {
+              cb();
+            }
+            // Ensure that the next message will only be read after callback is issued
+            storeDeliver();
+          };
           that._sendPacket(packet);
         } else if (outStore.destroy) {
           outStore.destroy();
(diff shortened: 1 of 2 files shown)


jdiamond:
I had to skip the style checks because I'm at work on a Windows box and they weren't working there. I'll fix the style errors found by Travis when I'm home on my Mac.

mcollina:
I'm :+1: on this. Fix the code style so that travis can pass.

I think we should get all the code style stuff pass on windows (maybe move to [standard](http://npm.im/standard) or [semistandard](http://npm.im/semistandard)).

Anyway, would you like to help assembling the v2 release? It seems you are keen on performance (2000 concurrent qos1), and the new one is at least 3x faster. There are a couple more things to do.

Also, would you like to be a contributor to this module? check out CONTRIBUTING.md.

jdiamond:
Travis is passing now.

I think the problem on Windows has something to do with cmd.exe not expanding wildcards the way Unix-based shells do. So it's looking for a file literally named "lib/*" and can't find it. Might be an easy fix, but I also have no objection to using standard if switching to that would be better.

I am very interested in the performance improvements. CPU utilization on the servers using MQTT.js is very high. We are load balancing to them to allow us to scale, but it would be nice if they struggled less.

I just read CONTRIBUTING.md and will not force push anymore. I'm used to doing that on my personal branches to submit clean PRs (but never master), but I can not do it for this project.

On the v2 PR, I don't have any experience with Browserify so can't help there. A big change like switching to standard seems like it should come after the open PRs are merged to avoid conflicts?

mcollina:
@jdiamond add yourself to the README, I've made you a collaborator.

PR #362, opened 2015-12-22
Ensure subscribe callback is set before checkDisconnect

knolleary:
If `MqttClient.subscribe` is called with a topic and options object, but no callback function, then the options object is being passed to `_checkDisconnect` as the callback. If the client is currently disconnecting then this results in `TypeError: object is not a function` as it tries to invoke the callback which is actually the options object.

The `subscribe` function already does the necessary work to properly set `opts` and `callback` depending on what was passed into the function, but the call to `_checkDisconnect` happens before that is done.

This fix moves the call to `_checkDisconnect` to below the point where callback will be guaranteed to be a function.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -97,8 +97,6 @@ function MqttClient (streamBuilder, options) {
   this.disconnecting = false;
   // Packet queue
   this.queue = [];
-  // Are we intentionally disconnecting?
-  this.disconnecting = false;
   // connack timer
   this.connackTimer = null;
   // Reconnect timer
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks for spotting this @knolleary! Would you mind adding a quick unit test, so we don't have any regression on this in the future?

There is also a small code style issue in travis.

knolleary:
Well, that's a bit embarrassing - you'd think I'd know better.

Updates on their way shortly.

mcollina:
ahahha no worries. I keep doing these mistakes when on the other side :D.

BTW, node-red is one of the main dependents of MQTT.js. Would you like to help maintaining this one too?

knolleary:
Tests added and Travis passing.

Happy to chip in when I can, although node-red does keep me pretty busy.

PR #372, opened 2016-01-12
Fix to support TLS server and client auth for wss://

ryankurte:
Added options to pass ca, cert, and key options to the underlying websocket-stream instance. This allows use with self-signed CAs and certificate based client authentication.

Tested against a mosquitto install, but not sure how to test as part of this project as the server does not appear to (currently) support wss.

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -16,6 +16,21 @@ function buildBuilder (client, opts) {
     if (opts.hasOwnProperty('rejectUnauthorized')) {
       wsOpt.rejectUnauthorized = opts.rejectUnauthorized;
     }
+    if (opts.hasOwnProperty('ca')) {
+      wsOpt.ca = opts.ca;
+    }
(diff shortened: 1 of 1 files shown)


mcollina:
I think we should possibly copy all of these using a loop.

Regarding the test.. it might take a bit of work to get that rolling. We
are removing the server in the next version anyway.
Il giorno mar 12 gen 2016 alle 04:02 Ryan <notifications@github.com> ha
scritto:

ryankurte:
Fair enough, is there a reason the opts argument isn't just passed directly?

With regards to testing, [test.mosquitto.org](http://test.mosquitto.org/) has endpoints for each connection type, so the examples / tests could locally run against that.

ryankurte:
@mcollina what is required to get this merged in?

mcollina:
Me doing it :)

ryankurte:
Awesome, thanks!

PR #374, opened 2016-01-18
Bumped dependencies

psorowka:
- mqtt-connection
- mqtt-packet
- websocket-stream

I had to increase the timeouts for the ping deferal test in order to get the tests to run.. We should discuss on what's going on there. Functionality seems to be fine and keepalive < 1sec doesn't seem to be a realistic requirement. 

btw the test with 0.1s doesn't run on v1.6.3 at my machine either - but obviously they passed on your side..

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,11 +51,11 @@
     "help-me": "^0.1.0",
     "inherits": "^2.0.1",
     "minimist": "^1.1.0",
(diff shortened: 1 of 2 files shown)


psorowka:
ahh sorry just realized there is #345 already in the pipeline. I'll close this and try to push that one forward

PR #375, opened 2016-01-22
bump websocket-stream to 3.0.1 for nsp fix

ralphtheninja:
mqtt-nsp

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "mqtt-packet": "^3.2.0",
     "readable-stream": "~1.0.2",
     "reinterval": "^1.0.1",
-    "websocket-stream": "^2.0.2",
+    "websocket-stream": "^3.0.1",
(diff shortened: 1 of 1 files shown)


mcollina:
Yeah, I know of this particular issue :).
I didn't fix/release it before now because we are not affected by that specific bug (we are validating that we are __always__ sending a `Buffer`). Also websocket-stream changes a bit in v3, as it becomes an object mode stream.

But you are right that we need to get this fixed and release asap, just to please nsp.

ralphtheninja:
> we are validating that we are always sending a Buffer

Aah cool. That's good to know! So my projects using `mqtt.js` can ignore nsp on this particular issue.

mcollina:
Released as 1.7.0.

ralphtheninja:
> Released as 1.7.0.

Rock n roll. Thanks Matteo!

PR #385, opened 2016-02-16
Clear the connack timeout on close event

gkwicker:
Shouldn't the connack timeout be cleared if a close event is received?  If it isn't cleared, it can interrupt a reconnect attempt in progress.


4rzael:
Yes, IMHO, you're right ! Thanks !

PR #386, opened 2016-02-22
Added DefinitelyTyped badge to README.md

PekkaPLeppanen:
I created a TypeScript type definition of the MQTT.js a while ago. It can be found in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repository. This instance provides a badge for a project, which has a type definition.

Instead of the experimental badge, one could just mention about the type definitions somewhere in the text or just leave the README as it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 [![Build Status](https://travis-ci.org/mqttjs/MQTT.js.svg)](https://travis-ci.org/mqttjs/MQTT.js)
 
+[![DefinitelyTyped](http://img.shields.io/badge/Definitely-Typed-blue.svg)](https://github.com/borisyankov/DefinitelyTyped/tree/master/mqtt)
+
 [![NPM](https://nodei.co/npm/mqtt.png)](https://nodei.co/npm/mqtt/)
 [![NPM](https://nodei.co/npm-dl/mqtt.png)](https://nodei.co/npm/mqtt/)
 


mcollina:
A couple of questions here, as I know nothing about TypeScript. Should this be better hosted in this repo? Does TS support configuring these from `package.json` or something similar?

PekkaPLeppanen:
To host the type definitions in this repo is an option, but it might be better to keep it in the DefinitelyTyped.

1. Type definitions can have references to other type definitions. MQTT is referencing NodeJS. These references are maintained in the same repo.
2. DefinitelyTyped repo is a go-to-repo when it comes to type definitions.
3. There're Gulp, Grunt and IDE plugins to automatically fetch definition files for packages listed in package.json or bower.json. These plugins use DefinitelyTyped repo.

Configuring the TS from JSON file is only for building phase AFAIK.

mcollina:
That's great, so it seems the DefinitelyTyped repo is the way to go. But how do you handle type differences in different versions of the library?

PekkaPLeppanen:
The guide to support multiple API versions is at the DefinitelyTyped [contribution guide](http://definitelytyped.org/guides/contributing.html).

mcollina:
Awesome, thanks! Might I ask you to watch this repo and keep this in sync? Also check out #345.

(If you want to contribute more here, you are highly welcome to! :))

PekkaPLeppanen:
Will do. It's in my interests to do so, because I work for IoT projects in a company where MQTT is a important part of the solution stack.

PR #387, opened 2016-02-24
Introduce 'fastClientDisconnectDetect' option

gkwicker:
If the client publishes faster than the keepalive interval, PINGREQ messages keep getting rescheduled and will not be sent.  If connectivity is lost between the client and the broker during such a period, the client may not be able to detect that its published messages aren't reaching the broker.  This change introduces a new option ('fastClientDisconnectDetect') to **not** reschedule PINGREQ messages when sending packets.

This is in accordance with the MQTT spec v3.1.1, paragraph 3.1.2.10 (Keep Alive), which states that clients may send PINGREQ at any time:
```
The Client can send PINGREQ at any time, irrespective of the Keep Alive value, and use the PINGRESP to determine that the network and the Server are working.
```

### DIFF
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -605,6 +605,114 @@ module.exports = function (server, config) {
         done();
       });
     });
+    it('should not checkPing if publishing at a higher rate than keepalive', function (done) {
+      var interval = 3,
+        client = connect({keepalive: interval});
+
+      client._checkPing = sinon.spy();
+
+      client.once('connect', function () {
+
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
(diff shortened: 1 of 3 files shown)


mcollina:
Linting fails.

Can you please add a unit test and update the README? Also, I would try to shorten the name of the configuration, something like `reschedulePings`, and add a default to `true`.

test/abstract_client.js
```diff
@@ -605,6 +605,114 @@ module.exports = function (server, config) {
...
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
+        client.publish('foo', 'bar');
+        clock.tick(interval * 500);
```
mcollina: maybe a for loop?

mcollina:
Thanks! Merging and releasing!

Issue #389, opened 2016-02-29
Microsoft Edge does not receive callbacks while connection succeded


Pedrozena:
Apparently, the event "connect" is never fired while using mqtt.js with microsoft Edge (25.10586.0.0).
From my tests, the broker states that the Microsoft Edge client is correctly connected, but this information does not get intercepted by the callback function attached to the event "connect" (as it correctly happen using Firefox or Chrome)

mcollina:
You are possibly in the best position to fix this, I have access to a win
10 box, but I will have no time for the next 3 weeks :/. Maybe @sandro-k?
Il giorno lun 29 feb 2016 alle 13:38 Pedrozena <notifications@github.com>
ha scritto:

sandro-k:
I can have a look, but I only have the issue on IE10, Edge works fine.

Pedrozena:
@sandro-k  I got it working properly in IE11..

seriousme:
I have the same issue, it works fine in Chrome and IE11, but the "connect" event is not picked up  by Edge (version 10586.0.0) 
Broker (Mosca ;-)) shows the the client timing out:

```
{"name":"mosca","hostname":"Mypc","pid":1460,"client":"mqttjs_77d2c9cd", "level":30,"msg":"client connected","time":"2016-03-29T16:46:57.729Z","v":0}
{"name":"mosca","hostname":"Mypc","pid":1460,"client":"mqttjs_77d2c9cd","level":30,"msg":"keepalive timeout","time":"2016-03-29T16:47:12.731Z","v":0}
{"name":"mosca","hostname":"Mypc","pid":1460,"client":"mqttjs_77d2c9cd","level":30,"msg":"closed","time":"2016-03-29T16:47:12.731Z","v":0}
```

RangerMauve:
Seems like it'd be an error in [websocket-stream](https://github.com/maxogden/websocket-stream/issues)

mcollina:
Might be. Can you reproduce just using websocket-stream?

RangerMauve:
@mcollina We're currently looking into this on my end, will update when we figure out what's wrong. So far it looks like [onReady](https://github.com/maxogden/websocket-stream/blob/master/stream.js#L88) in `websocket-stream` isn't being invoked for some reason.

seriousme:
[...]

PR #394, opened 2016-04-09
Saucelab integration on Travis

mcollina:


mcollina:
@sandro-k please have a look at the setup. You might want to do the same for your polymer components.

Any opinions on this, should I merge?

This might help in making sure #389 does not repeat again. However it supports only an ancient version of Edge :(, so we can't really add it here (MQTT.js does not work on that version of edge, because of https://github.com/feross/buffer/issues/86.

sandro-k:
@mcollina I can have a look, but right now I am in the US and will not be able to do this before may.

mcollina:
@sandro-k no worries :). Merging now.

Issue #397, opened 2016-04-19
How to turn off outgoing message cache?


Densaugeo:
I have an MQTT client using mqtt.js, and publishing messages to a broker. If the broker goes offline, the client saves its messages up and sends them all at once when the server reconnects, creating a flood of traffic. Is there a way to turn off this caching? I don't need to preserve old messages, only saving the most recent is fine.

Also, I have qos set to zero. My understanding is that this means messages should be 'fire and forget', with no caching.

mcollina:
So, caching is the wrong term, as we are not caching anything. The correct term is 'offline queue', or 'send queue'. Can you please reframe your issue? 

Disabling the queue is currently not possible.

I agree with you that we should address this is a bug that will need fixing in a next major release #345 (or after that one). The whole queue system needs to be highly refactored to address various issues. I currently have no bandwidth to do so.

Densaugeo:
I can't speak to refactoring the whole system, but after looking at Client.js a bit, it shouldn't be hard to add an option for 'sendQueue' or something like that, and check it in MqttClient.prototype._sendPacket() before adding a packet to the queue. I can send a PR for this in a bit.

I haven't contributed to this project before, is there anywhere I would need to update tests and documentation other than Readme.md and test/client.js?

mcollina:
> I can't speak to refactoring the whole system, but after looking at Client.js a bit, it shouldn't be hard to add an option for 'sendQueue' or something like that, and check it in MqttClient.prototype._sendPacket() before adding a packet to the queue. I can send a PR for this in a bit.

That can solve your issue, but that option will break a lot of the behavior of MQTT.js. The whole point is that you need to check if it's a publish QoS 0 and then discard that. All the other stuff would need to be queued anyway.

> I haven't contributed to this project before, is there anywhere I would need to update tests and documentation other than Readme.md and test/client.js?

@Densaugeo no.

Densaugeo:
Would it be reasonable to only queue outgoing packets if they have QoS > 0? I think this would be correct behavior according to the MQTT spec, but would also mean a change of behavior (since QoS 0 packets are currently being queued). Is that why you mention v2?

For the meantime, I propose an option to enable this behavior (while still queuing QoS 1 and 2 packets regardless).

mcollina:
Yes to all questions, good work!

4rzael:
Hum, looks like this issue can be closed, as it's been fixed in #399 ?
@mcollina

[...]

PR #399, opened 2016-04-19
Add .qosZeroNotQueued option to MqttClient

Densaugeo:
My proposal for #397

The default behavior remains identical to existing behavior. If a client's .qosZeroNotQueued option is set to true, then it will stop queuing QoS zero messages, but will still queue QoS 1 and 2 messages if disconnected.

I wasn't sure what branch to put this in, and it isn't a breaking change, so I put it in master. Let me know if it should go somewhere else.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -91,7 +91,7 @@ function MqttClient (streamBuilder, options) {
   this.incomingStore = this.options.incomingStore || new Store();
 
   // Should QoS zero messages be queued when the connection is broken?
-  this.qosZeroNotQueued = this.options.qosZeroNotQueued || false;
+  this.queueQoSZero = null == this.options.queueQoSZero ? true : this.options.queueQoSZero;
 
   // Ping timer, setup in _setupPingTimer
   this.pingTimer = null;
@@ -600,14 +600,10 @@ MqttClient.prototype._cleanUp = function (forced, done) {
  */
 MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
-    if (0 === packet.qos && this.qosZeroNotQueued) {
-      if (cb) {
-        cb(new Error('No connection to broker'));
-      } else {
-        this.emit('error', new Error('No connection to broker'));
-      }
-    } else {
+    if (0 < packet.qos || 'publish' !== packet.cmd || this.queueQoSZero) {
       this.queue.push({ packet: packet, cb: cb });
+    } else if (cb) {
+      cb(new Error('No connection to broker'));
     }
 
     return;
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -597,7 +600,17 @@ MqttClient.prototype._cleanUp = function (forced, done) {
  */
 MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
-    return this.queue.push({ packet: packet, cb: cb });
+    if (0 === packet.qos && this.qosZeroNotQueued) {
```
mcollina: you should also check if `cmd`  is `publish`. The `qos` bit has different meaning with other packets.

lib/client.js
```diff
@@ -597,7 +600,17 @@ MqttClient.prototype._cleanUp = function (forced, done) {
...
+      if (cb) {
+        cb(new Error('No connection to broker'));
+      } else {
+        this.emit('error', new Error('No connection to broker'));
```
mcollina: I'm not convinced we should error the client for sending a message when not connected.
If it's QoS 0, then you are accepting that the packet can be lost. I think we should just ignore this packet if there is no callback.

test/abstract_client.js
```diff
@@ -271,6 +271,23 @@ module.exports = function (server, config) {
       });
     });
 
+    it('should not queue qos 0 messages if qosZeroNotQueued is true', function () {
+      var client = connect({qosZeroNotQueued: true});
+
+      client.publish('test', 'test', {qos: 0}, function () {/* Silence 'no connection' error */});
```
mcollina: you should add an assertion verifying that an error is emitted.

mcollina:
I think we can revert the option logic, something like `queueQoSzero`, defaulting it to true.

Densaugeo:
I went back and forth on whether it should error - it is QoS zero, but I also favor error-handling by default. If you think we should skip it, I'll adjust it accordingly.

For the option logic, I originally had it as a default-true option, which I changed because it needed a ternary operator or something similar to set it (`this.options.queueQoSZero || true` will always return true...). Maybe not the best reason. I suppose keeping the API easy to understand is more important.

mcollina:
negative options are hard to reason about. I'm happy with the ternary.

`cb(new Error('..'))` is nice, I'm concerned only with `emit('error')`.

mcollina:
I'm 👍 on this. Let's wait a couple of days if any others have opinions.

Densaugeo:
Having problems with Travis. Ran the tests twice, the first the 'should publish a message (offline)' test timed out once in the Node 4 environment. The second time the same test timed out, but in the 0.12 environment this time. Maybe the test timeouts need to be extended for Travis?

Looks like the finished build went back to pending and is rerunning the 0.12 part? Not really sure what is going.

mcollina:
@Densaugeo it was me rerunning the failed task.

Probably the timeout on travis needs to be extended, now they are passing anyway.

mcollina:
Thanks! Release as v1.9.0!

Densaugeo:
Just updated my dependencies. Thanks!

PR #405, opened 2016-05-08
Added mqtt.Client.isConnected()

### LINKED ISSUE #379: No way to get state of client (connected or not)
Right now, the implementation seems to require that the state of the connection be kept externally by monitoring all of the various events (connect, close, reconnect, error, etc.).  It would be extremely valuable to have a method on the mqtt client object to check the connection state.

4rzael:
Fixes #379


mcollina:
I think we can just expose the connected property. Why wrapping it in a accessor?

4rzael:
Using a getter makes explicit the fact that it's just a read-only value, and that it **must not** be changed by hand, because it's used internally.

But we can just expose it, sure.

mcollina:
I think it will make more sense in the current API to have just connected. Any other opinions?

RangerMauve:
Personally, I like the simplicity of the `connected` attribute more than an `isConnected` method.

RangerMauve:
However, I don't really see myself using it much since it seems like reacting to events is a better way to go in the long run

PR #406, opened 2016-05-11
Removed [options] param from mqtt.Client#unsubscribe

shylesh107:
The number of parameters for mqtt.Client#unsubscribe was wrong in the documentation. Unsubscribe method doesn't accept an options parameter.


mcollina:
Thanks for the fix! :)

Issue #407, opened 2016-05-15
Propose 'publish' event for MqttClient


Densaugeo:
I've got a client that sometimes needs to log everything it publishes. Perhaps adding a 'publish' event to the client would be useful?

**Existing method 1**
So far, I can add a log statement after every publish:
~~~
mqtt_connection.publish('/some/topic', 'a message');
logger.debug('Sent to MQTT:', { topic: '/some/topic', message: 'a message' });
...
~~~
but that is verbose, and I may miss a log statement.

**Existing method 2** (doesn't actually work this way)
There's a handy callback option:
~~~
var logger_cb = function(err, result) {
  logger.debug('Sent to MQTT:', result);
};
mqtt_connection.publish('/some/topic', 'a message', logger_cb);
~~~
but the callback only gets error arguments, no information about successful publishes is sent to the callback! So, with that approach, I would need a different callback for each publish, which looks basically the same as the first option.

**Existing method 3**
Or, I could make a custom .publishAndLog method:
~~~
mqtt.MqttClient.prototype.publishAndLog = function(topic, message, opts, callback) {
  this.publish(topic, message, opts, callback);
  
  logger.debug('Sent to MQTT:', { topic: topic, message: message });
}
mqtt_connection.publishAndLog('/some/topic', 'a message');
...
~~~
which works rather well, but I would prefer to use the standard .publish() method to avoid confusion.

**This would be nicer**
I would like to add a publish event to the client, so that I could use something like this:
~~~
mqtt_connection.on('publish', function(topic, message, packet) {
  logger.debug('Sent to MQTT:', { topic: topic, message: message });
});
mqtt_connection.publish('/some/topic', 'a message');
...
~~~
Passing the internal packet object to a callback on success could also be helpful, to allow something like the `logger_cb` case above.

Before I get started on it, are there any comments/suggestions for the new event? I am thinking of making it fire when the cb would normally be called on success, and giving it the same arguments as the 'message' event.

mcollina:
I'm 👍 on firing the event, go ahead.

PR #410, opened 2016-05-24
Add publish event to mqtt.Client

Densaugeo:
For #407.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -29,22 +29,21 @@ function defaultId () {
 }
 
 function sendPacket (client, packet, cb) {
-  function emitAndCb () {
-    if ('publish' === packet.cmd) {
-      client.emit('publish', packet.topic, packet.payload, packet);
-    }
-    if (cb) {
-      cb();
-    }
-  }
-
   try {
     var buf = mqttPacket.generate(packet);
 
-    if (client.stream.write(buf) && (cb || 'publish' === packet.cmd)) {
-      emitAndCb();
-    } else if (cb || 'publish' === packet.cmd) {
-      client.stream.once('drain', emitAndCb);
+    if (client.stream.write(buf)) {
+      client.emit('packetsent', packet.topic, packet.payload, packet);
+      if (cb) {
+        cb();
+      }
+    } else {
+      client.stream.once('drain', function () {
+        client.emit('packetsent', packet.topic, packet.payload, packet);
+        if (cb) {
+          cb();
+        }
+      });
     }
   } catch (err) {
     if (cb) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -270,6 +270,8 @@ MqttClient.prototype._setupStream = function () {
 };
 
 MqttClient.prototype._handlePacket = function (packet, done) {
+  this.emit('packetreceive', packet);
+
   switch (packet.cmd) {
     case 'publish':
       this._handlePublish(packet, done);
--- a/lib/client.js
+++ b/lib/client.js
@@ -29,25 +29,22 @@ function defaultId () {
 }
 
 function sendPacket (client, packet, cb) {
+  function emitAndCb () {
+    if ('publish' === packet.cmd) {
+      client.emit('publish', packet.topic, packet.payload, packet);
+    }
+    if (cb) {
+      cb();
+    }
+  }
+
   try {
     var buf = mqttPacket.generate(packet);
 
-    if (client.stream.write(buf)) {
-      if ('publish' === packet.cmd) {
-        client.emit('publish', packet.topic, packet.payload, packet);
-      }
-      if (cb) {
-        cb();
-      }
-    } else {
-      client.stream.once('drain', function () {
-        if ('publish' === packet.cmd) {
-          client.emit('publish', packet.topic, packet.payload, packet);
-        }
-        if (cb) {
-          cb();
-        }
-      });
+    if (client.stream.write(buf) && (cb || 'publish' === packet.cmd)) {
+      emitAndCb();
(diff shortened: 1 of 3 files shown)


test/abstract_client.js
```diff
@@ -419,6 +419,25 @@ module.exports = function (server, config) {
       });
     });
 
+    it('should emit a publish event', function (done) {
+      var client = connect(),
+        payload = 'test_payload',
+        topic = 'test_topic';
+
+      client.once('publish', function (topic, message, packet) {
+        topic.should.equal(topic);
```
sandro-k: At this point it is not 100% clear to which `topic` is referenced, you can use `packet.topic.should.equal(topic)` or rename `topic = 'test_topic'` to `testTopic = 'test_topic'`

Densaugeo: Nonetheless, you're right. The topic.should.equal line isn't actually testing anything. I'll adjust that.

I thought I checked this thread earlier but didn't see any comments. Oh well.

sandro-k:
LGTM 👍

lib/client.js
```diff
@@ -29,12 +29,22 @@ function defaultId () {
...
 
 function sendPacket (client, packet, cb) {
+  function emitAndCb () {
+    if ('publish' === packet.cmd) {
+      client.emit('publish', packet.topic, packet.payload, packet);
+    }
+    if (cb) {
+      cb();
+    }
+  }
```
mcollina: Using a function here is definitely not ok, as it will cause a perceivable slowdown in the benchmarks (at least in #345). Moreover, this function will be allocated even for every other packet (where there is no need).

mcollina: Given how the logic work, we can generate a wrapping function only if `write` returns false.

However, for logging purposes, we might even emit it anyway even if `write` returns false.

mcollina:
Given the examples in #407, those will have a _different_ meaning than the callback for QoS 1 and 2.
I am not sure this is the behavior we want.

sandro-k:
Maybe it would be sufficient to log QoS1 and 2 after the PUBACK, as QoS0 is fire and forget anyway.

mcollina:
Maybe there is another option: emit a `packetSent`  event, to be emitted after `write()`, with _every_ packet, and then leave the QoS handling to the user. That will solve all the logging issues, and in fact it can serve a greater audience, as it can track the full flow of messages in the client.

Densaugeo:
So, you are proposing `packetSent` be fired after all writes, including for non-publish packets? That sounds reasonable - the events can be easily filtered, and that would allow more thorough logging if it is desired.

mcollina:
@Densaugeo exactly that.

Densaugeo:
It doesn't look like MQTT.js has any two-word event names yet, so I went with `packetsent` all lower case to match the naming convention in browsers.

lib/client.js
```diff
@@ -31,10 +31,19 @@ function defaultId () {
...
+      if (cb) {
+        cb();
+      }
+    } else {
+      client.stream.once('drain', function () {
+        client.emit('packetsent', packet.topic, packet.payload, packet);
+        if (cb) {
+          cb();
+        }
+      });
```
mcollina: waiting "drain" does not really help in guaranteeing a packet was sent. It's used only for handling backpressure.
I would avoid allocating a function and emitting the event anyway, without waiting.

Also, I would just have `client.emit('packetsent', packet)`.

mcollina:
How about we also add `'packetreceived'` event? The whole debuggability of MQTT.js always lacked those.

are there any opinion pro or against camelCase in events? I do not care.

4rzael:
I totally agree for the `packreceived` event.

However, I don't know about the camelCase in events. If there is such a naming convention (I haven't heard of it), following it would be a good idea.

mcollina:
how about we use `on('receive')` and `on('send')`? Shorter and there is no problem of disambiguation here.

4rzael:
I agree with the `receive` and `send` events. 

However, it might create some confusion between the `message` and the `receive` events, if not documented correctly. (it should be explicit that this event should **NOT** be used to receive messages, as it also fires on non-publish packets)

mcollina:
i guess we should name correctly in the README: these are debug events for 3rd parties that want to support logging etc.

Densaugeo:
The closest there is to a convention for event names is the event names used in browsers, which are all lower case (`keypress`, `doubleclick`, etc.). I'd be a little concerned about confusing `receive` with `message` and `send` with `publish`.

The topic and message arguments aren't necessary, that was only to make it match the arguments for `message` events. For a debugging event, that isn't really important.

mcollina:
Let's follow the browser convention and also use present tense verbs. How about sendpacket and receivepacket?
Anyway, any of these are ok, bikeshedding names is not really helping.

Densaugeo:
Both good points. I'll update it when I'm back from my convention and awake.

Densaugeo:
So I've changed `packetsent` to `packetsend`, which follows the browser convention of all lowercase letters, present-tense verbs, and nouns first. I propose that this is enough discussion of event names

Is the description for the readme clear? I'd like it to be very obvious that this is for debugging, not listening to messages.

If all are in agreement on the `packetsend` event, I will add a `packetreceive` event to match. After skimming through code, it looks like it could be emitted in `._handlePacket()` or maybe `.process()`.

mcollina:
I would emit it in handlePacket, before the packet is handled, so we can
use t to debug error situations.
Il giorno mer 1 giu 2016 alle 01:39 Densaugeo <notifications@github.com> ha
scritto:

mcollina:
Thanks, releasing soon!

PR #420, opened 2016-06-13
Export MqttClient from lib/connect/index.js to support browserization

gkwicker:
This change allows a browserified application to use the MqttClient exported by mqtt.js.

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -134,3 +134,4 @@ function connect (brokerUrl, opts) {
 
 module.exports = connect;
 module.exports.connect = connect;
+module.exports.MqttClient = MqttClient;


mcollina:
Can you please explain the rationale of this a bit further, as the main way to connect should be through `mqtt.connect()`?

(I know it's probably because of aws-iot-sdk).

gkwicker:
Hi Matteo – we haven't been able to use the mqtt.connect() method because a WebSocket connection to AWS IoT passes variable authentication information in the URL query string.  If the connection has to be restarted (due to error/network connectivity issue/etc...), the query string part of the URL will be different.  As a result, we pass a [custom streamBuilder function](https://github.com/aws/aws-iot-device-sdk-js/blob/master/device/index.js#L487) to the MqttClient object when it is instantiated to take care of this.  This doesn't cause any issues when running in Node, but when browserified we see an issue which requires that MqttClient be exported from lib/connect/index.js.

mcollina:
thx!

PR #426, opened 2016-06-23
Use "mqtt" as WS protocol unless using v3

### LINKED ISSUE #408: mqtt over websockets using google chrome
Hello,

I am using this MQTT client in a web application to communicate with a python broker.

In the connect message, the browser (Chrome) outputs in the console the following error:

`WebSocket connection to 'ws://127.0.0.1:8888/' failed: Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received`

In my broker, the following error is presented:

`[2016-05-19 11:46:57,027] :: INFO :: hbmqtt.broker :: Connection from 127.0.0.1:53516 on listener 'ws-mqtt'
[2016-05-19 11:46:57,027] :: INFO :: websockets.protocol :: Failing the WebSocket connection: 1006  
WARNING :: hbmqtt.broker :: MQTT-3.1.0-1: Can't read first packet an CONNECT: No more data`

However, using Firefox, as well as Safari, the MQTT client works fine.

Broker Output for Firefox:

`[2016-05-19 11:47:32,843] :: INFO :: hbmqtt.broker :: Connection from 127.0.0.1:53524 on listener 'ws-mqtt'`

I have a test [project](https://github.com/vasco-santos/mqttOverWebSockets) for this issue.

RangerMauve:
Fixes #408

### DIFF
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -256,18 +256,18 @@ module.exports = function (server, config) {
 
     it('should use mqtt as the protocol by default', function (done) {
       server.once('client', function (client) {
-        client.stream.socket.protocol.should.equal("mqtt");
+        client.stream.socket.protocol.should.equal('mqtt');
         done();
-        client.end();
+        client.disconnect();
       });
       connect();
     });
 
     it('should use mqttv3.1 as the protocol if using v3.1', function (done) {
       server.once('client', function (client) {
-        client.stream.socket.protocol.should.equal("mqtt");
+        client.stream.socket.protocol.should.equal('mqttv3.1');
         done();
-        client.end();
+        client.disconnect();
       });
       connect({
         protocolId: 'MQIsdp',
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -253,6 +253,28 @@ module.exports = function (server, config) {
       client1.end();
       client2.end();
     });
+
+    it('should use mqtt as the protocol by default', function (done) {
+      server.once('client', function (client) {
+        client.stream.socket.protocol.should.equal("mqtt");
+        done();
+        client.end();
+      });
+      connect();
+    });
+
+    it('should use mqttv3.1 as the protocol if using v3.1', function (done) {
+      server.once('client', function (client) {
+        client.stream.socket.protocol.should.equal("mqtt");
+        done();
+        client.end();
+      });
+      connect({
+        protocolId: 'MQIsdp',
+        protocolVersion: 3
+      });
+    });
+
   });
 
   describe('offline messages', function () {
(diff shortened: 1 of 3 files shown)


mcollina:
Can we add a unit test for this?

RangerMauve:
I'm having a bit of an issue trying to figure out how to check that the protocol was being passed correctly to the websocket. By the looks of it, it should be accessible via `client.stream.socket.protocol`. But it seems like [that only gets set by the server](https://github.com/websockets/ws/blob/f30da5dad547595c7df134be97be919b4cb5062a/lib/WebSocket.js#L552)

RangerMauve:
Alternatively I'd need to somehow set up servers that expect 3.1 or 3.1.1 for the tests, and I think that's a bit outside the range of what I can do at the moment.

mcollina:
@RangerMauve yes, it's get set by the server. Here is the server https://github.com/mqttjs/MQTT.js/blob/master/test/websocket_client.js#L22-L29. you can just create a test and check if the ws connection has the expected subprotocol.

RangerMauve:
What should I do about the encrypted sockets? I have a test over [here](https://github.com/mqttjs/MQTT.js/blob/master/test/abstract_client.js#L256) at the moment that works on non encrypted sockets properly

mcollina:
Don't worry about encrypted websockets.

RangerMauve:
Well, currently I have the test in abstract_client, so it's getting executed for encrypted sockets as well. Is there a better place for me to put it?

mcollina:
Can you please commit those, I don't understand. This is a test specific
for websockets.
Il giorno gio 23 giu 2016 alle 23:52 RangerMauve <notifications@github.com>
ha scritto:

RangerMauve:
Here's my attempt at the tests. It just checks to see that the protocol gets set, to the default, and to `mqttv3.1` if 3.1 is used as you described in the issue.

test/abstract_client.js
```diff
@@ -253,6 +253,28 @@ module.exports = function (server, config) {
...
+      server.once('client', function (client) {
+        client.stream.socket.protocol.should.equal('mqttv3.1');
+        done();
+        client.disconnect();
+      });
+      connect({
+        protocolId: 'MQIsdp',
+        protocolVersion: 3
+      });
+    });
```
mcollina: You should move these to https://github.com/mqttjs/MQTT.js/blob/master/test/websocket_client.js.

mcollina:
Perfect, thanks!

mcollina:
Released as 1.12.0.

vasco-santos:
Hello
First of all, thanks for the update. However, I tested this new version (1.12.0) using Google Chrome and the browser does not connect with the broker yet.

Here is my test project, with a python3 broker and a web page:
[Test Project](https://github.com/vasco-santos/mqttOverWebSockets)

Thanks for your help!

PR #429, opened 2016-06-28
Adding bower.json file to add to bower registry

ccravens:
It would be great for this package to be available via bower. This addition provides the required bower.json file, and the project can be registered on bower via "bower register mqtt https://github.com/mqttjs/MQTT.js"


ccravens:
Submitting new PR with updated bower.json file

PR #430, opened 2016-06-28
Adding bower.json file to add to bower registry

ccravens:
It would be great for this package to be available via bower. This addition provides the required bower.json file, and the project can be registered on bower via "bower register mqtt https://github.com/mqttjs/MQTT.js"

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -13,7 +13,7 @@
     "publish",
     "subscribe"
   ],
-  "homepage": "https://github.com/ccravens/MQTT.js",
+  "homepage": "https://github.com/mqttjs/MQTT.js",
   "moduleType": [],
   "private": true,
   "ignore": [
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,26 @@
+{
+  "name": "mqtt",
+  "description": "A library for the MQTT protocol",
+  "main": "mqtt.js",
+  "authors": [
+    "Adam Rudd <adamvrr@gmail.com>",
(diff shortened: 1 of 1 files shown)


RangerMauve:
Doesn't bower require the code to be built into something other than CommonJS?

mcollina:
Yes it does. I'm not fan of committing a built package to the repo, but I see it might be helpful for the bower community. Any other opinion?

sandro-k:
From my perspective 
> Bower  - A package manager for the web 

is a package is something that runs in the Browser or similar like Chrome Packaged Apps and therefor should include a bundled version of MQTT.js for the Browser. Publishing MQTT.js to Bower like this does not benefit over `npm install --save-dev mqtt` and then bundle it yourself. What would be your benefit of publishing `MQTT.js` to Bower?

ccravens:
Most JavaScript libraries are available via bower and a lot of projects use bower to separate front end dependencies from NPM for Node (server side) dependencies. There is nothing stopping someone from using NPM for mqtt.js, but it would be a nice to have for people that use bower. In my case I leverage Mqtt.js and Mosca extensively, and Mqtt.js is the only project I have to fetch from node_modules rather than bower.

mcollina:
@ccravens the problem with this setup is that it will not work. You need a browserified thing rather than the common js module.

sandro-k:
Bower allows to run a `postinstall script` that we can use to bundle MQTT.js for the Browser and avoid to publish it via Git.

mcollina:
Any updates on this? Feel free to reopen with a way to bundle `mqtt.js` with browserify.

Issue #172, opened 2014-04-20
odd behavior when subscribing to topics with wildcards


rjbordon:
Hi there,

I've having some issues when subscribing to topics with wildcards. If I subscribe my client to "devices/+/activation" a message published to "devices/qwerty/metadata" if being processed by my client, which is in fact an error.

I'm using mosquitto as MQTT broker. This implementation supports wildcards for topics, and when I try this functionality with the mosquitto built-in client it works as expected. (just in case: http://mosquitto.org/man/mosquitto-8.html)

Does MQTT.js support this functionality? or I missing something else?

Thx in advance.

VikramTiwari:
@mcollina Not exactly the same behaviour, but unable to publish/subscribe on the wildcard topics.

listening to `mqtt sub -t 'hello/vikram/hi' -h 'test.mosquitto.org' -v`
publishing to `mqtt pub -t 'hello/+/hi' -h 'test.mosquitto.org' -m 'from MQTT.js' `

getting no data on the subscriber side.
PS: Using node v4.1.0

mcollina:
You need to:

mqtt sub -t 'hello/+/hi' -h 'test.mosquitto.org' -v
mqtt pub -t 'hello/vikram/hi' -h 'test.mosquitto.org' -m 'from MQTT.js'

VikramTiwari:
Does this means I can subscribe on wildcard topics but can not publish to wildcard topics?

RangerMauve:
@VikramTiwari Yes. MQTT doesn't allow you to publish to topics with wildcards in them AFAIK.

You must publish to topics that match the wildcards.

anhldbk:
@mcollina This issue still persists. Here is the demo code:

```js
// test.js

var mqtt    = require('mqtt');

var client  = mqtt.connect('mqtt://test.mosquitto.org');
client.on('connect', function () {
  console.log('Connected')
  client.subscribe('#/:event/status') // <- using # will cause the problem below
});

client.on('offline', function () {
  console.log('Offline')
});
```

The output will be:
```text
$ node test.js
Connected
Offline
Connected
Offline
Connected
Offline
Connected
Offline
Connected
Offline
Connected
Offline
Connected
Offline
Connected
Offline
Connected
...
```

RangerMauve:
@anhldbk What makes you say that this is related to the wildcards? Is there an error you're seeing that would suggest that? Because it just looks like your socket is disconnecting and reconnecting for no particular reason.

[...]

PR #438, opened 2016-07-20
Modified subscribe() to validate topics before sending subscriptions to MQTT servers

anhldbk:
This is the PR addressing issue #172

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -10,6 +10,7 @@ var events = require('events'),
   Writable = require('readable-stream').Writable,
   inherits = require('inherits'),
   reInterval = require('reinterval'),
+  validations = require('./validations'),
   setImmediate = global.setImmediate || function (callback) {
     // works in node v0.8
     process.nextTick(callback);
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -403,6 +449,12 @@ MqttClient.prototype.subscribe = function () {
     callback = nop;
   }
 
+  invalidTopic = validateTopics(obj);
+  if ( null !== invalidTopic ) {
+    callback('Invalid topic ' + invalidTopic);
```
mcollina: Use a full error here, not a string.

lib/client.js
```diff
@@ -57,6 +57,51 @@ function storeAndSend (client, packet, cb) {
...
+ * @returns {String} If the topics is valid, returns null. Otherwise, returns the invalid one
+ */
+function validateTopics (topics) {
+  for (var i = 0; i < topics.length; i++) {
+    if ( !validateTopic(topics[i]) ) {
+      return topics[i];
+    }
+  }
+  return null;
+}
```
mcollina: It would be probably good to move these into another file, `client.js` is already too big.

test/abstract_client.js
```diff
@@ -255,6 +255,30 @@ module.exports = function (server, config) {
...
+
+    it('otherwise, should return an error (via callbacks)', function (done) {
+      var client = connect();
+      client.subscribe( ['#/event', 'event#', 'event+'], function (err) {
+        if (err) {
+          return done();
+        }
+        done(new Error('Validations do NOT work'));
+      });
+    });
```
mcollina: Can you please split this test into multiple cases? _all_ of those are invalid, but this test verifies that _just one_ is invalid.

test/abstract_client.js
```diff
@@ -255,6 +255,30 @@ module.exports = function (server, config) {
     });
   });
 
+  describe('Topic validations when subscribing', function () {
+
+    it('should be ok for well-formated topics', function (done) {
+      var client = connect();
+      client.subscribe( ['+', '+/event', 'event/+', '#', 'event/#'], function (err) {
```
mcollina: can you add also `a/b/c/d`?

sandro-k: and `a/+/c` - thanks

lib/client.js
anhldbk: Fixed! I've moved the implementations into `validations.js`

lib/client.js
anhldbk: Fixed

test/abstract_client.js
anhldbk: Tests are divided into smaller ones.

test/abstract_client.js
anhldbk: @mcollina , @sandro-k : All tests are included in.

test/abstract_client.js
```diff
@@ -255,6 +255,80 @@ module.exports = function (server, config) {
...
+      client.subscribe( 'system/+/#/event', function (err) {
+        if (err) {
+          return done();
+        }
+        done(new Error('Validations do NOT work'));
+      });
+    });
+
+
+
```
mcollina: Can you please remove the double newlines here and before?

anhldbk: I removed them :)

mcollina:
This is ok for me. Any other opinion?

anhldbk:
@mcollina So yes?

Issue #441, opened 2016-08-01
About function connect()


anhldbk:
I have the familiar code:

```js
var mqtt    = require('mqtt');

var options = {
  username: 'ping',
  password: '123456',
  host: '127.0.0.1',
  port: 1883,
}

var client  = mqtt.connect(options);
client.on('connect', function () {
  console.log('Connected')
});

client.on('error', function(){
  console.log('Error')
})

client.on('offline', function () {
  console.log('Offline')
});
```

My question is: How can we handle errors when invoking function `connect()`? 

Currently, here are some kind of errors & what printed out when I executed the code:

| Error | Event fired |
|-------|-------------|
|   Server is not available    |      `offline`       |
|    Invalid username or password    |     `error`        |

I wonder if we should add a callback to `connect()` to handle such failures? So we can have:

```js
var client  = mqtt.connect(options, function(err, result){
    if(err) {
        // something goes wrong
    }
})
```

@mcollina What do you think?

anhldbk:
And if the client goes offline, I will receive multiple offline events:

```sh
$ node test2.js
Connected
Offline
Offline
Offline
Offline
...
```

Do you think we should notify client once? That means: event `offline` will be fired once when the client first disconnects from the server.

mcollina:
offline is currently emitted even if a connection fails. I'm ok in a PR to address this.

anhldbk:
@mcollina What's about `connect` with callbacks?

anhldbk:
@mcollina Ping ping ping

mcollina:
I think the current interface is ok, we should just avoid emitting out `'offline'` every time it fails to connect. Maybe we should even add a limit, just to avoid a wait-forever loop.

The reason for _not_ having a callback is that it will introduce two ways of handling errors: one at the first connection, and one afterwards. Maybe that's what we should be talking about, as it might be the underlining issue at API-level.

Also, the current state "is it connected/disconnected/connecting/waiting" is not currently well exposed.

Any other opinion?

anhldbk:
@mcollina  you said about `two ways of handling errors`. Why don't we think it as `handling 2 types of errors`, including:
- `connection errors`: invalid username/password, offline service...
- `operation errors`:  any other one

mcollina:
if we specify a callback, what happens if both a callback and a `'error'` handler are specified?

RangerMauve:
[...]

PR #444, opened 2016-08-03
Fix for issue #441: Offline events should be fired once

anhldbk:


lib/client.js
```diff
@@ -32,10 +32,13 @@ function defaultId () {
 function sendPacket (client, packet, cb) {
   try {
     var buf = mqttPacket.generate(packet);
-    if (!client.stream.write(buf) && cb) {
-      client.stream.once('drain', cb);
-    } else if (cb) {
+
+    client.emit('packetsend', packet);
```
mcollina: why are you emitting `'packetsend'` here?
Can you send another PR for `'packetsend'`  and `'packetreceive'`, so it's easier to discuss?

mcollina:
Can you please add an explanation on the README of the various status flags? And maybe an explanation of the state machine?

anhldbk:
@mcollina Ok, I'm gonna update the README.

Btw, you said: 
```text
why are you emitting 'packetsend' here?
Can you send another PR for 'packetsend'  and 'packetreceive', so it's easier to discuss?
```
Still I really don't know with the diff about `emitting packetsend`. I looked at the code [Current client.js, line #274](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L274) and found no difference.

mcollina:
@anhldbk github diff messed things up, and that line showed up as added. I checked and I removed my comment.

anhldbk:
@mcollina I updated the README. Is that ok?

mcollina:
All good for me.

I think this is semver-minor. Anybody thinks it's semver-major?

RangerMauve:
I think it's minor. I doubt many people would have used the "offline" event in a way where it needs to be called each time a reconnection is attempted.

mcollina:
Ok, to be minor, merging and releasing.

PR #451, opened 2016-08-22
Update README.md

simenkid:
Add descriptions of callback signature to client.publish(), client.subscribe(), and client.unsubscribe().
It would be more clear for readers.


mcollina:
Thanks!!!

PR #455, opened 2016-08-24
ensure client.end() returns client even if called twice

### LINKED ISSUE #453: client.end() sometimes returns `true`
[Here's the line of code.](https://github.com/mqttjs/MQTT.js/blob/1fdc85dbda2270aca21c29009e6fd0f91258d0bc/lib/client.js#L437)

The comment above the method specifies `@returns {MqttClient} this - for chaining`, but doesn't this occasional `Boolean` prevent you from safely chaining?

brandly:
fixes #453

Let me know what you think! I wasn't quite sure where to place this test, so I can move that if needed.

Thanks for this project! I've really enjoyed using it. 🌟

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -530,7 +530,7 @@ MqttClient.prototype.end = function (force, cb) {
   }
 
   if (this.disconnecting) {
-    return true;
+    return this;
   }
 
   this._clearReconnect();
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks!!

Issue #460, opened 2016-09-13
A poor connection can cause a single packet to be sent many time on reconnect. Multiple duplicated subscriptions will kill the AWS IoT connection and will not recover.


miotabbott:
In certain situations a packet from the client `outgoingStore` can rewrite the `outgoing` callback for the packet to reference the `outgoingStore` flushing code (an anonymous function in `storeDeliver`) recursively on each unack'd packet.

AWS IoT does not tolerate many duplicate ack'd subscription messages and will kill the connection when this happens.

Unit test and fix in the works.

mcollina:
Thanks, a PR is highly welcomed!

miotabbott:
As a fun bonus, the unit test exhibits issue #450

pguerin-miovision:
@mcollina Any idea of the typical turn around time to accept and publish a pull request like this once it's ready?  Trying to sort out some short term options if we think it could be a while before this makes its way through. Thanks in advance for any feedback!

mcollina:
@pguerin-miovision roughly days or at most a week, as you can see from past prs. We all do this in our free time: if you want you can help.

miotabbott:
Patch for master up #468.
Build bot for #467 passed with build 779.

YoDaMa:
closing this issue because looks like #468 and #470 respond to this.

PR #467, opened 2016-09-27
Stop flushing the stored messages stream when the connection dies,

miotabbott:
the stream can get flushed multiple times in parallel otherwise.
Fixes issue #460

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -122,7 +122,8 @@ function MqttClient (streamBuilder, options) {
     }
 
     this.connected = true;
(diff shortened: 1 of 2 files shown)


mcollina:
There are some nits to be fixed: https://travis-ci.org/mqttjs/MQTT.js/jobs/163202357#L475-L505.

Also, this is targeting v1. Can you also send one for v2?

mcollina:
Same comments for #460. I'm ok if you work just on one of them, and then we backport the commit to the other one.

PR #468, opened 2016-09-28
Stop flushing the stored messages stream when the connection dies,

miotabbott:
the stream can get flushed multiple times in parrallel otherwise.

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -252,5 +252,69 @@ describe('MqttClient', function () {
         }
       }, 2000)
     })
+
+    it.only('should not send the same packet multiple times on a flaky connection', function (done) {
+      this.timeout(3500)
+
+      var KILL_COUNT = 4
+      var killedConnections = 0
+      var subCallbackCount = 0
+      var subIds = {}
+      var client = mqtt.connect({
+        port: port + 46,
+        host: 'localhost',
+        connectTimeout: 350,
+        reconnectPeriod: 300
+      })
+
+      var server2 = new Server(function (client) {
+        client.on('connect', function (packet) {
+          if (packet.clientId === 'invalid') {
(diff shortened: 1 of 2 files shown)


test/client.js
```diff
@@ -252,5 +252,70 @@ describe('MqttClient', function () {
...
+            return true
+          } else {
+            return false
+          }
+        })
+        if (duplicateId) {
+          done(new Error('Multiple duplicate acked subscriptions received for messageId ' + duplicateId))
+        } else {
+          done()
+        }
```
mcollina: we should close the server before calling done.

test/client.js
```diff
@@ -252,5 +252,70 @@ describe('MqttClient', function () {
...
+        client.end(true)
+      }, 3000)
```
mcollina: I am really scared about timeouts in tests, as they are often a source of flakyness.
Can we remove it?

test/client.js
```diff
@@ -252,5 +252,70 @@ describe('MqttClient', function () {
...
+      setTimeout(function () {
+        var duplicateId
+        Object.keys(subIds).some(function (messageId) {
+          if (subIds[messageId] > 1) {
+            duplicateId = subIds[messageId]
+            return true
+          } else {
+            return false
+          }
+        })
```
mcollina: Why are you using `some`, if you are not storing the result in  `duplicateId`?

miotabbott:
@mcollina requested changes made.

mcollina:
Thanks, releasing soon. Feel free to update your PR for v1.

miotabbott:
Thanks! Unfortunately a mistake slipped in with the last change. #469 is up to fix it.

PR #469, opened 2016-09-30
Remove a .only on a test.

miotabbott:
A test was singled out when developing a previous commit. Restore the unit tests to normal operation.

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -253,7 +253,7 @@ describe('MqttClient', function () {
       }, 2000)
     })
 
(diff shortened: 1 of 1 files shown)


mcollina:
The test is not passing on my box :(. Any idea?

miotabbott:
Seems to be passing on the build machine. What sort of error do you get on your box?

mcollina:
duplicate message on your test :/

miotabbott:
That's no good. I will try to reproduce.

Issue #450, opened 2016-08-17
Client doesn't send queued packets in the correct order (after reconnect).


WilliamHolmes:
I've noticed If there is a network disconnect and messages (payloads) are published they get saved to client.queue (<array>)

I've checked the queue and these are in the correct order.

When the client reconnects, the payloads doesn't seem to have been published in the correct order.
ie. shift

In my case

mqtt v1.13.0
host: **wss**://my_server_hostname
client > queueQoSZero : false
Message payloads have qos: 1

**Disconnected** from the network
sent 1, then 2, then ..., then 9.
I checked the queue and the payloads are in the correct order
![image](https://cloud.githubusercontent.com/assets/5538260/17746380/b6dbb6e0-64a7-11e6-9525-51e951de0205.png)

**Reconnect** to the network
The client.publish callbacks print ...
![image](https://cloud.githubusercontent.com/assets/5538260/17746109/85566454-64a6-11e6-935b-9bc802edd312.png)

How can I check that the payloads are indeed being requested in the correct order?

mcollina:
Which broker are you using? In any case, try version 1.14: I've released it today, with some improvement on the offline mode.
I don't have time to help debug this, but I'm happy to review a PR.

iharyadi:
Hi I am using RabitMQ version 3.6.5.  I am seeing the same issue with 1.14.1.  

Anything I can help to debug this issue?

mcollina:
Use a debugger, and understand why the messages are pulled off-order, and then send a pull request.

iharyadi:
Thanks for a quick response.  

I am just learning node and mqtt.  This is literally my second hours playing with it.  I try to install node-inspector.  It crashes for 6.4.  There is a bug there.  I will get my debugging environment any time soon.

For now,  here is what I did.  I change my broker to mosquitto and use qos2.  It seems that I can get the  message in ordered with qos2.

I need qos2 in the first place any way.  Just FYI for newbie like me,  RabbitMQ does not support qos2.

PR #470, opened 2016-10-01
Avoid storing subscribes in the outgoingStore

mcollina:
#460 #468 #469 #450 are about a subscribe that goes multiple times to AWS IoT.

I reverted your fixes @miotabbott because they were not working on my machine, and multiple duplicated subscribes where happening. I changed the queue logic, so only  PUBLISH packets are stored in the `outgoingStore`.

@miotabbott can you please review this? Also everybody else :).

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -268,6 +268,7 @@ describe('MqttClient', function () {
       })
 
       var server2 = new Server(function (client) {
+        client.on('error', function () {})
         client.on('connect', function (packet) {
           if (packet.clientId === 'invalid') {
             client.connack({returnCode: 2})
--- a/test/client.js
+++ b/test/client.js
@@ -7,6 +7,7 @@ var path = require('path')
 var abstractClientTests = require('./abstract_client')
 var net = require('net')
 var eos = require('end-of-stream')
+var Connection = require('mqtt-connection')
 var Server = require('./server')
 var port = 9876
 var server
@@ -253,7 +254,7 @@ describe('MqttClient', function () {
       }, 2000)
     })
 
-    it('should not send the same packet multiple times on a flaky connection', function (done) {
+    it('should not send the same subcribe multiple times on a flaky connection', function (done) {
       this.timeout(3500)
 
       var KILL_COUNT = 4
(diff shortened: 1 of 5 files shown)


mcollina:
@sandro-k @ralphtheninja would you mind to please review this?

miotabbott:
This patch will not fix #460. AWS IoT will kill the MQTT connection if many duplicate acked packets of any type get to the broker.

mcollina:
@miotabbott can you send a PR on top of this branch that checks that for PUBLISH? queuing up subscribes was wrong in the first place.

miotabbott:
PR ready #472 
The test fails with current code.

test/client.js
```diff
@@ -316,5 +314,69 @@ describe('MqttClient', function () {
...
+        })
+
+        c.on('publish', function onPublish (packet) {
+          if (killedConnections < KILL_COUNT) {
+            // Kill the first few pub attempts to simulate a flaky connection
+            killedConnections++
+            c.destroy()
+
+            // to avoid receiving inflight messages
+            c.removeListener('publish', onPublish)
```
mcollina: @miotabbott this line was missing in your test.  Basically the server connection was destroyed, but the client never received a corresponding puback. From the client perspective, the publish was never acked, so it had all the rights to retry.

miotabbott: Thanks!
@mcollina How can a `publish` event emit for `c` when it is destroyed on line 357?

mcollina: because `c.destroy` destroys the stream/socket, but the packet has been already received and parsed, so it will be delivered synchronously.

Please verify on AWS IoT if everything works correctly with this fix.

sandro-k:
I have to say that the test is a bit difficult to read, it is not always clear what `var client` is referring to.

mcollina:
@miotabbott can you send me a PR against this branch to address the issue @sandro-k brought up?

miotabbott:
@sandro-k Sorry about the ugly test, copy pasta from the test above has failed me.
@mcollina PR #473 ready. Please let me know if there is anything I can do to make the unit test more clear.

sandro-k:
Don't be sorry for sending in a PR 👍

mcollina:
I would like to get this shipped.
Apart from the ugly test, might I merge?

mcollina:
Merging now.

PR #471, opened 2016-10-03
Set default message (bin/pub) as empty string in minimist's config object

dhruvin2910:
`mqtt publish topic 0` should publish "0" instead of "". But `mqtt publish topic -m 0` behaves as expected.

### DIFF
--- a/bin/pub.js
+++ b/bin/pub.js
@@ -74,7 +74,8 @@ function start (args) {
       host: 'localhost',
       qos: 0,
       retain: false,
-      topic: ''
+      topic: '',
+      message: ''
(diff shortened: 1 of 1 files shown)


dhruvin2910:
Consider the command `mqtt publish topic 0`.

Since args._.shift() returns 0 [here](https://github.com/mqttjs/MQTT.js/pull/471/files#diff-3370a9205ffac2b50540c4d843c6de57L121), and it is a falsy value, the message gets defaulted to "".

mcollina:
Thanks, I will release this with #470

PR #473, opened 2016-10-04
Update duplicate ack test to check for publish messages.

miotabbott:


mcollina:
Can you please do the same for both tests, we need to check both subscribes and publishes.

mcollina:
@miotabbott any update?

PR #478, opened 2016-10-23
Default to using port 443 for wss connections. Fixes #477

### LINKED ISSUE #477: node.js: wss protocol default to port 80 instead of 443
The node.js (non-browser) connection builder defaults to port 80 for wss connections, while the browser-equivalent sets the port to 443.
See https://github.com/mqttjs/MQTT.js/blob/master/lib/connect/ws.js#L19

Since mqttjs swallows connection errors, this may cause hard-to-debug connection issues.

Should we add similar code here for defaulting the port to 443 for wss connections?

kintel:
As discussed.

Note: The test framework isn't set up to test this, e.g. by mocking websocket-stream, so I've omitted tests for this minor fix, as extending the test framework is a bit out of scope.

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -16,9 +16,17 @@ function buildBuilder (client, opts) {
     protocol: 'mqtt'
   }
   var host = opts.hostname || 'localhost'
-  var port = String(opts.port || 80)
   var path = opts.path || '/'
-  var url = opts.protocol + '://' + host + ':' + port + path
+
+  if (!opts.port) {
+    if (opts.protocol === 'wss') {
+      opts.port = 443
+    } else {
+      opts.port = 80
+    }
+  }
+
+  var url = opts.protocol + '://' + host + ':' + opts.port + path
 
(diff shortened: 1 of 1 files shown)


lib/connect/ws.js
```diff
@@ -16,9 +16,17 @@ function buildBuilder (client, opts) {
     protocol: 'mqtt'
   }
   var host = opts.hostname || 'localhost'
-  var port = String(opts.port || 80)
   var path = opts.path || '/'
-  var url = opts.protocol + '://' + host + ':' + port + path
+
+  if (!opts.port) {
+    if (opts.protocol === 'wss') {
```
ralphtheninja: `opts.port = opts.protocol === 'wss' ? 443 : 80`

kintel: I preferred to keep the same style as for the function below, assuming it was written that way for a reason.

ralphtheninja: Kk. Just a suggestion :)

PR #485, opened 2016-11-04
add clientId and aliases as 'string' types

dpjanes:
If the clientId is a number (OpenSensors.io does this), the minimist library will interpret it as a number, causing an error when the code goes to use it.

I've added clientId and its aliases to the 'string' types so things will continue to work.

Note that minimist explicitly requires you to spell out all the strings, it won't figure it out from the alias this. I've contact them but it appears the project maybe orphaned or at least not on the front burner. 

Also not that minimist doesn't take an 'integer' parameter so that could possible be removed from the dictionary passed to minimist. However, I did not make this change and its presence is harmless.

### DIFF
--- a/bin/pub.js
+++ b/bin/pub.js
@@ -51,7 +51,7 @@ function multisend (args) {
 
 function start (args) {
   args = minimist(args, {
-    string: ['hostname', 'username', 'password', 'key', 'cert', 'ca', 'message'],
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks!!!!

mcollina:
@dpjanes can you please remove that `'integer'` thing? I think that was in a very old minimist release.

Issue #496, opened 2016-11-18
subscribe to empty array


mepdr:
Subscribing to an empty array causes an application disconnection and the subscribe callback is not called.
The testing app below prints: 
```
offline 
close
reconnect
```

```js
var mqtt = require('mqtt')
var c  = mqtt.connect('ws://test.mosquitto.org:8080/mqtt')

c.on("message", function(topic, payload) {
    console.log(topic);
});

c.on("reconnect", function() {
    console.log("reconnect");
});

c.on("close", function() {
    console.log("close");
});

c.on("offline", function() {
    console.log("offline");
});

c.on("error", function() {
    console.log("error");
});

c.subscribe([], {qos:1}, function (err, granted) {
    console.log(err, granted);
});
```

mcollina:
Thanks for reporting, would you like to send a PR to fix this issue?

johnzeng:
According to MQTT 3.1.1 protocol ,  The payload of a SUBSCRIBE packet MUST contain at least one Topic Filter/QoS pair. A SUBSCRIBE packet with no payload is a protocol violation.

mcollina:
I think we should error on the client side, before  sending a message to the server.

yohei1126:
Has this issue already closed?

mcollina:
Fixed by #497

mcollina:
Thanks!!!

PR #497, opened 2016-11-21
don't send empty sub list

johnzeng:
fix the issue of #496

### DIFF
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -343,6 +343,17 @@ module.exports = function (server, config) {
       })
     })
 
+    it('should return an error (via callbacks) for empty topic list', function (done) {
+      var client = connect()
+      client.subscribe([], function (err) {
+        client.end()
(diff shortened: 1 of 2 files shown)


mcollina:

johnzeng:
yes, let me check about it

johnzeng:
should return an error (via callbacks) for empty topic list

test/abstract_client.js
```diff
@@ -343,6 +343,17 @@ module.exports = function (server, config) {
       })
     })
 
+    it('should return an error (via callbacks) for empty topic list', function (done) {
+      var client = connect()
+      client.subscribe([], function (err) {
```
mcollina: we return an `Error` here.

johnzeng: ok.

anhldbk: @mcollina His test is ok, right?

PR #498, opened 2016-11-21
subscribe calls back synchronously if invalid topic

ralphtheninja:


lib/client.js
```diff
@@ -421,7 +421,9 @@ MqttClient.prototype.subscribe = function () {
 
   invalidTopic = validations.validateTopics(obj)
   if (invalidTopic !== null) {
-    callback(new Error('Invalid topic ' + invalidTopic))
+    setImmediate(function () {
+      callback(new Error('Invalid topic ' + invalidTopic))
+    })
```
mcollina: we can do `setImmediate(callback, new Error('Invalid topic ' + invalidTopic))`

ralphtheninja: Cool. I did not know this :) Updated and force pushed.

mcollina:
thanks!

PR #511, opened 2016-12-05
manage ws(s) options (proxy agent e.g.)

Manu3756:
feedback welcome
badly tested

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -34,7 +34,6 @@ var urlKeys = [
 'hash' ];
 
 function buildBuilder (client, opts) {
-  console.log(opts);
   var wsOpt = {
     protocol: 'mqtt'
   }
@@ -67,7 +66,6 @@ function buildBuilder (client, opts) {
       wsOpt[prop]=opts[prop];
     }
   })
-  console.log(wsOpt);
   return websocket(url, [wsOpt.protocol], wsOpt)
 }
 
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -3,35 +3,35 @@
 var websocket = require('websocket-stream')
 var _URL = require('url')
 var mqttProperties = [
-'keepalive',
-'reschedulePings',
-'clientId',
-'protocolId',
-'protocolVersion',
-'clean',
-'reconnectPeriod',
-'connectTimeout',
-'username',
-'password',
-'incomingStore',
-'outgoingStore',
-'queueQoSZero',
-'will',
-];
+  'keepalive',
+  'reschedulePings',
+  'clientId',
+  'protocolId',
+  'protocolVersion',
+  'clean',
+  'reconnectPeriod',
+  'connectTimeout',
+  'username',
+  'password',
+  'incomingStore',
+  'outgoingStore',
+  'queueQoSZero',
+  'will'
+]
 
 var urlKeys = [
-'href',
-'slashes',
-'path',
-'protocol',
-'auth',
-'hostname',
-'port',
-'host',
-'pathname',
-'search',
-'query',
-'hash' ];
+  'href',
+  'slashes',
+  'path',
+  'protocol',
+  'auth',
+  'hostname',
+  'port',
+  'host',
+  'pathname',
+  'search',
+  'query',
+  'hash' ]
 
 function buildBuilder (client, opts) {
   var wsOpt = {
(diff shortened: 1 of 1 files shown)


mcollina:
Linting is failing. We use http://npm.im/standard.

lib/connect/ws.js
```diff
@@ -2,14 +2,36 @@
...
+'path',
+'protocol',
+'auth',
+'hostname',
+'port',
+'host',
+'pathname',
+'search',
+'query',
+'hash' ];
```
mcollina: why the need of all of this?

Manu3756: there are two ways to i guess: 
-set a list of options we want to support 

-or support everything and remove the overhead

mcollina: Ok, this is the first, correct?

mcollina:
cc @ralphtheninja @sandro-k can you please have a look?

lib/connect/ws.js
Manu3756: no this is the second

Manu3756: Maybe there is a 3. way. 

define a option key like
` wsOptions: {ws(s) Option Object}`
 its less code but you need to know that its done this way.

mcollina: can you please articulate that better?

Manu3756: I'll try sorry.

the 3. way would be to add a Options key to the list of defined once:

> options is the client connection options (see: the connect packet). Defaults:
keepalive: 10 seconds, set to 0 to disable
reschedulePings: reschedule ping messages after sending packets (default true)
clientId: 'mqttjs_' + Math.random().toString(16).substr(2, 8)
protocolId: 'MQTT'
protocolVersion: 4
clean: true, set to false to receive QoS 1 and 2 messages while offline
reconnectPeriod: 1000 milliseconds, interval between two reconnections
connectTimeout: 30 * 1000 milliseconds, time to wait before a CONNACK is received
username: the username required by your broker, if any
password: the password required by your broker, if any
incomingStore: a Store for the incoming packets
outgoingStore: a Store for the outgoing packets
queueQoSZero: if connection is broken, queue outgoing QoS zero messages (default true)
will: a message that will sent by the broker automatically when the client disconnect badly. The format is:
topic: the topic to publish
payload: the message to publish
qos: the QoS
retain: the retain flag

because we only have one option object and need to handle options for mqtt and for websockets 

so we could add a key to the given list like 

> wsOptions: options for the websocket connection 

then we can simply handover this wsOptions object to the websocket constructor and don't need list options we want to support (way 1) or remove the overhead.(way 2)
the overhead is everything in the options object we need for mqtt but not for the websocket connection.

mcollina:
Can you please include the example with the proxy in the examples folder?

lib/connect/ws.js
mcollina: Thanks

lib/connect/ws.js
```diff
@@ -32,15 +40,14 @@ function buildBuilder (client, opts) {
...
-        wsOpt[prop] = opts[prop]
-      }
-    })
-  }
-
-  return websocket(url, wsOpt)
+  /*delete some mqtt overhead*/
+  mqttProperties.forEach(function (prop) {
+    if (opts.hasOwnProperty(prop)) {
+      delete opts[prop];
```
ralphtheninja: I'm not sure I understand the reason for this. Is this because you want to prevent some options to be set? (i.e. the `mqttProperties` list)

Manu3756: It's bad anyway because wsOpt is a reference to opts and no "real" object. I changed this in a later commit in this branche. I'll do this because there are some options with the same key in mqtt and wss constructur (e.g. protocolVersion) 
If I don't delet them the websocket want to use this propertie too.

lib/connect/ws.js
```diff
@@ -41,12 +56,18 @@ function buildBuilder (client, opts) {
...
+    var propInKeys = false;
+    mqttProperties.concat(urlKeys).forEach(function(key){
+        if(prop === key){
+          propInKeys = true;
+        }
+    });
+    if (!propInKeys){
+      wsOpt[prop]=opts[prop];
     }
   })
```
ralphtheninja: What about doing something like:

```js
var propInKeys = mqttProperties.concat(urlKeys).some(function (key) {
  return prop === key
})
if (!propInKeys) { // .. }
```

lib/connect/ws.js
```diff
@@ -3,35 +3,35 @@
...
+  'path',
+  'protocol',
+  'auth',
+  'hostname',
+  'port',
+  'host',
+  'pathname',
+  'search',
+  'query',
+  'hash' ]
```
ralphtheninja: `]` on the next line

lib/connect/ws.js
Manu3756: Yes, would be better. The Problem that I can't set the property e.g. protocolVersion for websockets is still not solved because we can't decide if the mqtt protocolVersion or the websocket protocolVersion is ment. 
Have a look at #512. There we can do stuff like that.

mcollina:
@Manu3756 You might be trying to fix two issues at the same time. Why can't we just move the `agent` property?

After the bug is solved, then we can fix the "protection" issue.

Manu3756:
@mcollina 
We have one issue. We can't handle websocket options. If we just add the agent to the list we support 1 option more then before. And only for wss and not for ws. 
It's up to you anyway if you want to solv the "agent in wss" bug and wait for the "agent in ws", "protocolVersion in wss", "protocolVersion in ws" ... bug its okay.

mcollina:
@Manu3756 I am still not sure what problem are you trying to solve, and thus I am unable to have a preferred opinion. I would like some option that:

a. is neat and do not involve big arrays with all the properties
b. support `agent` both in `ws` and `wss`
c. do not break backward compatibility, especially for `wss`

Manu3756:
@mcollina 
Now I got your problem. 
For a and b have a look at #512 
but in #512 is no guarantee that it is 100% backward compatibil. 
I'll check how to change #512 that it fits to c) if you like.

mcollina:
That will be perfect, thanks.

PR #512, opened 2016-12-06
Ws options

### LINKED ISSUE #452: Is a HTTP_PROXY supported?
Hi,

How does this work behind locked-down networks i.e. where a HTTP_PROXY is the only means to reach the internet?

Is a HTTP_PROXY supported?

Thanks

Manu3756:
dealing with websocket options by adding the key  wsOptions (like agent for proxys) to the standard options object.

maybe smarter then  
#511 

fix  #452 for (secure)websockets

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,8 @@ The arguments are:
 * `streamBuilder` is a function that returns a subclass of the `Stream` class that supports
 the `connect` event. Typically a `net.Socket`.
 * `options` is the client connection options (see: the [connect packet](https://github.com/mcollina/mqtt-packet#connect)). Defaults:
-  * `wsOptions`: is the webSocket connection options. Default is `{}`.
+  * `wsOptions`: is the WebSocket connection options. Default is `{}`. 
+	 It's specific for WebSockets. For possible options have a look at: https://github.com/websockets/ws. 
   * `keepalive`: `10` seconds, set to `0` to disable
   * `reschedulePings`: reschedule ping messages after sending packets (default `true`)
   * `clientId`: `'mqttjs_' + Math.random().toString(16).substr(2, 8)`
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ The arguments are:
 the `connect` event. Typically a `net.Socket`.
 * `options` is the client connection options (see: the [connect packet](https://github.com/mcollina/mqtt-packet#connect)). Defaults:
   * `wsOptions`: is the WebSocket connection options. Default is `{}`. 
-	 It's specific for WebSockets. For possible options have a look at: https://github.com/websockets/ws. 
+	 It's specific for WebSockets. For possible options have a look at: https://github.com/websockets/ws/blob/master/doc/ws.md. 
   * `keepalive`: `10` seconds, set to `0` to disable
   * `reschedulePings`: reschedule ping messages after sending packets (default `true`)
   * `clientId`: `'mqttjs_' + Math.random().toString(16).substr(2, 8)`
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ The arguments are:
 * `streamBuilder` is a function that returns a subclass of the `Stream` class that supports
 the `connect` event. Typically a `net.Socket`.
 * `options` is the client connection options (see: the [connect packet](https://github.com/mcollina/mqtt-packet#connect)). Defaults:
+  * `wsOptions`: is the webSocket connection options. Default is `{}`.
   * `keepalive`: `10` seconds, set to `0` to disable
   * `reschedulePings`: reschedule ping messages after sending packets (default `true`)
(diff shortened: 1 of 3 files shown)


lib/connect/ws.js
```diff
@@ -2,14 +2,6 @@
...
 var websocket = require('websocket-stream')
 var _URL = require('url')
-var wssProperties = [
-  'rejectUnauthorized',
-  'ca',
-  'cert',
-  'key',
-  'pfx',
-  'passphrase'
-]
```
mcollina: I don't understand why we can't just add agent here.

Manu3756: It's, possible but there are many more options for a websocket 

>  * @param {String} address The URL to which to connect
 * @param {String[]} protocols The list of subprotocols
 * @param {Object} options Connection options
 * @param {String} option.protocol Value of the `Sec-WebSocket-Protocol` header
 * @param {(Boolean|Object)} options.perMessageDeflate Enable/disable permessage-deflate
 * @param {String} options.localAddress Local interface to bind for network connections
 * @param {Number} options.protocolVersion Value of the `Sec-WebSocket-Version` header
 * @param {Object} options.headers An object containing request headers
 * @param {String} options.origin Value of the `Sec-WebSocket-Version` header
 * @param {http.Agent} options.agent Use the specified Agent
 * @param {String} options.host Value of the `Host` header
 * @param {Function} options.checkServerIdentity A function to validate the server hostname
 * @param {Boolean} options.rejectUnauthorized Verify or not the server certificate
 * @param {String} options.passphrase The passphrase for the private key or pfx
 * @param {String} options.ciphers The ciphers to use or exclude
 * @param {(String|String[]|Buffer|Buffer[])} options.cert The certificate key
 * @param {(String|String[]|Buffer|Buffer[])} options.key The private key
 * @param {(String|Buffer)} options.pfx The private key, certificate, and CA certs
 * @param {(String|String[]|Buffer|Buffer[])} options.ca Trusted certificates
(https://github.com/websockets/ws/blob/master/lib/WebSocket.js)

mcollina:
cc @ralphtheninja this is the same of #511

Manu3756:
@mcollina @ralphtheninja 
pls have a look at the last commit

mcollina:
@Manu3756 this seems ok to me. Would you mind to add docs in the README?

@mqttjs/mqtt-js what do you think?

lib/connect/ws.js
```diff
@@ -24,11 +34,15 @@ function buildBuilder (client, opts) {
     wsOpt.protocol = 'mqttv3.1'
   }
 
-  if (opts.wsOptions) {
-    return websocket(url, [wsOpt.protocol], opts.wsOptions)
-  } else {
-    return websocket(url, wsOpt)
+  if (opts.protocol === 'wss') {
```
ralphtheninja: This feels like a common use case to me. I'd do:

```js
var select = require('select-keys')
opts.wsOptions = select(opts, wssProperties)
```

See https://github.com/mmalecki/select-keys/blob/master/lib/select-keys.js

ralphtheninja: Maybe this doesn't play nice with how `opts.wsOptions` is set in the rest of the code.

lib/connect/ws.js
```diff
@@ -36,7 +36,7 @@ function buildBuilder (client, opts) {
 
   if (opts.protocol === 'wss') {
     wssProperties.forEach(function (prop) {
-      if (opts.hasOwnProperty(prop)) {
+      if (opts.hasOwnProperty(prop) && !opts.wsOptions.hasOwnProperty(prop)) {
```
ralphtheninja: Oh okay, forget my previous comment. If this is really needed then `select-keys` might not be useful. Unless this step can be done first.

Manu3756:
@mcollina updated Readme pls have a look

README.md
```diff
@@ -187,6 +187,7 @@ The arguments are:
 * `streamBuilder` is a function that returns a subclass of the `Stream` class that supports
 the `connect` event. Typically a `net.Socket`.
 * `options` is the client connection options (see: the [connect packet](https://github.com/mcollina/mqtt-packet#connect)). Defaults:
+  * `wsOptions`: is the webSocket connection options. Default is `{}`.
```
mcollina: can you call it `WebSocket`? and mention it as _specific_ for WebSocket?

mcollina: Can you also mentions _all_ the possible options for WebSockets? Otherwise it is not really helpful and it would create confusion (even in another section underneath).

mcollina:
any updates?

Manu3756:
@mcollina Sorry was very busy... next try

Manu3756:
No, I think this should be a new PR?
I'm not sure if it's needed on browser side.

mcollina:
@Manu3756 no problem. My bad, the PR is fine this way.

Issue #392, opened 2016-03-24
In browser, mqtts:// gets replaced with ws:// rather than wss://


ksiegel:
When passing in a `mqtts://` URL in the browser, `connect/index.js` realizes that mqtts is unsupported and falls back to the first available protocol: `ws`

I'd like to see some logic that would replace mqtts with wss when mqtts is unavailable. 

I can submit a PR but would like some feedback first!

Code in question: 
https://github.com/mqttjs/MQTT.js/blob/master/lib/connect/index.js#L105

Something along the lines of:

``` js
  if (!protocols[opts.protocol]) {
    if(opts.protocol === 'mqtts' && 'function' === typeof protocols.wss){
      // mqtts:// becomes wss://
      opts.protocol = 'wss';
    } else {
      // otherwise default to trying first available protocol
      opts.protocol = protocolList.filter(function (key) {
        return 'function' === typeof protocols[key];
      })[0];
    }
  }
```

Thanks!

mcollina:
Yes, please send a PR! However you might want to overhaul that logic completely. The whole protocolList thing was used to to avoid having those kind of ifs, and it fails the purpose. Maybe a simple regexp if we are in a browser?

mcollina:
Closing for no activity.

mbrandau:
Seems like this problem was fixed in #518. Is there a reason why this issue is still open? @mcollina @ksiegel

PR #518, opened 2016-12-29
Use secure protocol when requested. 

wcatron:
Current issue: #392 

When requesting the `mqtts` protocol in the browser `ws` is used instead of `wss`. I attempted to solve the issue with desired feedback in issue #392. 

Solution: Check if requesting a secure protocol then skip any insecure protocols in the filter. I chose this because it still relied on the filter. The `isSecure` constant is very clear. The only thing funny is the mod I used to determine if the key was secure. I'm open to suggestions but this passes all the tests and works with my code. I took a look at creating a test case, but I don't think I'm familiar enough with your setup to do that.

Alternative solutions:
- reversing protocolList
- if statement like proposed in issue

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -105,8 +105,13 @@ function connect (brokerUrl, opts) {
   }
 
   if (!protocols[opts.protocol]) {
-    opts.protocol = protocolList.filter(function (key) {
-      return typeof protocols[key] === 'function'
+    const isSecure = (['mqtts', 'wss'].indexOf(opts.protocol) !== -1)
+    opts.protocol = protocolList.filter(function (key, index) {
+      if (isSecure && index % 2 === 0) {
+        // Skip insecure protocols when requesting a secure one.
+        return false
+      }
+      return (typeof protocols[key] === 'function')
     })[0]
   }
 
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -105,7 +105,7 @@ function connect (brokerUrl, opts) {
   }
 
   if (!protocols[opts.protocol]) {
-    const isSecure = (['mqtts', 'wss'].indexOf(opts.protocol) !== -1)
+    const isSecure = ['mqtts', 'wss'].indexOf(opts.protocol) !== -1
     opts.protocol = protocolList.filter(function (key, index) {
       if (isSecure && index % 2 === 0) {
         // Skip insecure protocols when requesting a secure one.
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -4,7 +4,6 @@ var MqttClient = require('../client')
 var url = require('url')
 var xtend = require('xtend')
 var protocols = {}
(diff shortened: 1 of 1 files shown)


lib/connect/index.js
```diff
@@ -105,8 +105,13 @@ function connect (brokerUrl, opts) {
   }
 
   if (!protocols[opts.protocol]) {
-    opts.protocol = protocolList.filter(function (key) {
-      return typeof protocols[key] === 'function'
+    const isSecure = (['mqtts', 'wss'].indexOf(opts.protocol) !== -1)
```
mcollina: there is no need for ()

mcollina:
Folks, any opinion on this? @ralphtheninja @sandro-k?

lib/connect/index.js
```diff
@@ -105,8 +105,13 @@ function connect (brokerUrl, opts) {
...
+    opts.protocol = protocolList.filter(function (key, index) {
+      if (isSecure && index % 2 === 0) {
```
ralphtheninja: I don't like this dependency on order in the protocol array. If the order changes in the array or other protocols are added this might break. Maybe have two arrays, one for secure protocols and one for insecure protocols, which in turn protocolList is built from.

ralphtheninja: And the `index % 2` could instead be a check against the insecure protocols array :)

ralphtheninja:
> The only thing funny is the mod I used to determine if the key was secure.

Agreed :)

lib/connect/index.js
wcatron: One of my goals was to not disrupt the existing code - might have taken it too far, usually `%`ing is a red flag for that :) . @mcollina requested in #392 that we not use too many ifs which is why I went with my solution. So yes we could do this:

```js
// Define secureProtocolList and protocolList above...
  if (!protocols[opts.protocol]) {
    const isSecure = secureProtocolList.indexOf(opts.protocol) !== -1
    if (isSecure) {
      opts.protocol = secureProtocolList.filter(function (key, index) {
        return (typeof protocols[key] === 'function')
      })[0]
    } else {
      opts.protocol = protocolList.filter(function (key, index) {
        return (typeof protocols[key] === 'function')
      })[0]
    }
  }
```

Issues: we have a `protocols` object a `protocolList` array and a `secureProtocolList` array. Two of which are only used in this one block of code.

or I just pushed a solution that instead makes the dependency much more obvious and removes the `protocolList`. This block was the only part that used the `protocolList` so it makes sense to just include it directly and no one will go making changes without looking inside the filter function.

There was always a requirement on the order of the array so I don't think it's an issue that it has to go insecure, secure, since before it had to be in that order too (otherwise I would have just changed the order). The problem is with the list being defined so far away it could have been altered too easily. My last commit fixes that :)

mcollina:
@wcatron we can even do a slightly bigger refactoring and remove the need for that array altogether.

Issue #509, opened 2016-12-04
Connect on event


chimit:
Sorry for the stupid question, but how can I initiate a connection on another event? For example, on button push? I'm using MQTT.js in the browser.

mcollina:
You can't. you create the connection when the button is clicked.

chimit:
Sorry, @mcollina, maybe I'm not clear enough. I want to bind the connection to some event like this:
```
const client = mqtt.createClient({
    host: 'abc',
    port: 8883,
    protocol: 'mqtts',
    username: 'abc',
    password: 'abc',
});

$('#connect').click(function() {
    client.connect();
```
Is it still impossible?

mcollina:
Using the current API, yes.

I'm happy to discuss a semver-major contribution on this, if you are willing to help.

PR #519, opened 2016-12-30
Client connection delaying

4rzael:
As stated in #509, the connection is currently made in the Client constructor.

It makes codes like this one impossible :
 
```js
const client = mqtt.createClient({
    host: 'abc',
    port: 8883,
    protocol: 'mqtts',
    username: 'abc',
    password: 'abc',
});

$('#connect').click(function() {
    client.startConnection();
}
```

This pull request adds : 
* An option in the Client constructor : `delayConnection` (boolean). Defaults to `false`.
When set to true, it will ask the constructor not to start the connection automatically.
* A public method : `startConnection()` : When `delayConnection` has been passed to the constructor, a call to this function starts the connection

This is a minor change in the API, as it is backward-compatible


README.md
```diff
@@ -336,6 +338,13 @@ Unsubscribe from a topic or topics
 * `callback` - `function (err)`, fired on unsuback. An error occurs if client is disconnecting.
 
 -------------------------------------------------------
+<a name="startConnection"></a>
+### mqtt.Client#startConnection()
+
+Manually starts the connection.
+Use it when `delayConnection` has been passed to the constructor.
+
+-------------------------------------------------------
```
mcollina: how about we call it `start`? and maybe call the option `start`? I think it might play better with the rest of the API, specifically about `end()`.

lib/client.js
```diff
@@ -229,6 +250,7 @@ MqttClient.prototype._setupStream = function () {
   this._clearReconnect()
 
   this.stream = this.streamBuilder(this)
+  this.streamIsSetup = true
```
mcollina: how about calling it `_started`?

test/client.js
```diff
@@ -379,4 +379,32 @@ describe('MqttClient', function () {
...
+    it('should not connect on creation if the argument delayConnection is set to true', function (done) {
+      this.timeout(1000)
+
+      var client = mqtt.connect({ port: port, host: 'localhost', keepalive: 1, delayConnection: true })
+
+      setTimeout(function () {
+        client.streamIsSetup.should.equal(false)
+        client.connected.should.equal(false)
+        done()
+      }, 500)
```
mcollina: can you reduce this?

test/client.js
```diff
@@ -379,4 +379,32 @@ describe('MqttClient', function () {
...
+    })
+
+    it('should connect when the startConnection() function is called', function (done) {
+      this.timeout(1000)
+
+      var client = mqtt.connect({ port: port, host: 'localhost', keepalive: 1, delayConnection: true })
+
+      client.startConnection()
```
mcollina: Also, can you verify that _before_ that `startConnection()`  does not automatically connect to the server?

4rzael: I don't really understand this comment.

That's what I tested in the precedent test (I don't know if it's done correctly, though).

Should I do it in this one too ?

4rzael:
I changed what you asked me to (the names are way better this way), excepted this : 

> Also, can you verify that before that startConnection()  does not automatically connect to the server?

because I didn't really understood what you wanted.

lib/client.js
```diff
@@ -209,11 +213,28 @@ function MqttClient (streamBuilder, options) {
 
   events.EventEmitter.call(this)
 
-  this._setupStream()
+  if (this.options.start) {
+    this._setupStream()
+  }
```
mcollina: you can just call `this.start()` here.

mcollina:
What I mean is that we need to test that we should have a test that:

1. create a client with `start: false`  in the options
2. add an event handler into the server, like https://github.com/4rzael/MQTT.js/blob/7659a9f57b3de87d1851ec9150529b1270455b7d/test/client.js#L155. In the handler we fail the test if the client connects.
3. after 200ms (or similar), it removes the handler and calls start(). It also adds a new handler to check if there is in fact a client.

mcollina:
Any updates on this?

mcollina:
Closing, feel free to reopen.

PR #526, opened 2017-01-20
add a method to get the last id

### LINKED ISSUE #525: How to get message id
In my project, I need a way to get packet id for each packet because of the following situation.
Can I get a packet id for each packet?

* Some MQTT broker respond result of client's action asuch as publish/subscribe to special topic.
* The broker usually send packet id that is sent by the client so that the client can check result for each packet.

* Because of the above situation, the client save pakcet id locally so that the client can check result later.
* When the client receives result of the action from the broker, the client match packet id from the broker and packet id saved locally to check which actions is successful or not.

yohei1126:
fixes #525

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -928,4 +928,11 @@ MqttClient.prototype._nextId = function () {
   return id
 }
 
+/**
+ * _nextId
+ */
+MqttClient.prototype.getLastId = function () {
+  return (this.nextId === 1) ? 65535 : (this.nextId - 1)
+}
+
 module.exports = MqttClient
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -920,12 +921,19 @@ MqttClient.prototype._handlePubrel = function (packet, callback) {
...
-  return id
+  return this.lastId
+}
+
+/**
+ * getLastMessageId
+ */
+MqttClient.prototype.getLastMessageId = function () {
+  return this.lastId
 }
```
mcollina: why do not we just compute this from `nextId`?

yohei1126: sure. I will try it.

PR #527, opened 2017-01-21
add a function to get last id

### LINKED ISSUE #525: How to get message id

yohei1126:
fixes #525

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ See `mqtt help <command>` for the command help.
   * <a href="#handleMessage"><code>mqtt.Client#<b>handleMessage()</b></code></a>
   * <a href="#connected"><code>mqtt.Client#<b>connected</b></code></a>
   * <a href="#reconnecting"><code>mqtt.Client#<b>reconnecting</b></code></a>
-  * <a href="#getLastId"><code>mqtt.Client#<b>getLastId</b></code></a>
+  * <a href="#getLastMessageId"><code>mqtt.Client#<b>getLastMessageId</b></code></a>
   * <a href="#store"><code>mqtt.<b>Store()</b></code></a>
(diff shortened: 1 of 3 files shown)


mcollina:
Awesome! Can you add it to the README as well?

yohei1126:
@mcollina add the description for the function. please check it.

mcollina:
@yohei1126 while reviewing, it seems better to have `getLastMessageId()`  instead of `getLastId()`,  as it is not super clear in the docs.

Also, you should mention this is for sent messages only.

yohei1126:
please check it.

ogis-onishi:
The following type definiton for TypeScript need to be updated. I wll send pq as well. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/mqtt/index.d.ts

yohei1126:
The following type definiton for TypeScript need to be updated. I wll send pq as well.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/mqtt/index.d.ts

yohei1126:
sending pq now. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14179

yohei1126:
The pull request is approved. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14179#pullrequestreview-17859178

Waiting for the npm module @types/mqtt get updated.

PR #530, opened 2017-01-28
Optimization & new security test

roccomuso:
- Setting Standard mocha env.
- New nsp pre-test.


package.json
```diff
@@ -70,9 +70,10 @@
   },
   "devDependencies": {
     "browserify": "^13.1.1",
-    "mocha": "*",
     "mkdirp": "^0.5.1",
+    "mocha": "*",
```
mcollina: while we are at it, can we limit this with ^?

package.json
```diff
@@ -86,11 +87,8 @@
     "zuul-ngrok": "^4.0.0"
   },
   "standard": {
-    "globals": [
-      "it",
-      "describe",
-      "beforeEach",
-      "afterEach"
+    "env": [
+      "mocha"
```
mcollina: awesome!

package.json
```diff
@@ -20,7 +20,7 @@
   "main": "mqtt.js",
   "scripts": {
     "test": "mocha --bail",
-    "pretest": "standard",
+    "pretest": "standard && nsp",
```
mcollina: we can also add snazzy here, it's nice!

PR #539, opened 2017-02-12
Add transformWsUrl option, and fix some travis/browser test issues

### LINKED ISSUE #535: Add an urlSigner (func) option to connect to pass to the ws streambuilder
For AWS IoT wss connections urls require Sigv4 signing. These urls seemingly expire, so eventually reconnect doesn't work and you need a fresh signature. 

You can see here usage of an old version of mqtt.js: https://github.com/aws/aws-iot-device-sdk-js/blob/master/device/index.js#L502-L521

I'm using mqtt.js 2.x directly and my own url signer implementation. 

I will possibly get around to a pull request but first would like to see how do people feel about adding an `urlSigner` function to the connection options, passed along to the stream builder[s]?

### LINKED ISSUE #537: Travis build script is broken
> - test $SAUCE_USERNAME && npm run sauce-test || echo 'not running on saucelabs'

Note the above will always return 0 yes?

eg.
```zsh
➜  MQTT.js git:(master) ✗ python -c "import sys; sys.exit(1)"
➜  MQTT.js git:(master) ✗ echo $?
1
➜  MQTT.js git:(master) ✗ export SAUCE_USERNAME=wink
➜  MQTT.js git:(master) ✗ test $SAUCE_USERNAME && python -c "import sys; sys.exit(1)" || echo 'not running on saucelabs'
not running on saucelabs
➜  MQTT.js git:(master) ✗ echo $?
0
```
Please see: https://travis-ci.org/mqttjs/MQTT.js/jobs/196752870#L842-L870

### LINKED ISSUE #538: Browser tests broken
re: #537 

See saucelabs results:
image

I tried running `npm run browser-test` locally, and I got some errors, due to this:
https://github.com/mqttjs/MQTT.js/blob/master/test/browser/server.js#L106-L107

```js
function start (startPort, done) {
  var server = http.createServer()
  var wss = new WebSocketServer({server: server})

  wss.on('connection', function (ws) {
    var stream, connection
    if (ws.protocol !== 'mqttv3.1') {
      return ws.end()
    }
```

I'm not sure if it's due to some of my local copy refactoring, but one thing is for sure, the `ws` object has no end property. It *does* have a close method.

sublimator:
Testing Travis and saucelabs

Fixes #538 
Fixes #535 
Fixes #537

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,8 +188,8 @@ The arguments are:
 * `streamBuilder` is a function that returns a subclass of the `Stream` class that supports
 the `connect` event. Typically a `net.Socket`.
 * `options` is the client connection options (see: the [connect packet](https://github.com/mcollina/mqtt-packet#connect)). Defaults:
-  * `wsOptions`: is the WebSocket connection options. Default is `{}`. 
-	 It's specific for WebSockets. For possible options have a look at: https://github.com/websockets/ws/blob/master/doc/ws.md. 
+  * `wsOptions`: is the WebSocket connection options. Default is `{}`.
+     It's specific for WebSockets. For possible options have a look at: https://github.com/websockets/ws/blob/master/doc/ws.md.
   * `keepalive`: `10` seconds, set to `0` to disable
   * `reschedulePings`: reschedule ping messages after sending packets (default `true`)
   * `clientId`: `'mqttjs_' + Math.random().toString(16).substr(2, 8)`
@@ -212,6 +212,9 @@ the `connect` event. Typically a `net.Socket`.
     * `payload`: the message to publish
     * `qos`: the QoS
     * `retain`: the retain flag
+  * `transformWsUrl` : optional `(url, options, client) => url` function
+        For ws/wss protocols only. Can be used to implement signing
+        urls which upon reconnect can have become expired.
 
 In case mqtts (mqtt over tls) is required, the `options` object is
 passed through to
@@ -237,9 +240,9 @@ version 1.3 and 1.4 works fine without those.
 
 `function (connack) {}`
 
-Emitted on successful (re)connection (i.e. connack rc=0). 
-* `connack` received connack packet. When `clean` connection option is `false` and server has a previous session 
-for `clientId` connection option, then `connack.sessionPresent` flag is `true`. When that is the case, 
+Emitted on successful (re)connection (i.e. connack rc=0).
+* `connack` received connack packet. When `clean` connection option is `false` and server has a previous session
+for `clientId` connection option, then `connack.sessionPresent` flag is `true`. When that is the case,
 you may rely on stored session and prefer not to send subscribe commands for the client.
 
 #### Event `'reconnect'`
@@ -318,7 +321,7 @@ Subscribe to a topic or topics
 
 * `topic` is a `String` topic to subscribe to or an `Array` of
   topics to subscribe to. It can also be an object, it has as object
-  keys the topic name and as value the QoS, like `{'test1': 0, 'test2': 1}`. 
+  keys the topic name and as value the QoS, like `{'test1': 0, 'test2': 1}`.
   MQTT `topic` wildcard characters are supported (`+` - for single level and `#` - for multi level)
 * `options` is the options to subscribe with, including:
   * `qos` qos subscription level, default 0
@@ -475,6 +478,29 @@ you can then use mqtt.js in the browser with the same api than node's one.
 ```
 
 Your broker should accept websocket connection (see [MQTT over Websockets](https://github.com/mcollina/mosca/wiki/MQTT-over-Websockets) to setup [Mosca](http://mcollina.github.io/mosca/)).
+
+<a name="signedurls"></a>
+### Signed WebSocket Urls
+
+If you need to sign an url, for example for [AWS IoT](http://docs.aws.amazon.com/iot/latest/developerguide/protocols.html#mqtt-ws),
+then you can pass in a `transformWsUrl` function to the <a href="#connect"><code>mqtt.<b>connect()</b></code></a> options
+This is needed because signed urls have an expiry and eventually upon reconnects, a new signed url needs to be created:
+
+```js
+// This module doesn't actually exist, just an example
+var awsIotUrlSigner = require('awsIotUrlSigner')
+mqtt.connect('wss://a2ukbzaqo9vbpb.iot.ap-southeast-1.amazonaws.com/mqtt', {
+  transformWsUrl: function (url, options, client) {
+    // It's possible to inspect some state on options(pre parsed url components)
+    // and the client (reconnect state etc)
+    return awsIotUrlSigner(url)
+  }
+})
+
+// Now every time a new WebSocket connection is opened (hopefully not that
+// often) we get a freshly signed url
+
+```
 
 <a name="qos"></a>
 ## About QoS
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 [![DefinitelyTyped](http://img.shields.io/badge/Definitely-Typed-blue.svg)](https://github.com/borisyankov/DefinitelyTyped/tree/master/mqtt)
 
+[![codecov](https://codecov.io/gh/mqttjs/MQTT.js/branch/master/graph/badge.svg)](https://codecov.io/gh/mqttjs/MQTT.js)
+
 [![NPM](https://nodei.co/npm/mqtt.png)](https://nodei.co/npm/mqtt/)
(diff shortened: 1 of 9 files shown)


sublimator:
So I see there's a pre commit hook (specified in package.json, didn't know you could do that :) ) that runs the `test` script but not `browser-test`, which off course seems to be because zuul just hangs when the --local option is passed. It would be nice if that project had some kind of ability to `process.exit( passed ? 0 :  1)`.   

I guess at the least we could add `browser-build` to the pre-commit list of scripts?

lib/connect/ws.js
```diff
@@ -1,71 +1,78 @@
...
-  // for Web Workers! P.S: typeof(document) !== undefined may be becoming the faster one these days.
-  if (typeof (document) !== 'undefined') {
-    parsed = _URL.parse(document.URL)
-  } else {
+function buildBuilderBrowser (client, opts) {
+  // For Web Workers!
+  // P.S: typeof(document) === undefined may be becoming the faster one these days.
+  // TOOD: faster than what??
+  if (typeof (document) === 'undefined') {
+    // TODO: Why doesn't this check for `host` ? re: `Specify host manually.`
```
mcollina: why those todos?

cc @sandro-k

mcollina:
Good job on the testing side, thanks!!!!!!

sublimator:
Todos r questions to you guys and reminders to myself! I can remove if/b4 u
merge

sublimator:
Will add docs/strip todos after lunch

lib/connect/ws.js
sandro-k: I am not sure what the question is on line 
```
71 // TODO: Why doesn't this check for `host` ? re: `Specify host manually.`
```

sublimator:
re: TODO: Why doesn't this check for `host` ? re: `Specify host manually.`

look at the old code:
```js
function buildBuilderBrowser (mqttClient, opts) {
  var wsOpt = {
    protocol: 'mqtt'
  }
  var url
  var parsed

  // for Web Workers! P.S: typeof(document) !== undefined may be becoming the faster one these days.
  if (typeof (document) !== 'undefined') {
    parsed = _URL.parse(document.URL)
  } else {
    throw new Error('Could not determine host. Specify host manually.')
  }
```

The code doesn't even give the user a chance to specify a host manually, no?

sublimator:
I'm not quite sure about the implications of WebWorkers and browsers but it seems like the old error message indicated that you ought to be able to specify a host manually somehow? 

What should the correct behaviour (ou showing my Aussienessness lelz) be? Should it only throw an error if there's not enough information in the `opts` to piece together an url?

sublimator:
> I'm more keen on passing a function to connect that will return the URL (or call a callback with an URL).

My thoughts on that:

If you passed in solely a function, it would return a fully signed url, and then as the code stands, it would be parsed, in the connect function, which it must do, due to having to route to the corresponding stream builders for the appropriate protocol. 

Then later, inside the stream builder, you would need logic to check if there's an url builder function and either run that, or use the current logic to build the url from the `options`. For the first connection, this would mean running the url builder function twice, once for the connect protocols[url.parse(urlBuilder())] routing, and then again inside the stream builder. Or add some convolution to avoid this.

I guess you could specify the protocol manually in the connect options and bypass this need: `{protocol: 'wss', url: urlBuilder()}` ?

Anyway, I think this is an efficient/pragmatic and reasonably non-invasive way of solving the issue, as the transformUrl function gets passed the url/options/client and can use either the url or the already parsed components, in the options, and check any possibly relevant client state.

I'm a little fuzzy on how the urls are used by the other protocols, but it seems it's nothing other than a convenient way of expressing domain/ip/port etc, whereas with WebSockets, the url is actually transported over the wire. Correct me if I'm wrong there.

sandro-k:
You check works just fine..  Currently I am creating the `document` object before MQTT.js is running [See L25](https://github.com/mqttjs/mqtt-worker/blob/master/mqttWorker.js#L25). Ideally it should be possible to pass this option to MQTT.js.

mcollina:
> Anyway, I think this is an efficient/pragmatic and reasonably non-invasive way of solving the issue, as the transformUrl function gets passed the url/options/client and can use either the url or the already parsed components, in the options, and check any possibly relevant client state.

Agreed, let's try to get this in then.

Would you mind to adding the docs?

sublimator:
@sandro-k 

I'm still confused. Should it only parse the document.URL and throw an error if it cant if there's not enough information in the `opts` to piece together an url?

sublimator:
Btw, this comment here seems stale, https://github.com/mqttjs/mqtt-worker/blob/master/mqttWorker.js#L23-L25, the url is passed in as the first arg no?

sublimator:
```js
function buildBuilderBrowser (client, opts) {
  if (!opts.hostname) {
    opts.hostname = opts.host
  }

  if (!opts.hostname) {
    if (typeof (document) === 'undefined') {
      throw new Error('Could not determine host. Specify host manually.')
    }
    var parsed = urlModule.parse(document.URL)
    opts.hostname = parsed.hostname

    if (!opts.port) {
      opts.port = parsed.port
    }
  }
  return createWebSocket(client, opts)
}
```

This seems like the correct logic that would be inline with the Error string

sandro-k:
@sublimator Yes you are correct.. looks very clean.

sublimator:
@mcollina I added draft @ docs

@sandro-k  I removed the comment about webworkers, can you provide me with an update one to add?

README.md
```diff
@@ -212,6 +235,7 @@ the `connect` event. Typically a `net.Socket`.
     * `payload`: the message to publish
     * `qos`: the QoS
     * `retain`: the retain flag
+  * `transformUrl` : optional `(url, options, client) => url` function which can be used to implement signed urls which may expire, and cause troubles upon reconnect.
```
mcollina: this needs to be specified that is for websocket only.

README.md
```diff
@@ -104,6 +105,28 @@ run the example.
 
 to use MQTT.js in the browser see the [browserify](#browserify) section
 
+<a name="signedurls"></a>
+## Signed Urls
```
mcollina: can you move it after the `API`?

sublimator:
@mcollina 

I added codecov.io, it's really simple to set up, though of course, you'd have to do that yourself. You  can have a browse here: https://codecov.io/gh/sublimator/MQTT.js/src/d3a09f051f16459985d7c70fb64d497324395cd4/lib/connect/ws.js

lib/connect/ws.js
```diff
@@ -1,106 +1,89 @@
...
+}
 
+function buildBuilderBrowser (client, opts) {
   if (!opts.hostname) {
     opts.hostname = opts.host
   }
 
   if (!opts.hostname) {
+    if (typeof (document) === 'undefined') {
+      throw new Error('Could not determine host. Specify host manually.')
```
sandro-k: // Throwing an error in a Web Worker if no hostname is given, because we can not determann the hostname automatically. If connecting to localhost, please supply the hostname as an argument.

sublimator:
Heh, pretty slick, it's already got some info: https://codecov.io/gh/mqttjs/MQTT.js/src/262f872c997af410376660d238d8b92e00ce5e32/mqtt.js

But it's causing the builds to fail because there's no base commit to compare coverage to. I think you can set the tolerance, if you have repo rights and login via github. Probably easiest to just check the travis result, which seems to have passed and then once a build is triggered for master, the base should be set. I think ....

sublimator:
@sandro-k Why can't we just default to localhost in webworkers?

sublimator:
I'm completely unfamiliar with web workers

sandro-k:
No I would not do that.. the line ` var parsed = urlModule.parse(document.URL)` basically connects to the current hostname on the same port via Websockets. I don't even know a good usecase for hat behavior. On a Web Worker it would be even more confusing having the site at "https://footbar" and suddenly connecting Websockets to "http://localhost"...

sublimator:
Ok, makes sense :)

sublimator:
*@mcollina should I put the Signed WebSocket Urls section inside the
browser section?*

mcollina:
@sublimator yes, move the example at the top inside the browser section.

sublimator:
Updated and pushed. Sorry, fell asleep!

README.md
```diff
@@ -212,6 +214,9 @@ the `connect` event. Typically a `net.Socket`.
     * `payload`: the message to publish
     * `qos`: the QoS
     * `retain`: the retain flag
+  * `transformWsUrl` : optional `(url, options, client) => url` function
+        For wss/wss protocols only. Can be used to implement resigning expired
```
mcollina: this should read `ws/wss`

sublimator: oops

sublimator: pushed fix

PR #542, opened 2017-02-14
Trying to fix saucelabs integration

mcollina:
Following #539, the build is now broken. I've disabled the saucelabs runner in master to cut a release, but we shoud fix it.

@sublimator I've added you as a collaborator, push to this branch to run the hooks until it's green.

I'd like to have a "matrix" run, where we can run for each of the node versions, and just once for the browsers.

Ideally, we can mark that as "flaky", so that if it fails it's not a big issue.

If you need the credentials for saucelabs/ngrok, ping me via email and I'll send them to you.


sublimator:
Got it, yeah browser tests === goodness. Will take a look when I can :) 👍

sublimator:
Still haven't gotten around to this. We could add the browser-test to the prepublish script in the interim I guess for a quick tiny improvement. I'm not experienced in SauceLabs (yet, but want to be!)

YoDaMa:
closing because unrelevant.

PR #546, opened 2017-02-19
Tweak README

sublimator:
Just a little tweak of the badges, from when I was in a coffee shop the other day :) Looks a tad nicer IMHO


mcollina:
LGTM

Issue #536, opened 2017-02-12
Add TypeScript definitions for the api


sublimator:
Actually, there exists a `@types/mqtt` (which happens to be inconsistent with this latest version) but of course it would be ideal to have it in this repo (which is considered the best practice I believe).

mcollina:
Disclaimer: I am not a typescript user, and there is no one in the maintainers.
If you want to join and help, you are welcome!

In https://github.com/mqttjs/MQTT.js/pull/386 it was discussed this matter. I'm happy to embed the type definition here, if you want to help maintaining that.

cc @PekkaPLeppanen

sublimator:
Yeah, keen! we at @flomio are going full TypeScript.

sublimator:
k, once we got the other one sorted to your satisfaction, I'll get onto it :)

ogis-onishi:
I am using mqtt.js with TypeScript. I can contribute if you need my help.

sublimator:
Awesome :) There's a @types/mqtt package we can work with it's author with
:)

Just to find the time ...

PekkaPLeppanen:
I wrote the type definitions for the MQTT API v1. I was hoping that one day I'll write another definitions for v2 API also. Because a lot of time has passed and nothing happened from my part, I must say it would be great if someone else could do the new type definitions.
I am no longer using MQTT, because I changed to another company to a completely different type of project.

Anyway, to write the type definitions, you should take the API documentation and write the definitions from there. The missing parts can be found from the source. And make PR for the documentation, if there is something missing.

Also notice that the two versions of the type definitions can co-exist, if the API v1 is still maintained actively. If it's not maintained, one should rewrite the existing file.

sublimator:
I've been playing with async/await in TypeScript 2.1 and it's lovely. It would be nice to Promise-able the API while we're at it.

ogis-onishi:
And DefinitelyTypes repo must be updated as well.
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/mqtt

ogis-onishi:
I installed mqtt 2.5.0 by "npm install mqtt", but type definition files are not provided.

mcollina:
this should be done now in 2.5.1

yohei1126:
@sublimator 
I got the error "error TS2307: Cannot find module 'mqtt'." when compile.  How can I resolve the module?

sublimator:
[...]

PR #547, opened 2017-02-20
[WIP] Add TypeScript 2.1 typings

sublimator:
These headers are exported from a branch where I converted the lib/**/*.js to TypeScript. The tests all passed, when I modified them slightly to make sure they referenced package.json `main` pointing to the tsconfig.compilerOptions.outDir.

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -60,16 +60,47 @@ function parseBrokerUrl (brokerUrl) {
       // protocol
       // clientId
     {
-      // hostname: parsed.hostname,
       // NOTE: we use JUST the host, not parsed.host which includes the port
       host: parsed.hostname,
-      path: parsed.path || '/'
+      path: parsed.path || '/',
+      // TODO: remove
+      hostname: parsed.hostname
     })
 }
 
 function warnHostNameDeprecation () {
+  // The test uses hostname twice, once in `connect` when checking passed in
+  // opts and once in the acutall test to check the value is correct.
   console.warn('Use of mqtt.Client ' +
-    '`opts.hostname` is deprecated. Use `opts.host`')
+    '`opts.hostname` is deprecated. Use `opts.host`.\n' +
+    // Note we clone options in connect/ssl and delete 'hostname'
+    // so hostname isn't used.
+    patchHostName.usages +
+    ' usages detected.')
+}
+
+patchHostName.usages = 0
+function patchHostName (opts) {
+  if (!patchHostName.setHandler) {
+    process.once('exit', warnHostNameDeprecation)
+    patchHostName.setHandler = true
+  }
+
+  if (opts.hostname) {
+    var _hostname = opts.hostname
+    delete opts['hostname']
+    Object.defineProperty(opts, 'hostname', {
+      get: function () {
+        patchHostName.usages++
+        return _hostname
+      },
+      set: function (value) {
+        _hostname = value
+      },
+      configurable: true,
+      enumerable: true
+    })
+  }
 }
 
 /**
@@ -88,10 +119,9 @@ function connect (brokerUrl, opts) {
 
   if (opts.hostname) {
     opts.host = opts.hostname
-    delete opts['hostname']
+    patchHostName.usages++
+    // For testing
     warnHostNameDeprecation()
-    // So we can see it amongst jibberish
-    process.once('exit', warnHostNameDeprecation)
   }
 
   if (brokerUrl) {
@@ -101,6 +131,8 @@ function connect (brokerUrl, opts) {
       throw new Error('Missing protocol')
     }
   }
+
+  patchHostName(opts)
 
   // Someone out in the wild might be using {auth: 'user:pass'} out in the wild
   // Keep old behaviour? Tests actually pass without this.
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -21,17 +21,49 @@ protocols.wss = require('./ws')
  *
  * @param {Object} [opts] option object
  */
-function parseAuthOptions (opts) {
+function parseAuthOptions (auth, opts) {
   var matches
-  if (opts.auth) {
-    matches = opts.auth.match(/^(.+):(.+)$/)
+  if (auth) {
+    matches = auth.match(/^(.+):(.+)$/)
     if (matches) {
       opts.username = matches[1]
       opts.password = matches[2]
     } else {
-      opts.username = opts.auth
+      opts.username = auth
     }
   }
+}
+
+function parseBrokerUrl (brokerUrl) {
+  var opts = {}
+  var parsed = url.parse(brokerUrl, true)
+  if (parsed.port != null) {
+    opts.port = Number(parsed.port)
+  }
+  if (parsed.protocol) {
+    opts.protocol = parsed.protocol.replace(/:$/, '')
+  }
+
+  // Parse username/password
+  parseAuthOptions(parsed.auth, opts)
+
+  // Support clientId passed in the query string of the url
+  if (parsed.query && typeof parsed.query.clientId === 'string') {
+    opts.clientId = parsed.query.clientId
+  }
+
+  return xtend(
+    opts,
+      // username
+      // password
+      // port
+      // protocol
+      // clientId
+    {
+      hostname: parsed.hostname,
+      host: parsed.host,
+      path: parsed.path || '/'
+    })
 }
 
 /**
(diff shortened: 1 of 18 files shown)


sublimator:
re: #536 

I extended the client options from Url, because of the funny parsing done, but actually the `port` is more generally a `number` so need to fix that. 

Also TODO: remove the `_\w+` private functions

typings/lib/client.d.ts
```diff
@@ -0,0 +1,255 @@
...
+    stream: IStream;
+    streamBuilder: (MqttClient) => IStream;
+    options: ClientOptions;
+    queueQoSZero: boolean;
+    on(event: 'message', cb: OnMessageCallback): this;
+    on(event: 'packetsend' | 'packetreceive', cb: OnPacketCallback): this;
+    on(event: 'error', cb: OnErrorCallback): this;
+    on(event: string, cb: Function): any;
+    once(event: 'message', cb: OnMessageCallback): this;
+    once(event: 'packetsend' | 'packetreceive', cb: OnPacketCallback): this;
```
sublimator: image

image

typings/lib/client.d.ts
```diff
@@ -0,0 +1,255 @@
...
+     * @param {Function} cb - called when the client has been closed
+     *
+     * @api public
+     */
+    end(force?: boolean, cb?: boolean): this;
+    /**
+     * _reconnect - implement reconnection
+     * @api privateish
+     */
+    private _reconnect();
```
sublimator: need to kill these off

typings/lib/client-options.d.ts
```diff
@@ -0,0 +1,113 @@
+/// <reference types="node" />
+import { Url } from 'url';
+import Store from './store';
+import { QoS } from './types';
+/**
+ * MQTT CLIENT
+ */
+export interface ClientOptions extends SecureClientOptions, Url {
```
sublimator: Should just inline the `Url` options ... playing around

sublimator:
https://github.com/Microsoft/TypeScript/issues/1867#issuecomment-209018980

Hrmmm

sublimator:
Removing the private members was my initial impulse also, but one of the m$
devs made a point that it helps when extending the class to not
accidentally over-ride private members. The tsc compiler will balk on any
attempts to use the members, even on extended classes, where `protected`
must be used.

Having said that, you could argue it's less a maintenance burden to only
expose the public API.

Thoughts?

I'm easy either way really. Out at the moment so can't make any changes.

One other thing to consider is how much of the `url.parse` options do we
actually want in the ClientOptions. The 'slashes' etc ... it's kinda messy.

RangerMauve:
With MQTT in the wild, I haven't ever needed to extend it directly, favoring composition over inheritance when adding new capabilities. I would vote for not exposing the private variables for simplicity. (for autocomplete, for example)

sublimator:
So it seems like the consensus is to remove the private variables. There was some discussion about having an option to omit them in the automatically generated .d.ts files (which is what these are) but seems like MS dug their heels in on this one. I'll just remove them manually. 

@mcollina As far as tests go, we'd need to have a typescript compiler added to the dev-dependencies, which is not really a biggie. What sort of testing did you have in mind? Just simple tests (connect, subscribe, publish, receive etc) or more extensive ones to somehow notice when the declarations ones get out of sync?

mcollina:
At this point, just making sure if those are valid it's good.

mcollina:
I mean linting them.

sublimator:
Yeah, that seems like a good simple step 👍 
Ok, I will remove the private variables, install `tslint` and add it to the CI build :) 

Probably tomorrow

package.json
```diff
@@ -84,6 +86,9 @@
     "snazzy": "^6.0.0",
     "standard": "^8.6.0",
     "through2": "^2.0.3",
+    "tslint": "^4.5.1",
+    "tslint-config-standard": "^4.0.0",
+    "typescript": "^2.2.1",
```
sublimator: typescript is a peer dependency of tslint

types/lib/client-options.d.ts
```diff
@@ -0,0 +1,124 @@
+import { MqttClient } from './client'
+import { Store } from './store'
+import { QoS } from './types'
+
+export interface IClientOptions extends ISecureClientOptions {
+  // The options are extended by url.parse() fields in connect() function
```
sublimator: @mcollina 

What do you want exposed here? And once we've decided, should we only extend the IClientOptions with those, rather than 
just grabbing the kitchen sink?

```
> u.parse('mqtt://www.broker.com/mqtt')
Url {
  protocol: 'mqtt:',
  slashes: true,
  auth: null,
  host: 'www.broker.com',
  port: null,
  hostname: 'www.broker.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/mqtt',
  path: '/mqtt',
  href: 'mqtt://www.broker.com/mqtt' }
```

I'll switch to TypeScript branch and `find usages` and report back

types/lib/client.d.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+/**
+ * MqttClient constructor
+ *
+ * @param {Stream} stream - stream
+ * @param {Object} [options] - connection options
+ * (see Connection#connect)
+ */
+export declare class MqttClient extends events.EventEmitter {
+  // public connackTimer: NodeJS.Timer
```
sublimator: these weren't prefixed with a `_` so I marked them public (default in TypeScript if not specifide) (the first tslint configuration, before I used the `standard` variant wanted explicit so I just used the --fix options) 

In any case, I guess we don't really want them exposed?

types/lib/client-options.d.ts
sublimator: @mcollina I had a crack at cleaning up the url.parse/xtend section. Take a look

lib/connect/index.js
```diff
@@ -49,26 +81,16 @@ function connect (brokerUrl, opts) {
...
 
-  // merge in the auth options if supplied
-  parseAuthOptions(opts)
-
-  // support clientId passed in the query string of the url
-  if (opts.query && typeof opts.query.clientId === 'string') {
-    opts.clientId = opts.query.clientId
-  }
+  // Someone out in the wild might be using {auth: 'user:pass'} out in the wild
+  // Keep old behaviour? Tests actually pass without this.
```
sublimator: Personally I'd just remove this ...

lib/connect/index.js
```diff
@@ -49,26 +81,16 @@ function connect (brokerUrl, opts) {
...
   if (brokerUrl) {
-    var parsed = url.parse(brokerUrl, true)
-    if (parsed.port != null) {
-      parsed.port = Number(parsed.port)
-    }
-
-    opts = xtend(parsed, opts)
-
-    if (opts.protocol === null) {
+    // Options object always override brokerUrl specified options
```
sublimator: I did it this way due to a test:

```
    it('should throw an error when called with no protocol specified - with options', function () {
      (function () {
        mqtt.connect('tcp://foo.bar.com', { protocol: null })
      }).should.throw('Missing protocol')
    })
```

lib/connect/index.js
```diff
@@ -21,19 +21,51 @@ protocols.wss = require('./ws')
...
+
+  return xtend(
+    opts,
+      // username
+      // password
+      // port
+      // protocol
+      // clientId
+    {
+      hostname: parsed.hostname,
```
sublimator: It's interesting that the `servers` option uses {host, port} keys:

```
      opts.host = opts.servers[client._reconnectCount].host
      opts.port = opts.servers[client._reconnectCount].port
      opts.hostname = opts.host
```

Perhaps at some point it would be nice just to use one or the other

sublimator: I mean use only one of host or hostname consistently

sublimator: Hrmmm, ok, so I looked up the url.parse() function and

image

host is supposedly the hosname + ':' + port

sublimator:
@mcollina 

Ok, I removed private variables, and deprecated the `hostname` option in lieu of `host` which is used in the `servers: [{host: 'localhost', port: 1337},...]` option, for consistency. I also removed most of the jibberish url parse options (hash, query, etc), reducing it to just `port` (as a number) `host, path, protocol` and `clientId` and `username` / `password`.

sublimator:
The tests on Travis pass, but because of Js code changes related to hostname, the coverage has gone `down` by a fraction so it's marked as failed.

lib/connect/index.js
```diff
@@ -21,19 +21,57 @@ protocols.wss = require('./ws')
...
+
+  return xtend(
+    opts,
+      // username
+      // password
+      // port
+      // protocol
+      // clientId
+    {
+      // hostname: parsed.hostname,
```
sublimator: You could argue that to be consistent with deprecating, should leave the `hostname` until it's completely removed. Assuming you guys are down with that direction.

mcollina:
Can you please do not touch/remove/alter any existing functionality in this PR? I'm not sure what you want to achieve, and that is a semver-major change, while this is not.

I'll comment on the typescript questions later on.

sublimator:
My intention was to clean up the messy extending of the IClientOptions with
*ALL* the url.parse variables. It's quite glaring once you see it all
mapped out in TypeScript declarations, so I got a bit carried away making
alterations, while removing the url jibberish like `hash`/`query` etc. We
use hostname/host variously, while in server we use `host`.

I added a deprecation warning for `hostname`, for later removal but if
removing the rest warrants a sem-ver change then so be it. Is that an
issue? I can separate the changes into multiple pull requests if you'd like.

sublimator:
^ while in the *servers* config array we use `host`.

mcollina:
@sublimator are you ok in closing this? Or it's a mistake?

sublimator:
I will open another one later on :) Out at the moment

Issue #209, opened 2014-07-20
Message handlers can fail but puback already sent to server


jdiamond:
At qos 1+, shouldn't the message handler have a chance to succeed or fail before acknowledging the message was successfully received?

mcollina:
Released in 1.0.0. See https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L613-L623.

rossjourdain:
Hey @mcollina, I'd quite like to use this but I'm not sure how to prevent the puback from going out if I haven't managed to take care of the message yet.  Is that possible now?

mcollina:
@rossjourdain it has been possible for over a year. Maybe the docs are not great (please send PR :P), and definitely there is an issue with #425. Add an `handleMessage` function to the client, and call the callback when you are done. 

Any contribution/help into the project is welcomed!

rossjourdain:
I'm more that happy to contribute how ever I can once I get my head around this ;)
I can't seem to find any examples of the handleMessage process.

When I set the handleMessage function, it fires no trouble but the puback packet goes out regardless.  From my reading of the code, there's no way anything but that can happen.

I guess one way to do it is to make handleMessage take a success callback as well as a failure callback.  Something like this?

```
client.handleMessage = function(packet, success_cb, failure_cb) {
  console.log('handleMessage callback: ' + ' ' + packet.cmd);

  if(processTheMessage) {
    success_cb();
  } else {
    failure_cb():
  }
}
```

While the client.js side is something like this:

Modified version of:
https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L776-L800

```
switch (qos) {
    case 2:
      this.incomingStore.put(packet, function () {
        that._sendPacket({cmd: 'pubrec', messageId: mid}, done);
      });
      break;
    case 1:
      this.emit('message', topic, message, packet);
      // Send the puback on success only
      this.handleMessage(packet, function(){
        that._sendPacket({cmd: 'puback', messageId: mid}, done);
      }, done);
      break;
    case 0:
      this.emit('message', topic, message, packet);
      this.handleMessage(packet, done, done);
      break;
    default:
      // do nothing
      // log or throw an error about unknown qos
      break;
  }
```

Does that make any sense at all?

mcollina:
Just to clarify, I think that's not ok to have two functions.
I think that we **must** ACK a message, independently of the error. If the callback is called with an `'error'`, we should probably forward that `'error'` to the client itself.

[...]

PR #549, opened 2017-02-21
Fix for issue #209: Change PUBACK for QOS 1 to be on callback from handleMessage

willshulstad:
Fixed issue #209  where puback was sent before processing was complete for QOS 1.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -804,17 +804,25 @@ MqttClient.prototype._handlePublish = function (packet, done) {
       })
       break
     case 1:
-      // do not wait sending a puback
-      // no callback passed
-      this._sendPacket({
-        cmd: 'puback',
-        messageId: mid
-      })
-      /* falls through */
     case 0:
       // emit the message event for both qos 1 and 0
       this.emit('message', topic, message, packet)
-      this.handleMessage(packet, done)
+      this.handleMessage(packet, function (err) {
+        if (qos > 0) {
+          // for QOS 1, expect to send back the ack
+          if (!err) {
+            // if not error was returned, send the ack back
+            that._sendPacket({
+              cmd: 'puback',
+              messageId: mid
+            })
+          } else {
+            // error occured.  broker will resend message later
+            // implmentations that have handleMessage respond with an error should have logic to decide whether or not to disconnect/reconnect
+          }
+        }
+        done()
+      })
       break
     default:
       // do nothing


jdiamond:
Thanks for this!

You don't think it should automatically disconnect when the `handleMessage` callback receives an error?

willshulstad:
I don't think that the error itself should cause a disconnect internally in the client.  The only time that handleMessage will return an error is if the specific implementation has error checking logic in the handleMessage method.  If we leave the disconnect logic up to the specific implementation, then the implementor can decide if a disconnect is necessary or not.  

The use case we have is that we are using the client to watch for messages on the broker and storing them to our DB using rest services.  If those rest services are down, we don't want to ack the message because our client is very lightweight and does not persist anything to disk internally.  We rely on the broker to own the message until we have successfully sent the message on to our rest services.  

In the case that the rest services are temporarily down, we would not disconnect the client from the broker, we would just wait for the broker to resend the message and hope that the rest services are up by then.  If however, we had a different issue (internal exception, multiple sequential failures, etc.) then we would decide to disconnect from the broker.

mcollina:
@willshulstad I do not think this PR is correct. Specifically:

>  In the QoS 1 delivery protocol, the Receiver MUST respond with a PUBACK Packet containing the Packet Identifier from the incoming PUBLISH Packet, having accepted ownership of the Application Message (Section 4.3.2).

The important part for this PR is "accepted ownership", which I think it is about reception, **not** having successfully processed the message.

willshulstad:
@mcollina 
The problem with that interpretation is the client does not inherently handle persistence.  Since the application that is implementing handleMessage will be handling persistence, it should be allowed to decide on whether to accept ownership.  I don't think "accepting ownership" is simply about receiving, but about accepting responsibility for the message.

The way you describe "accepting ownership" actually sounds more like what is expected for QOS 0, rather than QOS 1:

> In the QoS 0 delivery protocol, the Receiver Accepts ownership of the message when it receives the PUBLISH packet.

As it stands today, if the application has any issue persisting the message, the ACK will already have been sent and the broker will have removed the message.  At that point, the message will be lost forever.  That makes this client unusable for any application that requires message delivery integrity.

RangerMauve:
@willshulstad AFAIK MQTT.js does persistence via [a store](https://github.com/mqttjs/MQTT.js#store). If that isn't working for your use case, you could implement a custom store that integrates with your application.

willshulstad:
@RangerMauve I may be reading the code wrong, but it does not look like the message store is used for incoming messages with QOS 1.  The store seems to be used for outgoing messages (to handle the PUBACK from the broker) and QOS 2 (to handle the PUBREL from the broker).

kerber:
I agree with Will that it would be great to be able to have the implementation control whether it takes responsibility for the received message. In the event of an error in my client, I'd rather trust a broker to keep the data safe than I would my own implementation of a data store. This would also mean that I would have to make sure my data store is performant and truly durable.

mcollina:
There is `incomingStore` for that https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L87.
Specifically, `incomingStore` is used for QoS 2 packets: https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L802-L804.

> I agree with Will that it would be great to be able to have the implementation control whether it takes responsibility for the received message. In the event of an error in my client, I'd rather trust a broker to keep the data safe than I would my own implementation of a data store. This would also mean that I would have to make sure my data store is performant and truly durable.

I am not 100% convinced about this. Let's make the case that the broker has a "botched" message in the queue to be sent, one that will crash the client on receipt. With the current approach, the client will crash and likely restart. However, with the approach you are suggesting, the client would continue to crash, because the message is never removed from the broker queue.

kerber:
Correct me if I'm wrong, but it doesn't appear that the `incomingStore` handles recovery from any sort of failure, rather it is just responsible for keeping track for the QOS2 publish exchange. It doesn't actually help ensure that QOS1 or QOS2 messages are persisted once the client takes ownership and responsibility of the message. Further, there's no actual mechanism in this library to protect the messages it owns in the event of a server or application failure.

In any sort of serious application that I can think of, it would be preferable to have the implementation handle malformed packets, rather than just dumping data. Data integrity, I would think, is higher priority than handling bad messages, as the former is explicitly covered in the spec and the latter is an implementation detail.

mcollina:
> Correct me if I'm wrong, but it doesn't appear that the incomingStore handles recovery from any sort of failure, rather it is just responsible for keeping track for the QOS2 publish exchange. It doesn't actually help ensure that QOS1 or QOS2 messages are persisted once the client takes ownership and responsibility of the message. Further, there's no actual mechanism in this library to protect the messages it owns in the event of a server or application failure.

Definitely. You are probably looking for a different type of library or even protocol, with a different design choices. MQTT is a protocol is build for device -> broker communication, not a server <-> server protocol.

> In any sort of serious application that I can think of, it would be preferable to have the implementation handle malformed packets, rather than just dumping data. Data integrity, I would think, is higher priority than handling bad messages, as the former is explicitly covered in the spec and the latter is an implementation detail.

In fact, it's quite the opposite. MQTT is a protocol to be deployed in the field, and a supreme imperative is to keep working, even if there is some form of "panic". A malformed message in the the queue to the client can make a client crash over and over again at each reconnect.
I have seen this in real-life several time, and a lost message is always better in IoT systems.

From the spec:

> The receiver is not required to complete delivery of the Application Message before sending the PUBACK. When its original sender receives the PUBACK packet, ownership of the Application Message is transferred to the receiver.

Our current failure model is "if it crashes it crashes", and every data that is "in flight" is lost.

jdiamond:
@mcollina: I think you see MQTT.js as the "receiver" in the spec. To me, MQTT.js is just a library and my applications are the receivers and, therefore, should have the ability to decide when they accept ownership of a message.

I agree with you that not sending a PUBACK violates the spec, but sending it without giving the application any say feels too presumptuous to me. Should it at least be able to end the connection instead of sending a PUBACK?

I think overriding `handleMessage` is an advanced use case that only experienced developers should be doing and we should document it with appropriate warnings that the connection should really be terminated if it can't call back without an error. But having it gives others the ability to experiment with use cases that might be different from the ones we have experience with.

mcollina:
Closing for no activity, feel free to reopen!

PR #551, opened 2017-02-28
Various tweaks:

sublimator:
## Various tweaks:

* Make options.port a number, after parsing the brokerUrl
* Update some package.json dependencies (bar `ws`, left at 1.x)
* Make travis `script` execute `npm run ci` to send codecov results
* Add an .editorconfig file
* Add nodejs 4.3.2 to the build matrix for AWS Lambda
* Add C++11 compatible compiler for extensions

Mostly fairly innocuous changes, except the first one, which might give some people a little grief, so needs some consideration. The majority/meat of the tests all pass however (*one* needed adjusting to test the `options.port` set by a parsed url against number rather than a string). The port string/number inconsistency is something I noticed when porting the lib/* to TypeScript a while back. 

I'd like to get that option consistently a number, then move ahead with #547 

I added 4.3.2 to the build matrix just as a heads up. If it gets in the way of where people wanna go (e.g. updating ws to 2.x) we can just disable it. In the mean time, having the CI notify of breakages would be handy for AWS Lambda users, which has an ancient version of node, stuck in the past. Tsk Tsk AWS 👎


sublimator:
Anyone?

mcollina:
I'm currently traveling, I will review this asap.

sublimator:
No rush :)

PR #556, opened 2017-03-04
Resubscribe even if disconnect is before suback

### LINKED ISSUE #555: resubscribe on reconnect fails when disconnect happens before suback
I can submit a fix for this, but I want to be sure it's a legitimate issue before creating a pr.

1. **client** issues a subscribe to `topic1`
2. **server** issues a suback for `topic1`
3. **client** gets suback for `topic1`
4. **client** issues a subscribe to `topic2`
5. **server/client** connection is severed
6. **client** reconnects

**result**: client resubscribes to `topic1` but not `topic2`
**expected**: client resubscribes to `topic1` and `topic2`

This is because the topic is only added to the subscribed topic list after the suback: https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L465-L467

gerad:
Change resubscribe logic to resubscribe to topics on reconnect even if
the disconnect happened prior to the suback response.

Fix #555

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -90,7 +90,7 @@ function MqttClient (streamBuilder, options) {
   this.queueQoSZero = this.options.queueQoSZero === undefined ? true : this.options.queueQoSZero
 
   // map of subscribed topics to support reconnection
-  this._subscribedTopics = {}
+  this._resubscribeTopics = {}
 
   // Ping timer, setup in _setupPingTimer
   this.pingTimer = null
@@ -190,9 +190,9 @@ function MqttClient (streamBuilder, options) {
   })
 
   // resubscribe
-  this.on('connect', function () {
-    if (this.options.clean && Object.keys(this._subscribedTopics).length > 0) {
-      this.subscribe(this._subscribedTopics)
+  this.on('reconnect', function () {
+    if (this.options.clean && Object.keys(this._resubscribeTopics).length > 0) {
+      this.subscribe(this._resubscribeTopics)
     }
   })
 
@@ -460,12 +460,13 @@ MqttClient.prototype.subscribe = function () {
     messageId: this._nextId()
   }
 
+  // subscriptions to resubscribe to in case of disconnect
+  subs.forEach(function (sub) {
+    that._resubscribeTopics[sub.topic] = sub.qos
+  })
+
   this.outgoing[packet.messageId] = function (err, packet) {
     if (!err) {
-      subs.forEach(function (sub) {
-        that._subscribedTopics[sub.topic] = sub.qos
-      })
-
       var granted = packet.granted
       for (var i = 0; i < granted.length; i += 1) {
         subs[i].qos = granted[i]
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -190,9 +190,9 @@ function MqttClient (streamBuilder, options) {
   })
 
   // resubscribe
-  this.on('connect', function () {
-    if (this.options.clean && Object.keys(this._subscribedTopics).length > 0) {
-      this.subscribe(this._subscribedTopics)
+  this.on('reconnect', function () {
```
gerad: otherwise, the `subscribe` packet could be sent twice on the initial connect: once in the initial subscribe (prior to `connack`) and then subsequently during the connect event (since it was added to the `_resubscribeTopics` at initial subscribe time.

test/client.js
```diff
@@ -389,5 +389,32 @@ describe('MqttClient', function () {
...
+          } else {
+            connectCount.should.equal(2)
+            subscribeCount.should.equal(2)
+            client.end(true, done)
+          }
+        })
+      })
+
+      client.subscribe('hello')
+    })
```
gerad: i tried really hard to get this into `abstract_client.js` but the `websocket_client.js` tests all failed for me locally… there may be a better way

mcollina: Can you move it there? They should pass on travis.
Why are they failing?

lib/client.js
mcollina: can you add a unit test that cover this case, so we check we don't get a duplicate subscription?

gerad: That case is covered in the test I added already: https://github.com/mqttjs/MQTT.js/pull/556/files#diff-7e3bd8e7059a21bfd255649cbfc84b33R411

Do you want it in a separate test as well?

test/client.js
gerad: Here's why they're failing locally.

image

gerad: Moved it to `abstract_client.js` in a7e9867. Let's see what Travis thinks.

gerad: Same error is happening on Travis. I guess that's a good thing?

gerad: I fixed the test state leakage issue in e9c78db. The fix is somewhat gross, but at least it's self contained.

test/abstract_client.js
```diff
@@ -1536,5 +1536,52 @@ module.exports = function (server, config) {
...
+          server.on('client', listener)
+        })
+      })
+
+      it('should resubscribe even if disconnect is before suback', function (done) {
+        var client = mqtt.connect(Object.assign({ reconnectPeriod: 100 }, config))
+        var subscribeCount = 0
+        var connectCount = 0
+
+        server.on('client', function (serverClient) {
```
mcollina: have you tried just using `once('client')` here?

gerad: Yeah. It doesn't work. 

First of all it has to be done twice since the client reconnects, and even if I code around that, the existing event listeners blow up when I sever the connection in the initial subscription. TBH, I'm a bit surprised the rest of the tests work as well as they do that way.

PR #562, opened 2017-03-10
feat!: improve `connect(brokerUrl)` parsing

sublimator:
This pull request attempts to tidy up the client options object, by pulling in only relevant information from `url.parse(brokerUrl)` which the `client.options` gets extended with by use of `xtend` in the `connect()` function.

There are a lot of properties here that aren't really relevant.

```
> u.parse('mqtt://www.broker.com')
Url {
  protocol: 'mqtt:',
  slashes: true,
  auth: null,
  host: 'www.broker.com',
  port: null,
  hostname: 'www.broker.com',
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: 'mqtt://www.broker.com' }
```

Now the options object will only get extended with these properties:

* username
* password
* port
* protocol
* path
* host
* hostname
* clientId

`host` is set to `hostname` as the url.parse(...).host property includes the port.

Near as I can tell, once the various stream builders were updated to use `host` the only place in the code that uses `hostname` is in the tls.js where tls.connect() seems to access `hostname` in precedence over `host`. 

Note that the `servers` option is an array of `{port, host}` objects.  

I think it makes sense to deprecate `hostname`, for consistency sake, though if removal of the other options warrants a semver `breaking` change bump then may as well just remove it now?

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -21,16 +21,85 @@ protocols.wss = require('./ws')
  *
  * @param {Object} [opts] option object
  */
-function parseAuthOptions (opts) {
+function parseAuthOptions (auth, opts) {
   var matches
-  if (opts.auth) {
-    matches = opts.auth.match(/^(.+):(.+)$/)
+  if (auth) {
+    matches = auth.match(/^(.+):(.+)$/)
     if (matches) {
       opts.username = matches[1]
       opts.password = matches[2]
     } else {
-      opts.username = opts.auth
+      opts.username = auth
     }
+  }
+}
+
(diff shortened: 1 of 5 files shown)


sublimator:
Totally willing to hear `if it aint broke, don't fix it` on this. It's not really a big deal.

mcollina:
@sublimator this is an absolutely valid contribution. We can get it merged when we decide to bump the major version number.

robertsLando:
@sublimator are you interested to continue with this? we are releasing a new major. Still beta for now.

If yes please also remember  to update the destination branch to `main` instead of `master` 🙏🏼

PR #563, opened 2017-03-10
feat: jittered fibonacci back off strategy

### LINKED ISSUE #561: Change reconnect logic to exponential backoff with random delay
If a client loses the connection, the reconnect starts immediately. If your broker died and you restart it, all clients try to connect at the same time and may cause problems with traffic, or authorization logic and maybe kill it again. So in my opinion the reconnect should be changed to an exponential backoff with jitter, explained in [this article](https://www.awsarchitectureblog.com/2015/03/backoff.html).

What do you think about this idea?

sclausen:
To prevent all clients try to reconnect at the same time over and over again, which may be harmful to the broker, here is the option to activate a back off strategy I called jittered-fibonacci. It's not that brutal than exponential back off and it includes a small, configurable, deviance so the clients do not try to connect all at the same time.

Fixes #561


lib/client.js
```diff
@@ -21,6 +21,10 @@ var defaultConnectOptions = {
   protocolId: 'MQTT',
   protocolVersion: 4,
   reconnectPeriod: 1000,
+  backOffStrategy: null,
```
mcollina: I think `backOffStrategy` should be a `function`, rather than a string, and it should be more modular.
I'm happy if we provide this back off strategy as a separate module as well.

sclausen: I don't know, how I could make the backoff strategy more modular. It relies on some member variables and I could'nt think of a generalization of clearInterval/clearTimeout along other algorithm specific variables :-/
Could you explain your thoughts on this a little?

mcollina: We might just support an object `BackOffStrategy`, with methods `next()`, `reset()`, `unref()`. I'm throwing in ideas, I'm not 100% sure that comparing strings is the best option.

robertsLando:

PR #564, opened 2017-03-10
Added note about async-mqtt

RangerMauve:
Added it after the example so that it wasn't too high up. Also a link to what `async-await` is in case people didn't know about it.


RangerMauve:
👌 Hopefully people find this useful. 😄

PR #565, opened 2017-03-13
Add TypeScript types

sublimator:


sublimator:
Yeah, they probably should be, but I can't be bothered at this point in time :)

sublimator:
We can close this and wait until I can be bothered ... (Mondayitis haha) ... or just push it as is now and move them later on I guess ... 

Hrmmm Lemme look tomorrow, got other stuff need to attend to

mcollina:
let's merge this, then we can do that in a followup!
There is a definitely typed link in the README. Maybe we should remove that!

mcollina:
Do we need to indicate in some other way this library has full TypeScript definitions?

sublimator:
Removed the badge! Feel free make other any changes you want and merge!

sublimator:
> Do we need to indicate in some other way this library has full TypeScript definitions?

I'm not sure! I'm kinda new to TypeScript myself. You'll note that I added a `types` entry to the package.json, which the TypeScript compiler is aware of, so it all should "just work". As far as badges etc for advertising goes ...

sublimator:
Here's an example of a repo with inline .d.ts files: https://github.com/aws/aws-sdk-js. It has no badge, but has a README section, which we could add at some point. 

Could slap one of the badges from here up top: https://ellerbrock.github.io/typescript-badges/

mcollina:
A section in the README would be awesome :)

PR #587, opened 2017-04-04
Add a callback for the end function

### LINKED ISSUE #571: TypeScript Client.end declaration wrong

yohei1126:
fixes #571 

I think that the callback for the end functions should be explicitly defined as the same as the other callbacks.


yohei1126:
@sublimator Would you check it?

sublimator:
@yohei1126 Thanks! Merged

PR #603, opened 2017-04-25
Add duplicate flag to publish options

FrankAdesys:
For a specific problem I require to mark a message as duplicate when publishing to a topic. With this change the flag can be added to the options when publishing messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 ![mqtt.js](https://raw.githubusercontent.com/mqttjs/MQTT.js/137ee0e3940c1f01049a30248c70f24dc6e6f829/MQTT.js.png)
 =======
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,6 +1,8 @@
 ![mqtt.js](https://raw.githubusercontent.com/mqttjs/MQTT.js/137ee0e3940c1f01049a30248c70f24dc6e6f829/MQTT.js.png)
 =======
 
+Allow setting "dup" flag when publishing a message
+
```
mcollina: this would need to be moved in the publish section of the README, just like all the other options.

FrankAdesys:
Hi, sorry for that (first time I contribute to a project this way), included a test, hope this is the right way to go.

PR #610, opened 2017-05-11
fix type of some fileds in ISecureClientOptions

### LINKED ISSUE #609: TypeScript: ISecureClientOptions is not compatible with Node.js API
As  mentioned in the README, in case mqtts (mqtt over tls) the options object is passed through to tls.connect(). But type of some fields on ISecureClientOptions, options for tls connection, are not compatible.

The current diffinition is as follows.

```
export interface ISecureClientOptions {
  /**
   * path to private key
   */
  key?: string
  /**
   * path to corresponding public cert
   */
  cert?: string
  ca?: string
  rejectUnauthorized?: boolean
}
```
Acording to [the Node.js API documentation](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options), 

* `key` should be `key: key?: string | string[] | Buffer | Buffer[] | Object[]`
* `cert` should be `cert?: string | string[] | Buffer | Buffer[]`
* `ca` should be `ca?: string | string[] | Buffer | Buffer[]`

yohei1126:
fix #609

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -87,14 +87,17 @@ export interface IClientOptions extends ISecureClientOptions {
 }
 export interface ISecureClientOptions {
   /**
-   * path to private key
+   * optional private keys in PEM format
    */
-  key?: string
+  key?: string | string[] | Buffer | Buffer[] | Object[]
   /**
-   * path to corresponding public cert
+   * optional cert chains in PEM format
    */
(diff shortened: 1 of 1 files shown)


mcollina:
@sublimator would you mind reviewing this?

sublimator:
Thanks @yohei1126! sorry so slow to merge, have been really busy!

PR #617, opened 2017-05-25
added additional checks for tls errors

### LINKED ISSUE #614: Unhandled error event on connection lost with TLS
Hey,

ascoltatori version: 4.1.0

Behavior:

1. Broker up and running
2. mqtt client connected
3. Broker goes down

```
events.js:163
      ^

```

eiabea:
may has some sideeffects, but fixes #614

### DIFF
--- a/lib/connect/tls.js
+++ b/lib/connect/tls.js
@@ -20,7 +20,8 @@ function buildBuilder (mqttClient, opts) {
 
   function handleTLSerrors (err) {
     // How can I get verify this error is a tls error?
-    if (opts.rejectUnauthorized) {
+    // Only throw error if it is not an connection refused error to reconnect
+    if (opts.rejectUnauthorized && err && err.code !== 'ECONNREFUSED') {
       mqttClient.emit('error', err)
     }
 


mcollina:
would you mind adding a unit test for this?

eiabea:
@mcollina sure, but this was just a "proof of concept"-PR, just to check if it is a good idea to check for "ECONNREFUSED" explicit

lib/connect/tls.js
```diff
@@ -20,7 +20,8 @@ function buildBuilder (mqttClient, opts) {
 
   function handleTLSerrors (err) {
     // How can I get verify this error is a tls error?
-    if (opts.rejectUnauthorized) {
+    // Only throw error if it is not an connection refused error to reconnect
+    if (opts.rejectUnauthorized && err && err.code !== 'ECONNREFUSED') {
```
mcollina: instead of checking if it is not an `'ECONNREFUSED'`, can we verify if it is a TLS error?

eiabea: it would be the cleaner approach i think, but i never managed to get one, so i wrote it this way.
i also tried to write a test for it, but failed due to the fact that i can not kill the mocked secure server during a test

mcollina: @eiabea you can create a test-specific server with https://github.com/mqttjs/MQTT.js/blob/master/test/client.js#L242.

mcollina:
Closing for no activity, feel free to reopen!

PR #621, opened 2017-05-26
Do not enqueue an unlimited amount of subscribes if disconnected

### LINKED ISSUE #618: auto reconnect bug
The auto reconnect functionality is effectively DoS'ing my mqtt broker.  The reconnect process is broken

Source File is Client.js
`MqttClient.prototype.subscribe` is called and at the end of this function, 

 `this._sendPacket(packet)` is called.

The first few line in sendPacket 

```
  if (!this.connected) {
    if (packet.qos > 0 || packet.cmd !== 'publish' || this.queueQoSZero) {
      this.queue.push({ packet: packet, cb: cb })
    } else if (cb) {
      cb(new Error('No connection to broker'))
    } return
  } 
```

This adds the topics to be resubscribed on connect to a queue.  This happens for every 1second that the client is disconnected.  If a client is disconnected for day that will be roughly 86400 topics in the queue.  When the client finally connects it subscribes to the same topics however many times it was added to the queue. 

```
// Send queued packets
  this.on('connect', function () {
    var queue = this.queue

    function deliver () {
      var entry = queue.shift()
      var packet = null

      if (!entry) {
        return
      }

      packet = entry.packet

      that._sendPacket(
        packet,
        function (err) {
          if (entry.cb) {
            entry.cb(err)
          }
          deliver()
        }
      )
    }

    deliver()
  })
```

I have clients that literally could disconnect for days or weeks at a time and this queue grows without limit.

mcollina:
Fixes #618.

@Shinjifei would you mind verifying that this solves your bug?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -189,11 +189,16 @@ function MqttClient (streamBuilder, options) {
     deliver()
   })
 
+  var firstConnection = true
   // resubscribe
-  this.on('reconnect', function () {
-    if (this.options.clean && Object.keys(this._resubscribeTopics).length > 0) {
+  this.on('connect', function () {
+    if (!firstConnection &&
+        this.options.clean &&
+        Object.keys(this._resubscribeTopics).length > 0) {
       this.subscribe(this._resubscribeTopics)
     }
+
+    firstConnection = false
   })
 
   // Clear ping timer
--- a/test/client.js
+++ b/test/client.js
@@ -326,6 +326,39 @@ describe('MqttClient', function () {
       })
     })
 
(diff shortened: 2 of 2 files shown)


Shinjifei:
Will try to test this weekend. Thanks.

Shinjifei:
Got to it quicker that I thought.  

Looks good, no duplicate subscriptions.  Speaking with my team we would love to have an option to turn of the resubscribe feature altogether.  I'll add a feature request.

Thanks for the quick fix.

mcollina:
@Shinjifei you can also fire a PR if you want.

PR #624, opened 2017-05-30
Use Store for qos>0 packets

### LINKED ISSUE #398: Store not processing messages between restarts
## Summary

Unsent messages are not processed from the store after restart.
## Steps to reproduce
1. Client loses connection
2. Client attempts to send message
3. Client process exits and restarts (e.g power loss)
4. Client reconnects
## Expected behaviour

Unsent messages are persisted in the store between restarts, and resent when the client restarts and reconnects to the broker.
## Actual behaviour

Unsent messages are not sent to the broker after restarting, despite using a store.
## Notes

Using QoS 1 and mqtt-level-store for persistence.

Possibly related to #343

### LINKED ISSUE #607: Messages published when disconnected are not added to the store
Messages posted while disconnected from the server are only added to the in-memory queue - not to the store. If the process is terminated before the MQTT server comes back online the messages are lost. This is a duplicate of #398.

To reproduce:
```javascript
"use strict";
var mqtt = require("mqtt");
var levelStore = require("mqtt-level-store");
var rl = require("readline").createInterface({ input: process.stdin, output: process.stdout });

var manager = levelStore("mqtt.db");
var client = mqtt.connect("mqtt://192.168.99.100", {
    incomingStore: manager.incoming,
    outgoingStore: manager.outgoing,
    clientId: "someclientid",
    clean: false
});

client.on("message", (topic, message) => {
    console.log("Received message: " + message);
});

client.on("connect", () => {
    console.log("Connected");
    client.subscribe("#");
});

client.publish("foo", "Hello, World!", { qos: 1 }, (err) => {
    if (err) {
        console.log("Publish failed", err);
    } else {
        console.log("Publish OK");
    }

    rl.question("Stop the MQTT server and press enter", () => {
        client.publish("foo", "Hello again!", { qos: 1 }, (err) => {
            if (err) {
                console.log("Publish failed", err);
            } else {
                console.log("Publish OK");
            }

            rl.close();
            client.end();
        });
    });
});
```

Output:
```
$ node app.js 
Connected
Received message: Hello, World!
Publish OK
Stop the MQTT server and press enter
^C
$ level mqtt.db/
level > get *
level > ^D
$ node app.js 
Connected
Received message: Hello, World!
Publish OK
Stop the MQTT server and press enter
```

The "Hello again" message is never received.

behrad:
`QOS=0` packets are buffered in-memory if `queueQoSZero` as before, however `QOS>0` packets are persisted in Store to survive client process crash.

Fixes #607 
Fixes #398

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -666,7 +666,7 @@ MqttClient.prototype._sendPacket = function (packet, cb) {
     if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
     } else if (packet.qos > 0) {
-      storePacket(this, packet, cb);
+      storePacket(this, packet, cb)
     } else if (cb) {
       cb(new Error('No connection to broker'))
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -41,6 +41,14 @@ function sendPacket (client, packet, cb) {
   }
 }
 
+function storePacket (client, packet, cb) {
+  client.outgoingStore.put(packet, function storedPacket (err) {
+    if (err) {
+      return cb && cb(err)
+    }
+  })
+}
+
 function storeAndSend (client, packet, cb) {
   client.outgoingStore.put(packet, function storedPacket (err) {
     if (err) {
@@ -655,8 +663,10 @@ MqttClient.prototype._cleanUp = function (forced, done) {
  */
 MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
-    if (packet.qos > 0 || packet.cmd !== 'publish' || this.queueQoSZero) {
+    if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
+    } else if (packet.qos > 0) {
+      storePacket(this, packet, cb);
     } else if (cb) {
       cb(new Error('No connection to broker'))
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -41,10 +41,6 @@ function sendPacket (client, packet, cb) {
   }
 }
 
-function storePacket (client, packet, cb) {
-  client.outgoingStore.put(packet, cb)
-}
-
 function storeAndSend (client, packet, cb) {
   client.outgoingStore.put(packet, function storedPacket (err) {
     if (err) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -42,11 +42,7 @@ function sendPacket (client, packet, cb) {
 }
 
 function storePacket (client, packet, cb) {
-  client.outgoingStore.put(packet, function storedPacket (err) {
-    if (err) {
-      return cb && cb(err)
-    }
-  })
+  client.outgoingStore.put(packet, cb)
 }
 
 function storeAndSend (client, packet, cb) {
@@ -666,7 +662,11 @@ MqttClient.prototype._sendPacket = function (packet, cb) {
     if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
     } else if (packet.qos > 0) {
-      storePacket(this, packet, cb)
+      storePacket(this, packet, function (err) {
+        if (err) {
+          return cb && cb(err)
+        }
+      })
     } else if (cb) {
       cb(new Error('No connection to broker'))
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -662,7 +662,7 @@ MqttClient.prototype._sendPacket = function (packet, cb) {
     if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
     } else if (packet.qos > 0) {
-      storePacket(this, packet, function (err) {
+      this.outgoingStore.put(packet, function (err) {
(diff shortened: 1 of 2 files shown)


mcollina:

behrad:
I don't know an easy test to simulate `mqtt.js` process crash, Have you got an idea @mcollina ?

mcollina:
what error was it crashing on?

lib/client.js
```diff
@@ -655,8 +663,10 @@ MqttClient.prototype._cleanUp = function (forced, done) {
  */
 MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
-    if (packet.qos > 0 || packet.cmd !== 'publish' || this.queueQoSZero) {
+    if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
+    } else if (packet.qos > 0) {
+      storePacket(this, packet, cb)
```
mcollina: this would store subscribes as well, do you think it will have any drawbacks?

behrad:
> what error was it crashing on?

Nothing. The crash is handmade to prove that mqtt.js will send stored qos>0 packets

lib/client.js
```diff
@@ -41,6 +41,14 @@ function sendPacket (client, packet, cb) {
   }
 }
 
+function storePacket (client, packet, cb) {
+  client.outgoingStore.put(packet, function storedPacket (err) {
+    if (err) {
+      return cb && cb(err)
+    }
+  })
```
mcollina: I think we should call `cb()` every time, not just in case of an error

mcollina: I think you can pass it directly to `put()`.

lib/client.js
behrad: why not to add a publish only gaurd?

lib/client.js
behrad: I think we shouldn't, I believe caller assumes `cb` as a (un)successful send operation, however storing a packet for later processing is an inflight state, caller should be suspended.
in a non-crash scenario,  `cb` will be called later when connection is ready and packet is sent. (if we'd called it before, it will be called twice!)

mcollina: we should always call the callback, if it's not there let's just noop it.

behrad: callback was not called when packet was queued, this is the same behaviour

mcollina:
I get what you are saying now. calling the function is wrong, but we are losing a reference to the callback by storing it. The queue is callback-aware, i.e we should call that callback when the subscribe is done, and by passing it to the store we loose a reference to it. I'm not sure what's the right fix atm :/.

RangerMauve:
Maybe have both in-memory (for callbacks) and in the store for persistence?

behrad:
> I get what you are saying now. calling the function is wrong, but we are losing a reference to the callback by storing it. The queue is callback-aware, i.e we should call that callback when the subscribe is done, and by passing it to the store we loose a reference to it. I'm not sure what's the right fix atm :/.

I checked this before proposing the PR, and hopefully callback is stored https://github.com/behrad/MQTT.js/blob/db3ce2f43b194d4cc662f3c64f959091ac37a632/lib/client.js#L386-L387 already :)

behrad:
> we should call that callback when the subscribe is done

I will guard only publishes

behrad:
> Maybe have both in-memory (for callbacks) and in the store for persistence?

I think, as a design change, mqttjs can drop in-memory queue, and store any inflight packet that should be queued.
let the `Store` to choose between memory/file system/...

RangerMauve:
> let the Store to choose between memory/file system/...

Sure, but you can't store callbacks in a file system. Callbakcs _have_ to be stored in memory

behrad:
> Callbakcs have to be stored in memory

they are already stored in `client#outgoing`

RangerMauve:
Disregard what I said, then. 😄

behrad:
I tested this with https://github.com/behrad/mqtt-nedb-store and my same issue was resolved.
One can also test again mqtt-level-store  and confirm this.

mcollina:
> > Callbakcs have to be stored in memory

> they are already stored in client#outgoing

Not true. That works only for QoS > 1.
However, we can limit the queue to publishes with QoS 0.

@RangerMauve may I get a LGTM?

@behrad would you like to join this org?

lib/client.js
```diff
@@ -655,8 +659,14 @@ MqttClient.prototype._cleanUp = function (forced, done) {
...
   if (!this.connected) {
-    if (packet.qos > 0 || packet.cmd !== 'publish' || this.queueQoSZero) {
+    if ((packet.qos === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
       this.queue.push({ packet: packet, cb: cb })
+    } else if (packet.qos > 0) {
+      storePacket(this, packet, function (err) {
+        if (err) {
+          return cb && cb(err)
+        }
+      })
```
mcollina: you can just avoid the `storePacket` function, and use `client.outgoingStore.put` here.

behrad:
> Not true. That works only for QoS > 1.
> However, we can limit the queue to publishes with QoS 0.

I also ment QoS > 1 publishes, this PR already limits queue to publishes with QoS 0. Can somebody also verify this?



> @behrad would you like to join this org?

My honor 👍

lib/client.js
behrad: 👍

behrad:
We can also add https://github.com/behrad/mqtt-nedb-store to the README if you like @mcollina

PR #626, opened 2017-05-31
TypeScript: QoS should be mandatory

### LINKED ISSUE #615: IClientSubscribeOptions's qos is defined as optional but required by code
https://github.com/mqttjs/MQTT.js/blob/master/types/lib/client-options.d.ts#L117
defines:
```
export interface IClientSubscribeOptions {
  /**
   * the QoS
   */
  qos?: QoS
}
```
but according to the js code, `qos` is required, otherwise, the client will not function properly.

yohei1126:
fixes #615

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -101,7 +101,7 @@ export interface IClientPublishOptions {
   /**
    * the QoS
    */
-  qos: QoS;
+  qos: QoS
   /**
    * the retain flag
    */
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -101,7 +101,7 @@ export interface IClientPublishOptions {
   /**
    * the QoS
    */
-  qos?: QoS
+  qos: QoS;
   /**
    * the retain flag
    */
@@ -115,5 +115,5 @@ export interface IClientSubscribeOptions {
   /**
    * the QoS
    */
-  qos?: QoS
+  qos: QoS
 }
--- a/types/lib/client.d.ts
+++ b/types/lib/client.d.ts
@@ -99,10 +99,8 @@ export declare class MqttClient extends events.EventEmitter {
    *     client.publish('topic', 'message', {qos: 1, retain: true})
    * @example client.publish('topic', 'message', console.log)
    */
-  public publish (topic: string, message: string | Buffer,
-                 opts: IClientPublishOptions, callback?: PacketCallback): this
-  public publish (topic: string, message: string | Buffer,
-                 callback?: PacketCallback): this
+  public publish (topic: string, message: string | Buffer, opts?: IClientPublishOptions, callback?: PacketCallback): this
+  public publish (topic: string, message: string | Buffer, callback?: PacketCallback): this
 
   /**
    * subscribe - subscribe to <topic>
@@ -120,13 +118,8 @@ export declare class MqttClient extends events.EventEmitter {
    * @example client.subscribe({'topic': 0, 'topic2': 1}, console.log)
    * @example client.subscribe('topic', console.log)
    */
-  public subscribe (topic:
-                     string
-                     | string[], opts: IClientSubscribeOptions, callback?: ClientSubscribeCallback): this
-  public subscribe (topic:
-                     string
-                     | string[]
-                     | ISubscriptionMap, callback?: ClientSubscribeCallback): this
+  public subscribe (topic: string | string[], opts?: IClientSubscribeOptions, callback?: ClientSubscribeCallback): this
+  public subscribe (topic: string | string[] | ISubscriptionMap, callback?: ClientSubscribeCallback): this
 
   /**
    * unsubscribe - unsubscribe from topic(s)
--- a/types/lib/client.d.ts
+++ b/types/lib/client.d.ts
@@ -99,8 +99,10 @@ export declare class MqttClient extends events.EventEmitter {
    *     client.publish('topic', 'message', {qos: 1, retain: true})
    * @example client.publish('topic', 'message', console.log)
    */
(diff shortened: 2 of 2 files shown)


yohei1126:
Please check @sublimator @mcollina

yohei1126:
@sublimator Could you check it?

yohei1126:
@sublimator any updates?

mcollina:
@yohei1126 would you like to help maintaining the typescript definitions of this library?
I'll add you to the org.

sublimator:
Sorry guys. I had a quick look the other day, but I wasn't sure. Been so busy :/

I asked this question: `Would you call it like publish('some/topic', null, function() {}) ?`

Let me try and answer that myself now.

sublimator:
Eh, I'm not even sure what I was asking myself before now haha :)

sublimator:
What cases are the changes to the new publish/subscribe overload
definitions intended to handle?

yohei1126:
@sublimator I can help it if you want.

yohei1126:
@sublimator see this issue. https://github.com/mqttjs/MQTT.js/issues/615

In original JS code, option object itself for publish/subscribe function is optional. In the case, qos is 0. But qos is not optional. If you specify option object for publish/subscribe function, you must specify qos.

This is not compatible to the TypeScript definition.

sublimator:
yes, would be great have someone else help out with TS maintenance.

Every time I remember to look at this, I'm usually wiped out and in email
reading mode, with little time or energy left

This seems fine and the qos change clearly a fix, however I'm questioning
the need for the publish/subscribe changes?

Can you state which cases it allows that it didn't before?

mcollina:
I think this might be more a bug in the implementation rather than a bug in the definition. For all intent and purposes `qos: 0` should be the default in any case. Can you send a PR if this is not the case?

yohei1126:
@sublimator 

As I mentioned, I made the `opts`, the option for publish / subscribe optional. This is the same implementation as the JS code. If you skip `opts`, QoS 0 will be set.
Before the changes, `opts` was mandatory. This is deifferent interface from the JS code.

yohei1126:
@mcollina 

I thinks it should be OK.

* subscribe
  * QoS 0 (and the other default option) will be set if the `opts` is skipped. https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L361
  * Callbask is not set for other than QoS 1/2. https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L387
  
* publish
  * QoS 0 (and the other default option) will be set if the `opts` is skipped. https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L360
  * Callbask is not set for other than QoS 1/2.

sublimator:
They (publish/subscribe) each had two overloads. One with 3 args, the other
with 2 (or 4/3 for publish) I thought any sensible case was already
covered. If you omitted opts, then you'd only have two args, and the 2 arg
override would be used, no?

Lets look at subscribe:

Subscribe(topic, opts) ( uses 3 arg overload)
Subscribe(topic, opts, callback) ( uses 3 arg overload)
subscribe(topic, callback) ( 2 arg overload)
Subscribe(topic) (2 arg overload)

The existing definition handles those cases fine. What other case is this
change intended to handle?

Is this something to do with strict null checks?

@mcollina

Are you saying passing an empty opts object should imply qos=0. And we
should fix the js rather than religiously follow the quirks in TS?

yohei1126:
@sublimator 

OK, I understand your point. the changes on publish/subscribe function is not necessary.

mcollina:
> Are you saying passing an empty opts object should imply qos=0. And we
should fix the js rather than religiously follow the quirks in TS?

Definitely, if that is not the case yet.

sublimator:
Hrmmm, looking at the JS code, it looks like we need to update the way the defaults are set, when an object is passed and missing some args. 

We already have a dependency on `xtend` (https://www.npmjs.com/package/xtend) so we could probably use that.

ogis-onishi:
The original issue #615 is "qos on the options are required". Can I fix this issue first then fix the following as an another issue?

```
Are you saying passing an empty opts object should imply qos=0. And we
should fix the js rather than religiously follow the quirks in TS?
Definitely, if that is not the case yet.
```

sublimator:
I think @mcollina was saying that qos should NOT be required? That there
should be a qos=0 default. I don't quite understand, sorry.

ogis-onishi:
OK. So both of `opts` object and qos are optional. Then default values are set if they are skipped. Right?

sublimator:
I believe that's what @mcollina wants

yohei1126:
I have thought about this a couple of minutes. I still feel an empty object as a parameter is weird as TypeScript interface.

```
client.subscribe('topic'); // this is OK.
client.subscribe('topic', {}); // this is weird for TypeScript
client.subscribe('topic', { qos: 0 }); // this is OK.
```

So, as TypeScript interface, option object can be optional but QoS in the option object must be specified. This is the same as my original suggestion.

But JS libary cannnot refuse empty object using type system. So, the default value must be set if an empty object is specified.

```
client.subscribe('topic', {}); // this cannot be refused in JS
```

mcollina:
> But JS libary cannnot refuse empty object using type system. So, the default value must be set if an empty object is specified.

Yes!

yohei1126:
The original issue #615 is about TypeScript interface. @sublimator could you check if this PR is OK?

I also added the issue on JS library itself here #638. I will fix it later.

yohei1126:
@sublimator Could you check it?

yohei1126:
@mcollina it seems @sublimator is super busy now..

sublimator:
Ugh, sorry, yes. Let me look in a few hours!

sublimator:
@yohei1126 

Many apologies for the long delay!

yohei1126:
thank you!
2017年7月12日(水) 13:09 sublimator <notifications@github.com>:

PR #635, opened 2017-06-11
Added checks for double subscription issue

GavinDmello:
This handles the double subscription issue on server restarts. This usually happens when the client has a subscribe call inside the connect event, ex the CLI <a>https://github.com/mqttjs/MQTT.js/blob/master/bin/sub.js#L90-L102</a> 
I'm not sure if this is the best way to handle this and open to a better way!

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -443,6 +443,7 @@ MqttClient.prototype.subscribe = function () {
   if (!opts) {
     opts = { qos: 0 }
   }
+
   if (Array.isArray(obj)) {
     obj.forEach(function (topic) {
       if (that._resubscribeTopics[topic] < opts.qos ||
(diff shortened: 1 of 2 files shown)


mcollina:
I think this is easier to be fixed in the CLI, subscribing on connect is not necessary anymore, as it will automatically reconnect anyway.

GavinDmello:
I was thinking of the same but the issue reoccurs for normal client scripts too. So everyone using subscribe command on connect will have to ensure it happens exactly once.

mcollina:
Ok! Could you add some unit tests for this?

GavinDmello:
Sure

lib/client.js
```diff
@@ -467,6 +480,11 @@ MqttClient.prototype.subscribe = function () {
     messageId: this._nextId()
   }
 
+  if (!subs.length) {
+    callback(null, [])
```
GavinDmello: @mcollina Will this break the API? I return an empty array if all subscriptions are duplicates

mcollina: I think this is ok, but add a unit test that covers this case.

mcollina:
Thanks!

PR #639, opened 2017-07-05
the default value must be set for an empty options parameter

### LINKED ISSUE #638: The default value must be set if an empty object is passed to subscribe/publish functions
https://github.com/mqttjs/MQTT.js/pull/626#issuecomment-308463305

JS libary cannnot refuse empty object as function parameter. So, the default value must be set if an empty object is specified.

* The default options for [publish](https://github.com/mqttjs/MQTT.js#publish) are  `{qos: 0, retain: false, dup: false}`
* The default options for [subscribe](https://github.com/mqttjs/MQTT.js#subscribe)) is `{qos: 0}`

yohei1126:
fixes #638

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -359,7 +359,7 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
   }
 
   // default opts
-  const defaultOpts = {qos: 0, retain: false, dup: false}
+  var defaultOpts = {qos: 0, retain: false, dup: false}
   opts = extend(defaultOpts, opts)
 
   if (this._checkDisconnecting(callback)) {
@@ -440,7 +440,7 @@ MqttClient.prototype.subscribe = function () {
     return this
   }
 
-  const defaultOpts = { qos: 0 }
+  var defaultOpts = { qos: 0 }
   opts = extend(defaultOpts, opts)
 
   if (Array.isArray(obj)) {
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -440,9 +440,8 @@ MqttClient.prototype.subscribe = function () {
     return this
   }
 
-  if (!opts) {
-    opts = { qos: 0 }
-  }
+  const defaultOpts = { qos: 0 }
```
mcollina: this should be `var`.

yohei1126: Thanks!

yohei1126:
@mcollina Could you check why the builds on Node 4.x were failed? The tests did not start.
https://travis-ci.org/mqttjs/MQTT.js/builds/250386104

yohei1126:
Oops. Is extend not available on Node 4.0?

```
Error: Cannot find module 'extend' from '/home/travis/build/mqttjs/MQTT.js/lib'
```

mcollina:
You should use the xtend module, not xtend, that's what we use through the normal dependencies. extend is likely in one of the tree, but we do not depend it directly and so it fails on an old version of npm.

yohei1126:
@mcollina Thanks. Please check it again.

mcollina:
Perfect!

PR #653, opened 2017-07-26
add support for weapp

taoqf:
For now, if using with webpack, need wait for https://github.com/jryans/timers-browserify/pull/24 and https://github.com/maxogden/websocket-stream/pull/129 merged.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,6 +17,5 @@
   "mocha": true,
   "indent": 2,
   "latedef": true,
-  "immed": true,
   "shadow": false
 }
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -11,6 +11,9 @@ if (process.title !== 'browser') {
   protocols.ssl = require('./tls')
   protocols.tls = require('./tls')
   protocols.mqtts = require('./tls')
+} else {
+  protocols.wx = require('./wx')
+  protocols.wxs = require('./wx')
 }
 
 protocols.ws = require('./ws')
@@ -72,7 +75,7 @@ function connect (brokerUrl, opts) {
 
   if (opts.cert && opts.key) {
     if (opts.protocol) {
-      if (['mqtts', 'wss'].indexOf(opts.protocol) === -1) {
+      if (['mqtts', 'wss', 'wxs'].indexOf(opts.protocol) === -1) {
         /*
          * jshint and eslint
          * complains that break from default cannot be reached after throw
@@ -89,6 +92,9 @@ function connect (brokerUrl, opts) {
           case 'ws':
             opts.protocol = 'wss'
             break
+          case 'wx':
+            opts.protocol = 'wxs'
+            break
           default:
             throw new Error('Unknown protocol for secure connection: "' + opts.protocol + '"!')
             break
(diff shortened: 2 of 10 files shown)


mcollina:
I'm a bit lost in what is weapp, wx and such.

taoqf:
https://mp.weixin.qq.com/cgi-bin/wx
I have added something to the readme,but tests could not be added because it's only could run in container.

taoqf:
@mcollina What can I do with the failled checks?

taoqf:
@mcollina Weixin(wx) is an application runs on android, iPhone, you may think it as [cordova](http://cordova.apache.org/), It allow h5 application runs in it, and it allow user use WebSocket using it's [api](https://mp.weixin.qq.com/debug/wxadoc/dev/api/network-socket.html).

taoqf:
Then could I add some unit test with failing judge? I think there are so many chinese programmer would need this.

.jshintrc
```diff
@@ -17,6 +17,5 @@
   "mocha": true,
   "indent": 2,
   "latedef": true,
-  "immed": true,
```
mcollina: why this?

taoqf: it was defined multi times. see line 4

mcollina:
Unit tests will be good?

taoqf:
I don't know how to add unit test yet, but I would try my best.

taoqf:
I have done some test, not unit test, it works good.

taoqf:
@mcollina so need I add unit test for this pr?

mcollina:
Yes.

taoqf:
Hi, @mcollina , I have add some test into `/test/wx.js` and passed `npm run weapp-test`.
is this the right way to add unit test?

taoqf:
Whatever I try, still could pass `codecov` check. any help?

mcollina:
I'll merge this. However, can you please restore `lib/connect/index.js` to what is there on master? It's all reported as changed in this PR.

lib/connect/index.js
```diff
@@ -1,146 +1,153 @@
...
+if (process.title !== 'browser') {
+  protocols.mqtt = require('./tcp')
+  protocols.tcp = require('./tcp')
+  protocols.ssl = require('./tls')
+  protocols.tls = require('./tls')
+  protocols.mqtts = require('./tls')
+}
+
+protocols.wx = require('./wx')
+protocols.wxs = require('./wx')
```
mcollina: we should not be adding `wx` if it's not a browser. the main `wx` object is not there.

taoqf:
@mcollina , Thanks, I move the code into `else` block, and I skiped the unit tests I added because it could not be running on node.js.

mcollina:
You can remove the skipped tests.

What have you changed to the lib/connect/index.js file? GH reports it was all changed, but it does not seems so. Can you please restore it to what was before and apply only your modifications?

taoqf:
I tried not skip the tests and it will cause a failing test after that test.
```bash
121 passing (16s)
  1 failing

  1) mqtt #connect should return an MqttClient when connect is called with wx:/ url:
     AssertionError: expected Function { name: '' } to throw exception
      at Assertion.fail (node_modules/should/cjs/should.js:258:17)
      at Assertion.value (node_modules/should/cjs/should.js:335:19)
      at Context.<anonymous> (test/mqtt.js:116:22)
```
I don't really know what was wrong with github, I had not changed the whole file `connect/index.js`. 
I will try to create a new pr later and see if gh reports the same.

vikingsailor:
@taoqf 在微信小程序里怎么使用mqtt连接AZURE?我已经在nodejs中联通了iot云，但是想在小程序中看到各设备的消息？如何实现？

taoqf:
Could you please describe azure?

PR #660, opened 2017-08-03
add support for weapp

taoqf:
https://github.com/mqttjs/MQTT.js/pull/653#issuecomment-319734340
I open this pr because of the gt reports.
@mcollina , stragely, I forked again and did the same change, I got failed message when I run `npm run weapp-test` first time. and I run npm test twice, it worked with no exception I got yesterday.
anyway, I hope this pr will be merged soon. thank you for your patience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ coverage
 test/typescript/.idea/*
 test/typescript/*.js
 test/typescript/*.map
+package-lock.json
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,6 +17,5 @@
   "mocha": true,
   "indent": 2,
   "latedef": true,
-  "immed": true,
   "shadow": false
 }
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ in JavaScript for node.js and the browser.
 * [Command Line Tools](#cli)
 * [API](#api)
 * [Browser](#browser)
+* [Weapp](#weapp)
 * [About QoS](#qos)
 * [TypeScript](#typescript)
 * [Contributing](#contributing)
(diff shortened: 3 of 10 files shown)


mcollina:
This is ok now!

.gitignore
```diff
@@ -14,3 +14,4 @@ coverage
 test/typescript/.idea/*
 test/typescript/*.js
 test/typescript/*.map
+package-lock.json
```
imZack: Why `package-lock.json` is in `.gitignore`?

> This file is intended to be committed into source repositories, and serves various purposes:
> 
> Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> 
> Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> 
> To facilitate greater visibility of tree changes through readable source control diffs.
> 
> And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

ref: https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md

mcollina: Because it's still a very experimmental technology and there have been a lot of bugs. Plus, it does not make the life of a maintainer easier. So, no, we should not use them.

imZack: Got it, I didn't notice about it too much before. Maybe we could also add a `.npmrc` file with `package-lock = false`

just like what eslint did https://github.com/eslint/eslint/issues/8742

mcollina: @imZack definitely. Would you like to send a PR? I've it disabled globally, so it's not a big deal for me, but I can see why it's needed for others.

boneskull:
generally I'd be in agreement, but I think we should reconsider this.

the [last build of master](https://travis-ci.org/mqttjs/MQTT.js/builds/274514804) passed, yet between then and now--without touching the code at all--something has broken to cause tests to fail.  

one failure I found when running `npm test` in my working copy prompted PR #689.  this PR also exhibits the *second* test failure, which I'm still trying to sort out.

the failing tests are adding significant friction to contributors, because we can't move forward until they're *all* fixed.  `package-lock` *could* mitigate this, but it *also* could result in false positives if the test failures are due to production deps and not just dev deps.

maybe someone has written an `npm bisect`-type tool--I'd love to be able to go back in time and recreate the `node_modules` structure of the last passing build!

mcollina:
@boneskull do you think we should remove support for weapp?

boneskull:
@mcollina sorry, I should have been more clear, I'm writing in reference to #662.

mcollina:
@boneskull you would like to revert package-lock? IMHO if it start to work as expected, I have no problem with it. I had more `rm -rf node_modules` in the last few months than in the rest of the years I have been doing node.

PR #677, opened 2017-09-05
Fixes #675 packet queuing only if "clean session" is false

### LINKED ISSUE #675: Too many duplicate `subscribe` requests after reconnect
Hi~

Sometimes, I have too many duplicate `subscribe` requests after reconnect.

I think it is because of the following queuing.
https://github.com/mqttjs/MQTT.js/blob/b1463afe6dcbe833005c8d333be4f0ad79338a45/lib/client.js#L713

Why not just return an error?
Or what about checking `clean session`?

leecoder:
in #675, there are too many replay packets.
I think the client will replay after reconnect when `clean session` is true.
So, replay packets could be queued only if `clean session` is false.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -709,7 +709,8 @@ MqttClient.prototype._cleanUp = function (forced, done) {
  */
 MqttClient.prototype._sendPacket = function (packet, cb) {
   if (!this.connected) {
-    if (((packet.qos || 0) === 0 && this.queueQoSZero) || packet.cmd !== 'publish') {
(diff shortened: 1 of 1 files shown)


mcollina:
why are you publishing when you are offline, if you do not want them to be sent?

This library aims for simplicity of usage, and it supports several patterns that are common in the Node.js world.

If the problem are too many subscribes, maybe there is a compltely different change to be made.

leecoder:
ok, I've changed to #679. WDYT?

PR #678, opened 2017-09-05
Enable handleMessage backpressure support for QoS 2 flows

BrandonSmith:


BrandonSmith:
@mcollina I didn't see a precedent for `handleMessage` tests, so I took a stab at it. My logic in the test is to publish from the server 10 messages and keep an incrementing counter of both the `message` event and `handleMessage` then compare the counters for equality at just before the `handleMessage` callback is called. The theory being that the `message` event should not be emitted until the callback is called. Please let me know if there should be deeper logic. Tests appear to pass for all QoS levels.

mcollina:
Thanks!!

PR #679, opened 2017-09-06
Fixes #675 eliminating duplicate subscribe

### LINKED ISSUE #675: Too many duplicate `subscribe` requests after reconnect

leecoder:
in #675, even if client would subscribe duplicated topics, it can be ignored.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -91,6 +91,9 @@ function MqttClient (streamBuilder, options) {
   // Should QoS zero messages be queued when the connection is broken?
   this.queueQoSZero = this.options.queueQoSZero === undefined ? true : this.options.queueQoSZero
 
+  // map of subscribed topics to dup check
+  this._subscribedTopics = {}
+
   // map of subscribed topics to support reconnection
   this._resubscribeTopics = {}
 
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -400,6 +403,15 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
   return this
 }
 
+MqttClient.prototype._checkSubscribedTopic = function (topic) {
+  if (!~~this._subscribedTopics[topic]) {
+    this._subscribedTopics[topic] = true
```
mcollina: this should be QoS aware, multiple subscribes with different QoS should be treated separately.

test/abstract_client.js
```diff
@@ -1838,6 +1838,43 @@ module.exports = function (server, config) {
...
+            // subscribes have taken place, then cleanup and exit
+            if (connectCount >= 2) {
+              subscribeCount.should.equal(2)
+              client.end(true, done)
+            }
+          })
+        })
+
+        // client subscribe init
+        client.subscribe('hello')
```
mcollina: we need to test that the callback is called anyway.

mcollina:
Is there any update on this one? Thanks!

psi-4ward:
+1

mcollina:
@psi-4ward would you like to send a fresh PR that includes the comments in my review? Thanks.

mcollina:
I’m closing this PR as it’s very clear it’s inactive.

Issue #682, opened 2017-09-13
ReschedulePings can cause a dropped connection to never be realized


duncanfinney:
Hi Guys,

I was noticing that when `reschedulePings` is set to `true` the sending of the ping packet and also the _check_ for a timed out connection is being rescheduled. 

I am pretty sure that only the _sending_ of the next ping should be rescheduled when a control packet goes out. There still needs to be a check each keepalive period to see if you've received a ping or not.

Example to reproduce:
```javascript
const mqtt = require('mqtt');
const client = mqtt.connect({ 
   keepalive: 10
   reschedulePings: true
    ... 
});

client.on('connect', () => {
    console.log('pull the WAN from your router right now, wait 30 seconds, plug it back in');
    setInterval(() => client.publish('foo', 'bar'), 1000);
});
```

Your client will never realize that they're disconnected.

I think that #630 may be fixed by fixing this as well.

OmgImAlexis:
@mcollina what's the go with this as the PR associated is closed.

mcollina:
In doubt, close ;). Those are extremely hold.

PR #683, opened 2017-09-13
[#682] Fix bug with reschedulePing causing a dropped connection to ne…

duncanfinney:
…ver be realized

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -97,8 +97,9 @@ function MqttClient (streamBuilder, options) {
   // map of a subscribe messageId and a topic
   this.messageIdToTopic = {}
 
-  // Ping timer, setup in _setupPingTimer
-  this.pingTimer = null
+  // Ping timers, setup in _setupPingTimers
+  this.sendPingTimer = null
+  this.checkPingTimer = null
   // Is the client connected?
   this.connected = false
   // Are we disconnecting?
@@ -164,7 +165,7 @@ function MqttClient (streamBuilder, options) {
   })
 
   // Setup ping timer
-  this.on('connect', this._setupPingTimer)
+  this.on('connect', this._setupPingTimers)
 
   // Send queued packets
   this.on('connect', function () {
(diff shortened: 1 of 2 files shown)


duncanfinney:
Fixes #682

mcollina:
I am 👎  with this change as is.

The spec reports:

> If a Client does not receive a PINGRESP Packet within a reasonable amount of time after it has sent a PINGREQ, it SHOULD close the Network Connection to the Server. (Section 3.1.2.10 - Keep Alive).

So, we should schedule the PINGRESP to be checked not at any given interval, but rather after a reasonable amount of time after a PINGREQ has been set.

duncanfinney:
I actually read the same thing with a college from work today and we came to a similar conclusion. We also thought the wording was a little vague.

This is what it is saying right? 

```
Whenever any control packet (including pingresp) arrives from the server:
   - reschedule _performServerPingCheck() at T=current_time + keepalive
```
    
```
function _performServerPingCheck() {
     sendAPing()
     setTimeout(() => {
        if (haven'tGottenAPingSinceISentIt()) { closeConnection() }
    }, opts.pingRTTMaximumMillis); 
}
```

We didn't know how to word the configuration or what the default value should be. What would be "a reasonable amount of time"? I was thinking 5 seconds.

mcollina:
My original implementation was: if you didn't receive one before sending the next one, then it's probably time to drop. I don't see what is wrong with that, and how it can cause problems (the code might, the principle seems sound).

duncanfinney:
It doesn't cause problems if reschedulePings is turned off. 

But if you have it set to true, anytime you publish a message it moves the check farther into the future (even when the broker has torn down the connection on the other end). 

https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L728

Since we publish messages faster than the keepalive period always, we never would hit `_checkPing()`

mcollina:
the timer is shifted only if the client is connected, as https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L711 dictates. 

The goal is to never hit `_checkPing()`, that's the point. If we are sending data and the connection has not been severed, we should not send, receive or check PINGREQ and PINGRESP.

duncanfinney:
That was the thing. The MQTT client isn't aware that the connection has been severed by the server because it was disconnected. It waits for the TCP to timeout which seems to be a really long time. 

There might be something wrong with the TCP settings... this is on a Linux Yocto build.

mcollina:
Have you tried setting the timeout on the socket itself with https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback?

Maybe we should provide that as option, or maybe set it automatically a 2x keepAlive.

arihantdaga:
Any news on this ??

mcollina:
This PR can be closed.
I think the suggestion in https://github.com/mqttjs/MQTT.js/pull/683#issuecomment-330186022 still holds.

PR #689, opened 2017-10-03
fix use of should.Array()

boneskull:
When running the test suite, I'm getting:

```
TypeError: This assertion does not expect any arguments. You may need to check your code
    at Assertion.assertZeroArguments (node_modules/should/cjs/should.js:280:13)
    at Assertion.<anonymous> (node_modules/should/cjs/should.js:1261:10)
    at Assertion.value [as Array] (node_modules/should/cjs/should.js:337:14)
    at test/abstract_client.js:323:27
    at MqttClient.subscribe (lib/client.js:488:5)
    at test/abstract_client.js:319:16
...etc
```

This fixes the problem.

Not sure if `should` broke semver here, but the installed version is ~~13.1.0~~ 13.1.1.

**UPDATE**:

[The last build failed](https://travis-ci.org/mqttjs/MQTT.js/builds/282832614) due to reasons unknown--I believe it's indicative of a bug either in `websocket-stream` or `ws`.

I've added 0d2dc50 which fixes this failing test, but *I don't really know why*.  The Node.js docs [suggest that `socket#destroy()` is not appropriate](https://nodejs.org/api/net.html#net_socket_destroy_exception) where we were using it, since there's not necessarily an "error" involved.

It may make sense to call `socket#destroy()` in some circumstances, but an explicit `client.end(true)` is *not* one of them.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -678,7 +678,7 @@ MqttClient.prototype._cleanUp = function (forced, done) {
   }
 
   if (forced) {
-    this.stream.destroy()
+    this.stream.end()
   } else {
     this._sendPacket(
       { cmd: 'disconnect' },
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -320,7 +320,7 @@ module.exports = function (server, config) {
           if (err) {
             return done()
           }
-          granted2.should.Array([])
+          granted2.should.Array()
           done()
         })
       })


boneskull:
I am seeing failures in the "queue offline messages" check with or without this change in Node.js v8.5.0:

```
  1) MqttClient offline messages should queue message until connected:
     Uncaught TypeError: Cannot convert undefined or null to object
      at Function.keys (<anonymous>)
      at Store.createStream (lib/store.js:9:1228)
      at MqttClient.<anonymous> (lib/client.js:9:5378)
      at MqttClient._handleConnack (lib/client.js:9:33562)
      at MqttClient._handlePacket (lib/client.js:9:13729)
      at process (lib/client.js:9:11215)
      at Writable.writable._write (lib/client.js:9:11676)
      at doWrite (node_modules/readable-stream/lib/_stream_writable.js:406:64)
      at writeOrBuffer (node_modules/readable-stream/lib/_stream_writable.js:395:5)
      at Writable.write (node_modules/readable-stream/lib/_stream_writable.js:322:11)
      at Socket.ondata (_stream_readable.js:642:20)
      at addChunk (_stream_readable.js:266:12)
      at readableAddChunk (_stream_readable.js:253:11)
      at Socket.Readable.push (_stream_readable.js:211:10)
      at TCP.onread (net.js:585:20)
```

boneskull:
I’m not sure. they fail for me after a fresh clone & pull on 8.5.0

boneskull:
@mcollina 

please see the [updated description](https://github.com/mqttjs/MQTT.js/pull/689#issue-262512297).

lib/client.js
```diff
@@ -678,7 +678,7 @@ MqttClient.prototype._cleanUp = function (forced, done) {
   }
 
   if (forced) {
-    this.stream.destroy()
+    this.stream.end()
```
mcollina: I would prefer this to stay as `destroy()`, or are you seeing test failures because of it? `destroy()` ensures no more data is sent in the socket.

boneskull:
it’s directly responsible for the failures. yet I’m no expert on this, I can only read the docs and come to the conclusion that destroy is only necessary if we have an error

mcollina:
This is fixed on master, thanks.

PR #690, opened 2017-10-03
remove unused, undocumented "disconnected" property from MqttClient

boneskull:
While probing around for the root cause of another bug, I discovered the `disconnected` property, which is incongruous with the documented `connected`, and also seemingly unnecessary.  

Prior to this change, situation arises where the `connected` property is `true` even if a call to `end()` has completed, which seems incorrect.  This *may* be the source of my bug, but haven't tested yet :smile:

The changes below don't seem to affect the test outcomes, since the `disconnected` property is not referenced anywhere other than the two places seen in this diff.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -117,10 +117,6 @@ function MqttClient (streamBuilder, options) {
 
   // Mark connected on connect
   this.on('connect', function () {
-    if (this.disconnected) {
-      return
-    }
-
     this.connected = true
     var outStore = null
     outStore = this.outgoingStore.createStream()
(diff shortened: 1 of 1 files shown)


mcollina:
Tests are failing on CI, are they related?

boneskull:
don’t believe so; I’m experiencing a lot of random flake running the tests.

boneskull:
I think a proper finite state machine might be a good idea.

mcollina:
Definitely.

boneskull:
I'm gonna withdraw this then

PR #693, opened 2017-10-10
update eslint to avoid a deprecated option

Tabrizian:
```space-after-keywords``` has actually been deprecated. This new configuration uses new option.
Hope it helps :)


mcollina:
We are using standard, and not really the eslint file. Would you mind removing it completely?

Tabrizian:
Job is done!

Issue #579, opened 2017-03-30
After calling mqttClient.end() also client is trying to reconnect


sowmyasoora:
I wanted to disconnect mqttClient when i am logging out of my application. To disconnect i am using mqttClient.end(). client is trying to reconnect even after calling mqttClient.end().

Thanks in advance

mcollina:
Can you please upload an example?

sowmyasoora:
@mcollina Thanks for reply.

I am trying to connect mqtt broker using user credentials once i logged into my app. Once after user session expired i am using mqttClient.end() to disconnect mqtt client. But still mqtt client is trying to reconnect. As user credentials are invalid by that time it never connects to broker. I wanted to stop reconnect once i force closed my mqtt connection. Please help me out how can i achieve this?

Is there any way i can set an option for retry count?

mcollina:
after you call `.end()`, it should not try to reconnect.

sowmyasoora:
Hi @mcollina , .end() is working fine. No worries. Its my mistake.  Thanks for your quick response.

I have another issue. When i am trying to connect to my MQTT broker if my server have more cookies (cookies size is around 950 bytes), not able to establish web socket connection with error "**failed: Connection closed before receiving a handshake response**". If cookies are less every thing was proper and working fine. Do you have any idea?

mcollina:
I'm closing this issue.
I do not know about the second error, it might be related to your application. Feel free to open another issue and attach a way to reproduce the problem.

boneskull:
> after you call .end(), it should not try to reconnect. I fear it's a bug, would you mind uploading an example so we can replicate?

I'm experiencing something like this; I'll try to isolate a reproducible case.

boneskull:
so [here](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L223):

```js
this.on('close', this._setupReconnect);
```

Calling `MqttClient#end()` will eventually result in the end of the client's `stream`, so when that happens, [this listener](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L276) fires:

```js
eos(this.stream, this.emit.bind(this, 'close'))
```

...which invokes `_setupReconnect`, which reconnects again...

boneskull:
Anyhow, I mentioned in #690 that a finite state machine might be in order to sort this out, because we have too many booleans flying around in here.

boneskull:
[...]

PR #695, opened 2017-10-11
show example of promisification failure

boneskull:
as [I mentioned](https://github.com/mqttjs/MQTT.js/issues/579#issuecomment-335876893) in #579, this illustrates a problem I'm seeing with disconnection.  see the build failure

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "mocha": "^3.5.3",
     "mqtt-connection": "^3.0.0",
     "nsp": "^2.8.0",
+    "pify": "^3.0.0",
     "pre-commit": "^1.2.2",
     "rimraf": "^2.6.2",
     "should": "^13.0.1",
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -8,6 +8,7 @@ var sinon = require('sinon')
 var mqtt = require('../')
 var xtend = require('xtend')
 var Server = require('./server')
+var pify = require('pify')
 var port = 9876
 
 module.exports = function (server, config) {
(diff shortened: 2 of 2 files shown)


RangerMauve:
I guess we should be waiting for the disconnect event before calling the callback as well,.

boneskull:
@RangerMauve I'm suspicious about whether or not `connect()` necessarily waits a tick before emitting `connect`.

RangerMauve:
@boneskull `connect` is emitted when we get the MQTT packet for [connack](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L825), since it requires a network trip it'll be guaranteed to be async.

boneskull:
> connect is emitted when we get the MQTT packet for connack, since it requires a network trip it'll be guaranteed to be async.

that's what I'd assume, but maybe my debugger is messed up...

boneskull:
yeah I think this is my misuse of `pify`

boneskull:
`pify` does not mutate

RangerMauve:
Yeah, if `connect()` in the tests is connecting to a local server, then it should be async. Don't see any way for a debugger to make it sync. We have a whole bunch of listeners for `connect` in the constructor, maybe they're interfering with each other or with the connect listener you've added in the `beforeEach`? Not sure if that makes sense, though.

Could you describe again what the problem you're seeing is? Does it time out or throw an error? Or was it that it's auto-reconnecting?

boneskull:
the root problem was auto-reconnection, but I think that's because of naive promisification on my part.  

(also there's the case where `disconnecting` is never set to `false` once it's been `true`--likely another issue entirely with as-of-yet unknown consequences)

PR #697, opened 2017-10-13
Fixed test failures encountered while running test cases, and allowing 'puback's to be sent after successfully processing message handlers

prabathabey:
This PR fixes 3 test failures caused by improper use of should.js assertion library to check for an empty array.


lib/client.js
```diff
@@ -874,15 +874,22 @@ MqttClient.prototype._handlePublish = function (packet, done) {
...
+        this.handleMessage(packet, done)
+        // send 'puback' if the above message handlers are executed
+        // successfully.
+        this._sendPacket({
+          cmd: 'puback',
+          messageId: mid
+        })
+      } catch (e) {
+        done()
+      }
```
mcollina: remove this try-catch.

We should also allocate a closure:

```js
this.emit('message', topic, message, packet)
this.handleMessage(packet, (err) => {
  if (err) {
    return done(err)
  }
  this._sendPacket({
     cmd: 'puback',
     messageId: mid
  })
})
```

prabathabey: @mcollina Thanks for taking time to review the PR and the feedback. I've added the fixes proposed and submitted.

prabathabey:
All test cases seem to successfully get executed locally. I'm in the process of further investigating the root cause of the above reported test failure.

test/abstract_client.js
```diff
@@ -780,6 +780,44 @@ module.exports = function (server, config) {
...
+          serverClient.publish({
+            messageId: messageId,
+            topic: topic,
+            payload: payload,
+            qos: qos
+          })
+          setTimeout(
+            function () {
+              (pubackReceived === true) ? done('Unexpected `puback` received') : done()
+            }, 1000)
```
mcollina: can you rewrite the test to avoid the setTimeout?

test/abstract_client.js
```diff
@@ -780,6 +780,44 @@ module.exports = function (server, config) {
...
+      var messageId = Math.floor(65535 * Math.random())
+      var payload = 'test'
+      var topic = 'test'
+      var qos = 1
+
+      var pubackReceived = false
+
+      client.handleMessage = function (packet, callback) {
+        callback(new Error('Error thrown by the application'))
+      }
```
mcollina: as we are erroring, should a PUBACK be sent anyway?

test/abstract_client.js
prabathabey: Sure. Will do.

test/abstract_client.js
prabathabey: IMHO, in an event where the handleMessage throws an error (could be due to the unavailability of the backend system that persists incoming messages, for example) we need to let the broker retry the messages at a later point. If we do send a PUBACK even amidst an error thrown by the application, does it not send a wrong indication to the broker that the message has successfully been processed by the application? WDYT?

prabathabey:
@mcollina Sure. I will take a look at extending the same for QoS 2 messages as well.

test/abstract_client.js
mcollina: You are right, no PUBACK if handleMessage errors.

test/abstract_client.js
```diff
@@ -780,6 +781,73 @@ module.exports = function (server, config) {
...
+    it('should not send a `puback` if the execution of `handleMessage` fails for messages with QoS `1`', function (done) {
+      var client = connect()
+
+      client.handleMessage = function (packet, callback) {
+        callback(new Error('Error thrown by the application'))
+      }
+
+      client._sendPacket = sinon.spy()
+
+      client._handlePublish({
```
prabathabey: @mcollina The only possible way I could think of for writing a test case (without using setTimeouts) against this particular functionality is by directly manipulating the '_handlePublish' private method, which is sub-optimal for obvious reasons. I'd appreciate if you could shed some light as to if there's any other better way to handle this.

test/abstract_client.js
```diff
@@ -7,6 +7,7 @@ var should = require('should')
 var sinon = require('sinon')
 var mqtt = require('../')
 var xtend = require('xtend')
+var assert = require('assert')
```
mcollina: we use should in this module.

test/abstract_client.js
mcollina: this is ok.

prabathabey: Thanks for the feedback.

test/abstract_client.js
prabathabey: Noted. I will refactor all references that already uses "assert" to use "should".

prabathabey: This is done.

mcollina:
Unfortunately 

```
1) should not send a `pubrel` if the execution of `handleMessage` fails for messages with QoS `2`
```

is very flaky on my box. Can you please send a PR to address the problem?

prabathabey:
Sure, I will check on this.

prabathabey:
I've looked into to the above issue, but unfortunately couldn't reproduce it even when I attempted running the test suite on a few hosts with different configurations including a t2.micro on AWS). Do we have a checklist of items that we use as part of this project to mark a test flaky? I'd appreciate any hint on how you've reproduced the issue, so I can try the same locally and try to isolate if this issue was a result of a regression introduced by the fix I've added earlier or something else.

mcollina:
I just run `npm test` on Mac OS X.

```
  357 passing (57s)
  3 failing

  1) MqttClient publishing should not send a `pubrel` if the execution of `handleMessage` fails for messages with QoS `2`:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  2) MqttSecureClient publishing should not send a `pubrel` if the execution of `handleMessage` fails for messages with QoS `2`:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  3) Websocket Client publishing should not send a `pubrel` if the execution of `handleMessage` fails for messages with QoS `2`:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

prabathabey:
Thanks. Managed to reproduce the issue on Mac OS X. I will fix this and send a PR.

prabathabey:
Apologies for the delay in responding. I've re-written the test case to eliminate the intermittent behaviours observed together with some minor improvements to the fix I'd already submitted in #712 . Appreciate if you could review the changes and provide your feedback if the fix needs any modifications.

PR #713, opened 2017-11-07
fix: add error when destroying stream to generate close event even when the network connection is down

### LINKED ISSUE #710: end() callback not called if `force` is true and the stream is already closed (eg. after a failed keepalive ping)
Hi,
thanks for putting out such a great library!

In a case where the network connection is disconnected and MqttClient.end() is called like so:
```js
var client = mqtt.connect(/*...*/);
// disconnect the network adapter...
// the 'close' event is fired by the client (after a failed keepalive ping for example)
// try to force close the client now:
client.end(true, function () { 
// something to do within the callback
});
```

The callback will never be called because the stream that is being destroyed in Client._cleanUp() [here](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L695) will never fire either a `close` or an `error` event unless an error is passed as an argument to `stream.destroy()`

Does that mean a user should not rely on the callback when calling `end()` with `force` set to `true`? 

note: the `close` event from the socket that is forwarded by the client has already been emitted correctly - maybe I should not call `end()` after receiving a `close` event? what's the recommended usage pattern here?

Happy to submit a PR if you think the error should be emitted and the callback should be called.

pierreca:
Fixes #710 

When a stream is closed because the network connection, the close even fires correctly but after that if a user calls `client.end()` with a callback. that callback will never get called. As far as I understand, that is because the stream object will only fire another close event (which is required for the callback to be called) if `stream.destroy()`. is called with an error argument.

Tried to find a way to simulate the conditions in unit tests but could only repro the behavior with an actual network disconnection... so no specific test. existing tests don't seem to bother though.


mcollina:
A unit test would be really good at this point. You can at least add a test that checks that an `'error'`  event is emitted on the stream.

mcollina:
Is there any update to this one?

pierreca:
@mcollina Sorry, slammed by work - I hope I can get back to that and add the test later, although I'm not sure I understand yet what exactly you think it'll test (I have to spend more time looking at it)

feel free to close if you don't like idling issues.

mcollina:
More or less I want  a test that verified that `'error'` is emitted on the stream if `this.end(true)` is passed in, and such callback is called.

Also, an example showing that such callback is indeed __not__ called without this change.

OmgImAlexis:
@pierreca any chance on the conflicts being resolved? If not can this please be closed and the associated issue updated so someone else can take a stab at this.

lib/client.js
```diff
@@ -1155,7 +1155,7 @@ MqttClient.prototype._cleanUp = function (forced, done) {
       flush(this.outgoing)
     }
     debug('_cleanUp :: (%s) :: destroying stream', this.options.clientId)
-    this.stream.destroy()
+    this.stream.destroy(new Error('stream forcefully closed by the client'))
```
robertsLando, line 1158: Please add a test that reproduces the issue

robertsLando (changes requested):

lib/client.js
robertsLando, line 1158:
```suggestion
    // pass an error here to be sure the 'error' event is emitted on stream, fixes #713
    this.stream.destroy(new Error('stream forcefully closed by the client'))
```

robertsLando:
It's not 100% clear to me why this happens. I mean based on nodejs docs the `close` event should be emitted every time destroy is called on a stream, it's just the `error` event that is emitted too if an error is passed. I keep this opened for now as I got some errors too in the past with `client.end` and I'm not sure if this could be a solution to that or nope

robertsLando:

PR #719, opened 2017-11-15
add support to allow empty string clientId

ophentis:
I am currently using a service that require clientId to be empty.
I modified the option to clientId to allow it to be "" and create a default when it is not `string`.
According to the spec, it should only be set empty along with clean session. 
I can add that part of checking if necessary.

### DIFF
--- a/test/mqtt.js
+++ b/test/mqtt.js
@@ -40,6 +40,20 @@ describe('mqtt', function () {
       c.options.should.have.property('username', 'user')
     })
 
+    it('should return an MqttClient with the clientid with random value', function () {
+      var c = mqtt.connect('mqtt://user@localhost:1883')
+
+      c.should.be.instanceOf(mqtt.MqttClient)
+      c.options.should.have.property('clientId')
+    })
+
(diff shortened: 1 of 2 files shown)


mcollina:

ophentis:
@mcollina thanks, I added the test.

test/mqtt.js
```diff
@@ -40,6 +40,20 @@ describe('mqtt', function () {
       c.options.should.have.property('username', 'user')
     })
 
+    it('should return an MqttClient with the clientid with random value', function () {
+      var c = mqtt.connect('mqtt://user@localhost:1883')
+
+      c.should.be.instanceOf(mqtt.MqttClient)
+      c.options.should.have.property('clientId')
```
mcollina: Can you call `c.end()` everywhere?

ophentis: I'll add to every cases? I think except one.

Llorx:
HM, shit, just created a PR with this haha. Srry for the doublepost.

mcollina:
Thanks!!!

PR #724, opened 2017-11-22
fix a bug that you can not reconnect on wechat

johnzeng:
On wechat, if the socket is closed, the socketOpen mark is not set to false , so you might send data before the websocket is opened, and the following codes will result in error:

```
function sendSocketMessage (msg) {
  if (socketOpen) {
    wx.sendSocketMessage({
      data: msg
    })
  } else {
    socketMsgQueue.push(msg)
  }
}
```

BTW, why are we pushing msg into a msg queue when the socket is not opened? If we send some message to socket before mqtt connection is done, those messages will be send to server, will it result in error?

### DIFF
--- a/lib/connect/wx.js
+++ b/lib/connect/wx.js
@@ -15,7 +15,6 @@ function sendSocketMessage (msg) {
 }
 
 function WebSocket (url, protocols) {
-
   var ws = {
     OPEN: 1,
     CLOSING: 2,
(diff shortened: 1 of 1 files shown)


mcollina:
I’m not really familiar with the wechat support. We can’t really test it, so it’s best effort. I really don’t have an answer for you :/.

PR #732, opened 2017-12-04
Fix in sendPacket to return error in the callback if qos > 0

v-divya110:
The sendPacket had cb as undefined in case of publish with qos > 0, reassigned the cb before returning in case of error.

### DIFF
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -8,6 +8,7 @@ var sinon = require('sinon')
 var mqtt = require('../')
 var xtend = require('xtend')
 var Server = require('./server')
+var Store = require('./../lib/store')
 var port = 9876
 
 module.exports = function (server, config) {
@@ -794,7 +795,8 @@ module.exports = function (server, config) {
         messageId: Math.floor(65535 * Math.random()),
         topic: 'test',
         payload: 'test',
-        qos: 1}, function (err) {
+        qos: 1
+      }, function (err) {
         should.exist(err)
       })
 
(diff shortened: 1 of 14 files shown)


test/abstract_store.js
```diff
@@ -130,4 +130,18 @@ module.exports = function abstractStoreTest (build) {
...
+    const store = new Store()
+    store.put = function (packet, cb) {
+      process.nextTick(cb, new Error('oops there is an error'))
+    }
+
+    store.put({messageId: 43}, function (err) {
+      if (err) {
+        done()
+      }
+    })
```
mcollina: this is not testing what you think you are testing. As you have modified the client, you should test the client, not the store.

mcollina:
CI didn't run for the last commit (not sure why). Can you also fix the conflicts with master?

test/abstract_client.js
```diff
@@ -450,56 +449,70 @@ module.exports = function (server, config) {
...
+    })
 
-      client.on('connect', function () {
-        client.subscribe('test', function () {
-          subscribeCalled = true
-        })
-        client.publish('test', 'test', function () {
-          client.end(false, function () {
-            subscribeCalled.should.be.equal(true)
-            done()
```
mcollina: why all these changes in the test? Can you only add the new one?

v-divya110: I did not add any of those it only got indented. The test did not run otherwise

mcollina: Can you please restore them as they where on master?

v-divya110: umm okay.. so you want me to revert the commit ?

v-divya110: I don't get what I should be doing given that the current master test/abstract-client has many more changes in it. Should I simply revert the merge and commit it again with my addition in the test script?

mcollina: Yes, that would be the preferred path.

v-divya110: okay so I reverted the merge and checked in my code again.

mcollina:
Can you please check in https://github.com/mqttjs/MQTT.js/pull/732/files?
I think there are still too many changes in there.
You might be better off in doing another branch and cherry pick your commits.

v-divya110:
yes I think I'll probably close this branch and create a new branch followed by a new pull request. There seem to be too many changes in this and I don't see a way of restoring it. It'll be easier to create a new branch instead

Issue #740, opened 2017-12-12
Inconsistency with semicolons in Typescript declaration files


TechnicalSoup:
I've noticed that the general coding style for this project is to omit optional semicolons, however I've noticed that some of the Typescript declaration files have semicolons in a few locations.

An example of this from the types.d.ts file is shown below
```javascript
 will?: {
    topic: string;
    payload: Buffer;
    qos?: QoS;
    retain?: boolean;
  }
```

In the majority of Typescript declaration files I've looked at the semicolons are not omitted (seems to be less prevalent in Typescript code in general). I'm happy to create a PR to correct this to keep it consistent across the file, and project in general but thought I'd raise an issue first to see if there was a preference to steer the declaration files to include semicolons.

mcollina:
Go ahead and propose a change that make sense. We follow standard in the rest of the codebase.

Is there a linter to catch these issues?

4rzael:
I haven't tried, but [tslint](https://palantir.github.io/tslint/) can probably be used for that, with the [semicolon](https://palantir.github.io/tslint/rules/semicolon/) rule

TechnicalSoup:
Yep - I've used tslint in the past and it has a semicolon rule that can applied as `always` where semicolons must be used and `never` where they must only be used when necessary.

I'm happy to propose changes to the declaration files, and even create a tslint.json file with some rules in it if you are looking to enforce a coding style for Typescript aspects of the project. There is a github repo that attempts to mirror the javascript coding standard in a tslint [here](https://github.com/blakeembrey/tslint-config-standard)

My personal preference for Typescript code is to include the semicolons as it seems to be more common when browsing through example code and other repos, however as the rest of this project omits them in the javascript code that may seems a little odd. 


If @mcollina has a preference I'm happy to head in that direction.

sublimator:
I'd say it is actually less common to have semicolons in TypeScript code,
though that might just be my preference speaking :) The fact that I don't
even know clearly speaks to the invisibility/redundancy of them. Given the
REQUIRED compilation step, it's pretty safe to omit semicolons in code, as
the compiler will pick up the few rare cases when they are actually
required ( wrapping expressions in parentheses )

I'd vote for consistently none :)

mcollina:
I abstain: I’m very unfamiliar with Typescript.

sublimator:
I don't really care either way :)

TechnicalSoup:
Thanks for the input @sublimator. 

[...]

PR #748, opened 2018-01-07
Remove unnecessary semicolons from TS declarations

TechnicalSoup:
Remove unnecessary semicolons from some of the Typescript declaration files - mostly where there are nested objects and arrays in the interfaces.

Addresses issue #740

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -10,7 +10,7 @@ export interface IClientOptions extends ISecureClientOptions {
   protocol?: 'wss' | 'ws' | 'mqtt' | 'mqtts' | 'tcp' | 'ssl' | 'wx' | 'wxs'
 
   wsOptions?: {
-    [x: string]: any;
+    [x: string]: any
   }
   /**
    *  10 seconds, set to 0 to disable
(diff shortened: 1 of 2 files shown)


mcollina:
can we add a check in tslint to validate that there are no semicolons?

TechnicalSoup:
@mcollina Yes there is, and it already included in the project. The `tslint` and `tslint-config-standard` modules are dev dependencies. The `tslint-config-standard` has a rule which checks for semicolons, and I expected that it would have picked these up. However there seems to be some inconsistency in the way this rule is applied for interfaces.

The linter will pick up the unnecessary semicolon for the cmd parameter in the example code below:
```javascript
export interface IConnectPacket extends IPacket {
  cmd: 'connect';
  clientId: string
  protocolVersion?: 4 | 3
}
```

but will not pick up the semicolons for the will object in the code below:
```javascript
export interface IConnectPacket extends IPacket {
  cmd: 'connect'
  clientId: string
  protocolVersion?: 4 | 3
  will?: {
    topic: string;
    payload: Buffer;
    qos?: QoS;
    retain?: boolean;
  }
}
```
From what I can gather this seems to be because the interfaces and type definitions in these files don't actually transpile directly to javascript code, and are a tool to define the API for anyone who wants to use the module in a Typescript project.

mcollina:
Thanks for the explanation!

Issue #703, opened 2017-10-19
Upgrade to mocha@4


mcollina:
Our test suite does not complete/exits anymore after the upgrade. It would be fantastic if someone could help.

prabathabey:
@mcollina I'd like to take a look at this. Will provide an update as soon as I manage to come up with a sizeable contribution.

boneskull:
I looked at this and can identify several problems:

- The reconnect timer is either really sticky or just needs an `unref()`; the interval will keep the process open. (I'm still not convinced the reconnect timer is working as intended, but that's a separate issue)
- The ping timer suffers from the same problem
- Many tests don't call `client.end()`, and if they do, they don't necessarily wait for it to finish
- The test server(s) themselves stay open

You can test this by grabbing [wtfnode](https://npm.im/wtfnode) and executing `wtfnode node_modules/.bin/_mocha` from a MQTT.js working copy.  When the process hangs, hit Ctrl-C to dump the async tasks.

I poked at this for about an hour, but IMO you're better off sticking `--exit` in `mocha.opts` unless you love pain.

boneskull:
FWIW, calling `close()` on a `net.Server` instance doesn't disconnect any clients, which I didn't realize...

mcollina:
Oh I didn’t know you could add a flag to mocha.opts to get the old behavior.
I would still prefer if someone could come up and clean up the current mess.

boneskull:
not to toot my :trumpet:, but if I'm finding it difficult & tedious to clean up, I don't suppose it'll be a breeze for anyone else :wink:

4rzael:
I tried to do some kind of tests, where I try the connection against multiple brokers:

This is my code:
```js
const should = require('should')
const mqtt = require('../../')

describe('protocol ws', () => {
  let client
  afterEach(() => {
    if (client) {
      client.on('close', () => {
        console.log('closed')
      })
      client.end(true)
      client = undefined
    }
  })

  it('should be able to connect to the broker', (done) => {
// in my case, urls are ws://localhost:2002, ws://localhost:2012 and ws://localhost:2022/mqtt, respectively for mosca, mosquitto, and emq.
    client = mqtt.connect(URL) 
    client.on('connect', () => {
      client.connected.should.equal(true)
      done()
    })
  })
})
```

## Broker configuration

I tried running it with mosca, mosquitto and emq, using the following docker images to run the brokers:

* mosca: `matteocollina/mosca`
[...]

PR #751, opened 2018-01-17
update mocha to v4.x

boneskull:
- uses `--exit` to hack around non-exiting process
* * *
This is a temporary fix for #703.

*Actual* fixes as [mentioned here](https://github.com/mqttjs/MQTT.js/issues/703#issuecomment-357872112) are another issue entirely.


mcollina:
Thanks!!!

Issue #505, opened 2016-11-28
Integration tests on Github Actions for all OSS brokers


mcollina:
It would be awesome to receive a contribution to do some integration testing against the major OSS brokers.

Something using Travis ENV support, or similar. It might even be a separate repo, feel free to discuss!

ralphtheninja:
@mcollina Do we have any integration tests implemented?

mcollina:
zero. But I think we should.

sandro-k:
I am on board.. Lets start with something simple like Mosca and expand from there.

mcollina:
I have no time to coordinate this :(. I would suggest we should start with Mosquitto and Mosca. Please go ahead and do it, I will be happy to review.

4rzael:
I completely agree on this. I don't have time to help right now, but I'll be happy to do so when possible.

sandro-k:
I am trying to free some time next week.

anhldbk:
Would you guys elaborate more on how we take integration tests? I think we should make a to-do list now. For example:

- [ ] Listing all OSS Brokers
- [ ] Specify what to test ...

sandro-k:
Thanks @anhldbk for getting this started again.

# Brokers
* Mosquitto
* Mosca

# What to test
* Connect 
  * SSL/TLS
* Subscribe
  * QOS 0,1,2
  * Retained 
* Publish
  * Retained
* LWT

# Platform 
* Node.js
* Browser

anhldbk:
Currently, I am using [emq broker](http://emqtt.io/) and find it great. It's written in Erlang (suitable for production environments). 

About `What to test`, should we have tests for  `MQTT Over WebSocket`?

mcollina:
@anhldbk I think so, yes.

mcollina:
It would be awesome to have integration tests also via SauceLabs.

anhldbk:
The integration test results should be organized into a report with:

1. Information about brokers
2. For each broker, we should have describe the running environment. It must be reproducible. (Docker may be a great relief)
3. Associated test scripts
4. Finally, a summary

@sandro-k , @mcollina is that ok?

mcollina:
@anhldbk yes! Go ahead and make it docker-based.

Bonus points (but keep it separate): websockets in browsers.

YoDaMa:
changed the name since this is still a relevant ask but we're using Github Actions now.

PR #754, opened 2018-01-26
Started brokers integration tests

4rzael:
I started working on the issue #505.

This PR allows travis-ci to run tests on different brokers: EMQ, mosca and mosquitto, which are each in their own docker containers, in order to improve reproducibility.

However, currently, I only have **very few** tests.

You can either choose to merge this PR, for other people to add tests later, or ask for tests in the comments, and I can try to implement them while I have the time to do so.

#### PS: First, I will try to fix the fact that the tests fail on EMQ in travis


mcollina:
I think the best way would be to start separate environments, so that we test mosquitto/mosca/eqmttd independently.

YoDaMa:
closing since we don't use Travis any more.

PR #763, opened 2018-02-06
fix issue while ending a disconnected client

gautaz:
Hello,

This is mainly to be able to handle cleanly situations where the client has been misconfigured.
I've added a test showing the issue that I am experimenting.

It is also partly related to #710 although it does not require a forced end.


gautaz:
@mcollina Thanks !
It can wait a few days, the project using it is a work in progress :-).

Issue #767, opened 2018-02-12
'end' event


gautaz:
Hello,

I was wondering if you (the maintainers) would consider adding an `end` event?

The use case would be multiple parts of some code using the same MQTT connection. Once the connection is asked to `end()` and it actually ends (with or without an error), this event would tell all listeners that the connection has effectively ended and is no longer usable.  The `end` event could be emitted just after the `end()` callback is called.

It is currently difficult to rely on the `close` event which is emitted directly by the underlying communication stream. The `close` event can be emitted multiple times and AFAIK listeners have to check the `disconnecting` attribute in order to tell if the connection is really ending.
It is even more difficult if the connection was never successful, in this particular case, no `close` event  will be emitted after an `end()` call.

I am interested by any proposal on this particular subject and I can send a PR if there is an agreement on how to handle this :-).

mcollina:
Definitely, would you like to send a PR with a proposal?

gautaz:
I am close to send the PR but since I have added additional test cases, I was able to reproduce a bug on my last PR (#763).

Do you want me to fix this in the same PR or do you want me to put this on a separate issue?

mcollina:
Send another PR!

gautaz:
Sorry for the delay, the bug is fixed by #770 and the `end` event is added by #771.

gautaz:
Hello, this issue went completely off my radar.
As the PR were merged, I guess I can close this now, sorry for the delay.

PR #770, opened 2018-02-14
Missing done check in _cleanUp

gautaz:
Here is the PR to fix the issue found here declared in this [comment](https://github.com/mqttjs/MQTT.js/issues/767#issuecomment-365211728).

I have also added a non regression test for this particular case.
Sorry for the mistake, my bad.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -743,7 +743,7 @@ MqttClient.prototype._cleanUp = function (forced, done) {
     this.pingTimer = null
   }
 
-  if (!this.connected) {
+  if (done && !this.connected) {
     this.stream.removeListener('close', done)
     done()
   }
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -1790,6 +1790,11 @@ module.exports = function (server, config) {
       })
     })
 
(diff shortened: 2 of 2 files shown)


gautaz:
Wow, that's strange, my modification has nothing to do with QOS.
@mcollina Any idea on what happened ?

mcollina:
I think it's just the CI that is flaky.

gautaz:
OK, I'll close and reopen the pull request just to restart the build.

mcollina:
@gautaz this seems to be breaking Node v4.3.2, which we still support because of AWS Lambda.

gautaz:
OK, I will try to reproduce in a Docker environment with node v4.3.2.
I'll keep you posted.

gautaz:
@mcollina I have tried to reproduce multiple times with [nodesource/wily:4.3.2](https://hub.docker.com/r/nodesource/wily/) but I did not succeed.

In the meantime all tests pass now on Travis. I guess I'll have to find a slower platform to reproduce.
Do you still want to hold on this issue or is it OK for now ?

PR #771, opened 2018-02-14
end event added and tested

gautaz:
Here is the support for an `end` event #767.

The additional tests may sometime reproduce the bug fixed by #770.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -588,7 +588,12 @@ MqttClient.prototype.end = function (force, cb) {
   function closeStores () {
     that.disconnected = true
     that.incomingStore.close(function () {
-      that.outgoingStore.close(cb)
+      that.outgoingStore.close(function () {
+        if (cb) {
+          cb.apply(null, arguments)
+        }
+        that.emit('end')
+      })
     })
     if (that._deferredReconnect) {
       that._deferredReconnect()
--- a/README.md
+++ b/README.md
@@ -275,6 +275,14 @@ Emitted when the client goes offline.
 Emitted when the client cannot connect (i.e. connack rc != 0) or when a
 parsing error occurs.
 
+#### Event `'end'`
+
+`function () {}`
+
+Emitted when <a href="#end"><code>mqtt.Client#<b>end()</b></code></a> is called.
+If a callback was passed to `mqtt.Client#end()`, this event is emitted once the
+callback returns.
+
 #### Event `'message'`
 
 `function (topic, message, packet) {}`
--- a/test/abstract_client.js
+++ b/test/abstract_client.js
@@ -73,6 +73,45 @@ module.exports = function (server, config) {
       })
     })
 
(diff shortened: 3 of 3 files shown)


gautaz:
OK, here tests seem to fail due to the bug fixed by #770.
They did not fail at first on my development platform, it must be related to the way asynchronous operations interleave.

@mcollina Once #770 is OK, I will merge it in my branch before asking for the pull request again.

lib/client.js
```diff
@@ -588,7 +588,12 @@ MqttClient.prototype.end = function (force, cb) {
   function closeStores () {
     that.disconnected = true
     that.incomingStore.close(function () {
-      that.outgoingStore.close(cb)
+      that.outgoingStore.close(function () {
```
mcollina: this can pass an error as well.

gautaz: You are right, I have blindly only considered initially the default store implementation but I have also to consider any external implementation.
I will pass any argument as is (to the `end` event as well).

mcollina: no error to `'end'` please.

gautaz: Done with no error on `end` event :-).
I have also added a test to check that the error is passed up to the `end()` callback.

gautaz:
@mcollina The same pattern seems to occur each time the CI build fails:

- some jobs are put on hold by Travis (I guess due to unavailable resources)
- the tests run by these jobs have started and have already initiated a MQTT connection
- when Travis resume these jobs too late, these connections fail with `ECONNRESET`

Is there a way to tell Travis not to start the six jobs at once but limit the number of concurrent jobs to, let say, 3?
This will not resolve the issue but might perhaps mitigate it.

gautaz:
Another way to mitigate the issue would be to increase the connection timeout when the tests are run by Travis.

gautaz:
@mcollina Since the build was failing due to CI issues and not to the PR itself, do you want me to change anything in the PR or is it OK for now ?

I tried to find a way to fix or mitigate the CI issues in the Travis docs but I did not found anything yet.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mqtt",
   "description": "A library for the MQTT protocol",
-  "version": "2.15.2",
+  "version": "2.15.3",
```
mcollina: Generically, I prefer if PRs did not tweak the version number. This is done as part of the release itself.
I'll land it anyway, but please remember for the next time.

mcollina: Oh, this was weird. Can you please rebase and squash your commits? It seems something went wrong with the merge. and a commit of myself is showing up here.

gautaz:
I had to rewrite a bit my branch history to fix the merge issue.
Still hitting the CI fail wall, I'll try again by closing/reopening.

gautaz:
@mcollina The merge issue is fixed and all CI checks are finally OK :-).

gautaz:
Woohoo !
@mcollina Thanks for your feedback and your support!

Issue #765, opened 2018-02-09
Call message callbacks upon disconnection / no more reconnections


periclesroalves:
I noticed that current message callbacks (`MqttClient.outgoing`) are not called with an error when connection suddenly drops. I guess this is expected as the library may reconnect automatically and the ack for those messages can still come back after reconnection, right?

However, these callbacks are not called with an error even when reconnection attempts are exhausted (in this case, to reconnect again the user would have to explicitly call `MqttClient.reconnect()`, effectively reinstantiating the stores). Would this be an accepted/desired behavior to have these callbacks called when reconnections will no longer be attempted by the library?

If not, can we at least call them with an error when the user calls `MqttClient.end()`, as this effectively triggers a cleanup and closes the store?

mcollina:
Oh, good spot, would you or @pierreca be able to help with a PR? Thansks!

pierreca:
@anthonyvercolano is going to take a look at this, thanks!

anthonyvercolano:
Looking at this problem a bit.  My thinking is that code such as the following should probably be inserted after line 725 in client.js (at the beginning of  _cleanup).

  if (this.outgoing) {
    this.outgoing.forEach((cb) => {
      cb(new Error('Connection closed'))
    })
  }

I'm not trying to make this a pull request.  Just hoping to hear an opinion as to whether I need to be doing a lot more work to maintain the other various structures holding information related to packets such as the various stores

pierreca:
The proposed fix for this is in #787

PR #787, opened 2018-03-06
Force callbacks when pingresp not received or when end called with true.

anthonyvercolano:
When a pingresp does not occur, things are forced closed.  However, callbacks in the outgoing queue are not invoked.

Added code to the cleanup, so that in the case of forced argument of true, will invoke the callbacks with  an error.

Also fixed a few spelling errors.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -43,6 +43,17 @@ function sendPacket (client, packet, cb) {
   }
 }
 
+function flush (queue) {
+  if (queue) {
+    Object.keys(queue).forEach(function (messageId) {
+      if (typeof queue[messageId] === 'function') {
+        queue[messageId](new Error('Connection closed'))
+        delete queue[messageId]
+      }
+    })
+  }
+}
+
 function storeAndSend (client, packet, cb) {
   client.outgoingStore.put(packet, function storedPacket (err) {
     if (err) {
@@ -135,7 +146,7 @@ function MqttClient (streamBuilder, options) {
           return
         }
 
-        // Avoid unnecesary stream read operations when disconnected
+        // Avoid unnecessary stream read operations when disconnected
         if (!that.disconnecting && !that.reconnectTimer && that.options.reconnectPeriod > 0) {
           outStore.read(0)
           cb = that.outgoing[packet.messageId]
@@ -722,6 +733,9 @@ MqttClient.prototype._cleanUp = function (forced, done) {
   }
 
   if (forced) {
+    if ((this.options.reconnectPeriod === 0) && this.options.clean) {
+      flush(this.outgoing)
+    }
     this.stream.destroy()
   } else {
     this._sendPacket(
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ test/typescript/.idea/*
 test/typescript/*.js
 test/typescript/*.map
 package-lock.json
+# VS Code stuff
+**/typings/**
+**/.vscode/**
--- a/test/client.js
+++ b/test/client.js
@@ -15,6 +15,14 @@ var Server = require('./server')
 var port = 9876
 var server
 
+function connOnlyServer () {
+  return new Server(function (client) {
+    client.on('connect', function (packet) {
+      client.connack({returnCode: 0})
+    })
+  })
+}
+
 /**
  * Test server
  */
@@ -108,7 +116,7 @@ describe('MqttClient', function () {
       client.end()
     })
 
-    it('should return 1 once the interal counter reached limit', function () {
+    it('should return 1 once the internal counter reached limit', function () {
       var client = mqtt.connect(config)
       client.nextId = 65535
 
@@ -117,7 +125,7 @@ describe('MqttClient', function () {
       client.end()
     })
 
-    it('should return 65535 for last message id once the interal counter reached limit', function () {
+    it('should return 65535 for last message id once the internal counter reached limit', function () {
       var client = mqtt.connect(config)
       client.nextId = 65535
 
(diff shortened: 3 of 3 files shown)


pierreca:
This is a proposed fix for #765 by the way

mcollina:
what would happen if a client connects with clean=false? We are expecting those callbacks to be called when the client reconnects and will be able to deliver those messages.

anthonyvercolano:
As it stands, even with my changes, with a reconnect value > 0 it will indeed attempt to resubmit those packets in the output store.

Would it be acceptable to only flush the packets in outgoing if reconnect = 0?  Should it also delete the "matching" packets in the output store?

With respect to the MQTT potocol, is it EVER reasonable to return with an error some packet with qos > 0?

If packets are allowed to complete with an error, it should be done in _cleanup(true).

mcollina:
> As it stands, even with my changes, with a reconnect value > 0 it will indeed attempt to resubmit those packets in the output store.

what do you mean by `reconnect > 0`?

> Would it be acceptable to only flush the packets in outgoing if reconnect = 0? Should it also delete the "matching" packets in the output store?

Packets in the store should never be deleted if `clean: false`, and with `clean: true` the logic is already there.

> With respect to the MQTT protocol, is it EVER reasonable to return with an error some packet with qos > 0?

only if we are dumping them from the queue, we should call the callback with an error. And only if `clean: true`.

> If packets are allowed to complete with an error, wouldn't cleanup(true) be the best place?

what do you mean?

mcollina:
> Who is doing the reconnect? Is it the reconnect logic built into mqttjs?

The reconnect logic is built in into mqtt.js and it happens automatically.

periclesroalves:
@mcollina, I exposed the same concern about reconnection in the original issue. However, it's always safe to call the callbacks with an error in the following scenario:

Connection was lost AND `reconnectPeriod > 0` AND (`options.clean == true` OR the client did not provide a custom store through `options.outgoingStore`, i.e., he's using the library internal store implementation).

In this scenario, the only way a reconnection can happen is if the user explicitly calls `client.reconnect()`. In this case, the store will either be clean (because of the `clean` flag) or the store is gonna be reinstantiated (because the user did not provide a custom store).

I think @anthonyvercolano's PR is good to go if these simple conditions are added to his code. Also, I think this would be the expected behavior for users that are explicitly telling the library not to reconnect automatically.

mcollina:
@periclesroalves I agree. Would you mind helping getting this to the finishing line?

periclesroalves:
Well, [these](https://github.com/periclesroalves/MQTT.js/commit/96a938feae06b8c9c5d3926d3bb2c0919feb2dcf) minor changes should do it. Maybe @anthonyvercolano want to add it to his PR, so we don't need a new one.

anthonyvercolano:
Left as a comment in the @periclesroalves pr:

Clean is the only thing that should matter here.

It is not at all clear why the outgoing store being supplied by the application should affect the whether the callbacks are called.

If the reconnectPeriod is zero, then the calling application should still be able to call reconnect. The session should still be restored.

A further concern. Pierre and I were looking at the reconnect code and we believe there is a problem in Mqtt.prototype.reconnect. If Mqtt.prototype.reconnect is invoked with no application supplied stores in the option parameter, then the internally created stores will absolutely be wiped out by the creation of new stores. This would seem to violate the clean == false semantics which should cause a publish to be resent on successful reconnect. Is there some documentation that states otherwise?

periclesroalves:
You're right, @anthonyvercolano. What I had in mind was the behavior already implemented in the library, but from the MQTT spec point of view, the only thing that should matter is the `clean` flag.

If we look at the spec, then we have more problems in the code:

> If CleanSession is set to 0, [...] The Client and Server MUST store the Session after the Client and Server are disconnected [MQTT-3.1.2-4].

As you pointed, if the user is using the internal store, the `reconnect` method will wipe the stores. According to this spec it shouldn't be doing that, regardless if the store is user-provided or internal.

> If CleanSession is set to 1, the Client and Server MUST discard any previous Session and start a new one. This Session lasts as long as the Network Connection.

The library does not clear the stores across automatic reconnections. The only case where it does it (except from the one mentioned above) is when the `end` method is called explicitly. According to this spec, the library should clear the stores even if it tries to reconnect automatically.

If these two things are fixed, then your statement is right: you can clear the callbacks upon disconnection simply based on the value of `clean`. Until then, we need to check `reconnectPeriod`
, or else you'll have a mismatch between the content of the stores an the callbacks.

mcollina:
> As you pointed, if the user is using the internal store, the reconnect method will wipe the stores. According to this spec it shouldn't be doing that, regardless if the store is user-provided or internal.

This was discussed in https://github.com/mqttjs/MQTT.js/pull/734#discussion_r155227295 and other comments in that PR. The problem is that `end()` will close the stores. So, new stores should be provided.
I'm at a loss on how you would change the current API to meet the spec.

> The library does not clear the stores across automatic reconnections. The only case where it does it (except from the one mentioned above) is when the end method is called explicitly. According to this spec, the library should clear the stores even if it tries to reconnect automatically.

Only if `clean: true`. If `clean: false`, the stores should be left as they are. Feel free to address this.

pierreca:
it looks like the Store object has a `clean` option too. Maybe we could apply the `clean` option passed to `connect` to the stores too when we initialize the client, and that way closing the store wouldn't wipe the inflight packet dictionnary. then modify reconnect to reuse existing stores if `clean === false`?

pierreca:
@mcollina @periclesroalves it looks like the fix for the stores/reconnect logic is a little larger than the initial issue. I believe @anthonyvercolano brought in the changes that @periclesroalves suggested - is it acceptable to merge this as a first step and then tackle the store clean/reconnect issue separately?

lib/client.js
```diff
@@ -749,6 +752,22 @@ MqttClient.prototype._cleanUp = function (forced, done) {
   }
 }
 
+/**
+ * _flush - Invoke (if it exists) all of the callbacks in an outgoing queue with an
+ *          error that we've disconnected
+ * @api private
+ */
+MqttClient.prototype._flush = function (queue) {
```
mcollina: There is no need to have this as an instance method, it can be a top-level function.

anthonyvercolano:
Made a top level function and squashed.

PR #805, opened 2018-04-09
Use protocol from server list if available.

### LINKED ISSUE #775: Servers option does not allow different protocol than brokerURL.
When connecting using the servers option the protocol can't be set. Need to add it as an option to the `IClientOptions` interface so that this works:
```
var client = connect(null, {
    username: 'Username',
    password: accessToken,
    reconnectPeriod: 1000 * 5,
    servers:[{
        host: getMQTTBrokerHostname(),
        port: 8883,
        protocol: 'wss'
    },{
        host: getMQTTBrokerHostname(),
        port: 1884,
        protocol: 'ws'
    }]
})
```

`opts.protocol = opts.servers[client._reconnectCount].protocol` Should just need to add this to line 129 of the `lib/connect/index.js` file. I can create a pull request if this proposal seems reasonable. My goal for this functionality is to use a secure server but fall back to insecure if need be.

wcatron:
Fixes #775 

- Note: Could in rare cases break poorly defined options.

## Works
```
                servers:[{
                    host: getMQTTBrokerHostname(),
                    port: 8883,
                    protocol: 'wss'
                },{
                    host: getMQTTBrokerHostname(),
                    port: 1884,
                    protocol: 'ws'
                }]
```
## Works
I imagine this is the most common use case.
```
                protocol: 'ws',
                servers:[{
                    host: getMQTTBrokerHostname(),
                    port: 1884
                }]
```

## Doesn't work
```
                protocol: 'ws',
                servers:[{
                    host: getMQTTBrokerHostname(),
                    port: 8883,
                    protocol: 'wss'
                },{
                    host: getMQTTBrokerHostname(),
                    port: 1884
                }]
```
This will not default the second attempt to the `ws`. The code will need to be refactor extensively in order to properly resolve protocols on a per attempt basis with a default. Alternatively the protocol for a server can be required instead of optional.

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -119,6 +119,10 @@ function connect (brokerUrl, opts) {
     throw new Error('Missing clientId for unclean clients')
   }
 
+  if (opts.protocol) {
+    opts.defaultProtocol = opts.protocol
+  }
+
   function wrapper (client) {
     if (opts.servers) {
       if (!client._reconnectCount || client._reconnectCount === opts.servers.length) {
@@ -127,7 +131,7 @@ function connect (brokerUrl, opts) {
 
       opts.host = opts.servers[client._reconnectCount].host
       opts.port = opts.servers[client._reconnectCount].port
-      opts.protocol = (!opts.servers[client._reconnectCount].protocol ? opts.protocol : opts.servers[client._reconnectCount].protocol)
+      opts.protocol = (!opts.servers[client._reconnectCount].protocol ? opts.defaultProtocol : opts.servers[client._reconnectCount].protocol)
       opts.hostname = opts.host
 
       client._reconnectCount++
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -127,6 +127,7 @@ function connect (brokerUrl, opts) {
 
       opts.host = opts.servers[client._reconnectCount].host
       opts.port = opts.servers[client._reconnectCount].port
+      opts.protocol = (!opts.servers[client._reconnectCount].protocol ? opts.protocol : opts.servers[client._reconnectCount].protocol)
       opts.hostname = opts.host
 
       client._reconnectCount++
--- a/test/client.js
+++ b/test/client.js
@@ -260,26 +260,28 @@ describe('MqttClient', function () {
       })
     })
 
-    it('should reconnect to multiple host-ports combination if servers is passed', function (done) {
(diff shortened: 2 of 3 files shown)


mcollina:

What "does not work" mean? Can you make sure it'd throw in that case?

wcatron:
By doesn't work I mean it doesn't use the default protocol.

Are you fine with _requiring_ protocol in the `IClientOptions` servers array objects? 

Change this:
```
servers?: Array<{
    host: string
    port: number
    protocol?: 'wss' | 'ws' | 'mqtt' | 'mqtts' | 'tcp' | 'ssl' | 'wx' | 'wxs'
  }>
```
to this:
```
servers?: Array<{
    host: string
    port: number
    protocol: 'wss' | 'ws' | 'mqtt' | 'mqtts' | 'tcp' | 'ssl' | 'wx' | 'wxs'
  }>
```

mcollina:
Which protocol would it use if it's not specified?

I'm not ok in doing a semver-major upgrade for this.

wcatron:
I don't want to either. So if it's not specified it will use whatever `opts.protocol` was set to previously. So in this
```
                protocol: 'ws',
                servers:[{
                    host: getMQTTBrokerHostname(),
                    port: 8883,
                    protocol: 'wss'
                },{
                    host: getMQTTBrokerHostname(),
                    port: 1884
                }]
```
 other example I believe it was using `wss`. If you never set it and you leave it optional. It will use the whatever you set `{ protocol: THIS, servers: [{...},{...}] }` to. So I guess it's backwards compatible because you didn't have ANY definition for protocol in the servers array object. So no one would currently be using the "not working" example above.

If I do 
```
if (opts.protocol) {
    opts.defaultProtocol = opts.protocol;
}
```
when it's parsing the initial protocol and then use `opts.defaultProtocol` when `opts.servers[client._reconnectCount].protocol` doesn't exist that should solve the problem.

mcollina:
Ok, can you add a unit test?

wcatron:
Sorry for the delay, finally got around to the unit tests. I modified the test responsible for testing port-host configurations for the `servers` option to include protocol and a default protocol. Everything should be backwards compatible and working now.

Do you want me to squash my commits or anything before issuing a pull request? Sorry I'm not very up to date on best practices.

wcatron:
I think this is ready to be merged! 👍

Issue #804, opened 2018-04-09
Don't rely on node built-ins


chrste90:
I'm trying to use mqtt.js in an angular v6 app.

The problem is that mqtt.js uses process which doesn't exist in a browser environment.
In older versions of angular the cli provided node-shims to get it working but as of v6 it doesn't do it anymore. See https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814 for more context.

It would be nice if mqtt.js wouldn't rely on node build-ins like process.

chrste90:
@mcollina Oh, you are right, there are some more problematic dependencies. If i start an angular app, i get this error:
```
Uncaught ReferenceError: process is not defined
    at Object../node_modules/process-nextick-args/index.js (index.js:3)
    at __webpack_require__ (bootstrap:74)
    at Object../node_modules/readable-stream/lib/_stream_readable.js (_stream_readable.js:26)
    at __webpack_require__ (bootstrap:74)
    at Object../node_modules/readable-stream/readable-browser.js (readable-browser.js:1)
    at __webpack_require__ (bootstrap:74)
    at Object../node_modules/mqtt/lib/store.js (store.js:8)
    at __webpack_require__ (bootstrap:74)
    at Object../node_modules/mqtt/lib/client.js (client.js:7)
    at __webpack_require__ (bootstrap:74)
```

So i think the main problematic dependency is readable-stream (Or their dependency process-nextick-args). Do you know if it is possible to replace readable-stream with another dependency or should i try to bring up the error on their site? But if i look at their repo: https://github.com/nodejs/readable-stream, it looks like it is a library especially for node.js, so i don't think it would be approriate to change the use of node built-ins there.

Also for this repo, i just searched for `process` in the code and found out that there are some uses of process.nextTick(). I'm not very familiar with this but i think it should be possible to replace `process.nextTick()` with `setTimeout()`? If so, i could create a PR changing the behaviour (or create a switch to either use nextTick on node oder setTimeout on browser).

The uses of process.env in the tests should do no harm if one would try to use this lib in angular.

But for bin/sub.js and bin/pub.js there are uses of process.exit(). Is this code used by consumers or only for examples in this repo (sorry, i'm not very familiar with mqtt.js yet :))

mcollina:
`setTimeout` has a completely different meaning than `process.nextTick`, and it would a _very slow_ replacement for Node.js.

[...]

PR #808, opened 2018-04-16
Added browser shim for process

sclausen:
As requested in #804, here is a browser shim for process, which is needed to use mqtt.js with angular-cli >= 6.0.0, since they removed the node-shims.

### DIFF
--- a/lib/process.js
+++ b/lib/process.js
@@ -1,0 +1,5 @@
+var process
+process.nextTick = setTimeout
+process.title = 'browser'
+
+module.exports = process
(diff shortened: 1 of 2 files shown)


RangerMauve:
Wouldn't it be better to have node-shims included in the build? Would this be defining the `process` global that's needed by the code?

RangerMauve:
Yeah, it's pretty finicky stuff. That's why I stick to browserify a lot these days.

RangerMauve:
Speaking of which, if you're in a rush and having trouble, you could turn mqtt.js into a browserified bundle and import that in your angular app.

sclausen:
Unfortunately, it's a library and I have no clue, how to embed a js-file in this bleeding edge angular-cli lib project.

PR #813, opened 2018-05-02
Fixed: weapp's real device environment can only send data of type String/ArrayBuffer

zousandian:
As weapp doc [mentioned](https://developers.weixin.qq.com/miniprogram/dev/api/network-socket.html#wxsendsocketmessageobject),
`wx.sendSocketMessage` can only send data of type String or ArrayBuffer.
The `msg` got here will be type of TypedArray, which can be sent successfully in wechat devtool (browser environment), but failed in real device (not real browser envirment). 
This PR fix it.


mcollina:
Thanks!

PR #814, opened 2018-05-03
Update: weapp add port number back when not 80 or 443

zousandian:
Directly remove weapp port number (#811) may be confused for people using a mqtt server listen on port like 1883 in development. 

Although they can only use port 80 or 443 in weapp, we should let weapp throw the error, rather than conceal it.


mcollina:
Thanks!

shanjunmei:
When I am using a port other than 80/443, it automatically adds the current port to the request path, which leads to an error. What should I do?

shanjunmei:
> When I am using a port other than 80/443, it automatically adds the current port to the request path, which leads to an error. What should I do?

image

PR #819, opened 2018-05-07
Use mqtt-packet typings.

saboya:
As discussed in https://github.com/mqttjs/mqtt-packet/pull/39, this PR uses `mqtt-packet`'s typings instead of defining them here.

To avoid breaking changes, I still exported all of the previously exported definitions. This might not be desirable, but I'll leave that up to you.

A version dependency change on `mqtt-packet` would be necessary as well.


mcollina:
Thanks, can you use this PR so that it uses the new mqtt-packet typings?

saboya:
@mcollina That ought to do it.

PR #827, opened 2018-05-30
Feature mqtt 5

scarry1992:
Hi. We've implemented the MQTT 5.0 support. Please have a look and merge if it's OK. We will also make similar pull-requests to mqtt-connection and mqtt-packet repos.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -417,9 +417,9 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
 
   if (this.options.protocolVersion === 5) {
     packet.properties = opts.properties
-    if (opts.properties &&
+    if (!this.options.properties || ((opts.properties && this.options.properties) &&
       ((opts.properties.topicAlias && this.options.properties.topicAliasMaximum && opts.properties.topicAlias > this.options.properties.topicAliasMaximum) ||
-        (!this.options.properties.topicAliasMaximum && opts.properties.topicAlias))) {
+        (!this.options.properties.topicAliasMaximum && opts.properties.topicAlias)))) {
       delete packet.properties.topicAlias
     }
   }
@@ -1028,14 +1028,16 @@ MqttClient.prototype._handleConnack = function (packet) {
   clearTimeout(this.connackTimer)
 
   if (packet.properties) {
-    if (packet.properties.topicAliasMaximum && this.options.properties.topicAliasMaximum) {
-      this.options.properties.topicAliasMaximum = packet.propperties.topicAliasMaximum
+    if (packet.properties.topicAliasMaximum) {
+      if (!this.options.properties) { this.options.properties = {} }
+      this.options.properties.topicAliasMaximum = packet.properties.topicAliasMaximum
     }
     if (packet.properties.serverKeepAlive && this.options.keepalive) {
       this.options.keepalive = packet.properties.serverKeepAlive
       this._shiftPingInterval()
     }
     if (packet.properties.maximumPacketSize) {
+      if (!this.options.properties) { this.options.properties = {} }
       this.options.properties.maximumPacketSize = packet.properties.maximumPacketSize
     }
   }
--- a/lib/client.js
+++ b/lib/client.js
@@ -34,7 +34,7 @@ function defaultId () {
 function sendPacket (client, packet, cb) {
   client.emit('packetsend', packet)
 
-  var result = mqttPacket.writeToStream(packet, client.stream)
+  var result = mqttPacket.writeToStream(packet, client.stream, client.options)
 
   if (!result && cb) {
     client.stream.once('drain', cb)
@@ -298,6 +298,17 @@ MqttClient.prototype._setupStream = function () {
   // Echo connection errors
   parser.on('error', this.emit.bind(this, 'error'))
 
+  // auth
+  if (this.options.properties) {
+    if (!this.options.properties.authenticationMethod && this.options.properties.authenticationData) {
+      this.emit('error', new Error('Packet has no Authentication Method'))
+    }
+    if (this.options.properties.authenticationMethod && this.options.authPacket && typeof this.options.authPacket === 'object') {
+      var authPacket = xtend({cmd: 'auth', reasonCode: 0}, this.options.authPacket)
+      sendPacket(this, authPacket)
+    }
+  }
+
   // many drain listeners are needed for qos 1 callbacks if the connection is intermittent
   this.stream.setMaxListeners(1000)
 
@@ -308,6 +319,12 @@ MqttClient.prototype._setupStream = function () {
 }
 
 MqttClient.prototype._handlePacket = function (packet, done) {
+  var options = this.options
+
+  if (options.protocolVersion === 5 && options.properties && options.properties.maximumPacketSize && options.properties.maximumPacketSize < packet.length) {
+    this.emit('error', new Error('exceeding packets size ' + packet.cmd))
+  }
+
   this.emit('packetreceive', packet)
 
   switch (packet.cmd) {
@@ -398,6 +415,15 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
     dup: opts.dup
   }
 
+  if (this.options.protocolVersion === 5) {
+    packet.properties = opts.properties
+    if (opts.properties &&
+      ((opts.properties.topicAlias && this.options.properties.topicAliasMaximum && opts.properties.topicAlias > this.options.properties.topicAliasMaximum) ||
+        (!this.options.properties.topicAliasMaximum && opts.properties.topicAlias))) {
+      delete packet.properties.topicAlias
+    }
+  }
+
   switch (opts.qos) {
     case 1:
     case 2:
@@ -427,7 +453,7 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
  * @api public
  * @example client.subscribe('topic');
  * @example client.subscribe('topic', {qos: 1});
- * @example client.subscribe({'topic': 0, 'topic2': 1}, console.log);
+ * @example client.subscribe({'topic': {qos: 0}, 'topic2': {qos: 1}}, console.log);
  * @example client.subscribe('topic', console.log);
  */
 MqttClient.prototype.subscribe = function () {
@@ -440,6 +466,7 @@ MqttClient.prototype.subscribe = function () {
   var opts = args.pop()
   var invalidTopic
   var that = this
+  var version = this.options.protocolVersion
 
   delete obj.resubscribe
 
@@ -462,31 +489,50 @@ MqttClient.prototype.subscribe = function () {
     return this
   }
 
-  var defaultOpts = { qos: 0 }
+  var defaultOpts = {
+    qos: 0
+  }
+  if (version === 5) {
+    defaultOpts.nl = false
+    defaultOpts.rap = false
+    defaultOpts.rh = 0
+  }
   opts = xtend(defaultOpts, opts)
 
   if (Array.isArray(obj)) {
     obj.forEach(function (topic) {
-      if (that._resubscribeTopics[topic] < opts.qos ||
-          !that._resubscribeTopics.hasOwnProperty(topic) ||
+      if (!that._resubscribeTopics.hasOwnProperty(topic) ||
+        that._resubscribeTopics[topic].qos < opts.qos ||
           resubscribe) {
-        subs.push({
+        var currentOpts = {
           topic: topic,
           qos: opts.qos
-        })
+        }
+        if (version === 5) {
+          currentOpts.nl = opts.nl
+          currentOpts.rap = opts.rap
+          currentOpts.rh = opts.rh
+        }
+        subs.push(currentOpts)
       }
     })
   } else {
     Object
       .keys(obj)
       .forEach(function (k) {
-        if (that._resubscribeTopics[k] < obj[k] ||
-            !that._resubscribeTopics.hasOwnProperty(k) ||
+        if (!that._resubscribeTopics.hasOwnProperty(k) ||
+          that._resubscribeTopics[k].qos < obj[k].qos ||
             resubscribe) {
-          subs.push({
+          var currentOpts = {
             topic: k,
-            qos: obj[k]
-          })
+            qos: obj[k].qos
+          }
+          if (version === 5) {
+            currentOpts.nl = obj[k].nl
+            currentOpts.rap = obj[k].rap
+            currentOpts.rh = obj[k].rh
+          }
+          subs.push(currentOpts)
         }
       })
   }
@@ -510,7 +556,12 @@ MqttClient.prototype.subscribe = function () {
     var topics = []
     subs.forEach(function (sub) {
       if (that.options.reconnectPeriod > 0) {
-        that._resubscribeTopics[sub.topic] = sub.qos
+        that._resubscribeTopics[sub.topic] = { qos: sub.qos }
+        if (version === 5) {
+          that._resubscribeTopics[sub.topic].nl = sub.nl || false
+          that._resubscribeTopics[sub.topic].rap = sub.rap || false
+          that._resubscribeTopics[sub.topic].rh = sub.rh || 0
+        }
         topics.push(sub.topic)
       }
     })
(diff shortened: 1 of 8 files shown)


mcollina:
Good work, I’ll wait to land this until mqtt-packet and mqtt-connection has been updated and release.

Do you think this is a semver-minor change?

scarry1992:
HI, thanks for approving my commit. (the main answer in mitt-packet)

scarry1992:
Hello, can you publish new version of mqtt-connect to npm? And we could update this package as well.

mcollina:
@scarry1992 can you please update this so that tests are passing? Please do not bump the version or merge this. Thanks.

scarry1992:
@mcollina yes, I can, but I needed in updated version of mqtt-connection. After updating mqtt-connection packet in npm, I will update MQTT.js and all test will be OK. I'm sorry for mqtt-connection merge.

mcollina:
It is now.

scarry1992:
@mcollina Do I need to write more tests?

mcollina:
I think so, yes. Look at the areas that are not covered in codecov.

lib/client.js
```diff
@@ -298,6 +298,17 @@ MqttClient.prototype._setupStream = function () {
   // Echo connection errors
   parser.on('error', this.emit.bind(this, 'error'))
 
+  // auth
+  if (this.options.properties) {
+    if (!this.options.properties.authenticationMethod && this.options.properties.authenticationData) {
+      this.emit('error', new Error('Packet has no Authentication Method'))
```
mcollina: we might want to return after this? Emitting `'error'`  is a distructive action.

lib/client.js
```diff
@@ -308,6 +319,12 @@ MqttClient.prototype._setupStream = function () {
 }
 
 MqttClient.prototype._handlePacket = function (packet, done) {
+  var options = this.options
+
+  if (options.protocolVersion === 5 && options.properties && options.properties.maximumPacketSize && options.properties.maximumPacketSize < packet.length) {
+    this.emit('error', new Error('exceeding packets size ' + packet.cmd))
```
mcollina: if we are emitting error, we should really destroy the client or fully disconnect (and reconnect).

lib/client.js
```diff
@@ -585,14 +653,65 @@ MqttClient.prototype.unsubscribe = function (topic, callback) {
...
+  //   154: 'Retain not supported',
+  //   155: 'QoS not supported',
+  //   156: 'Use another server',
+  //   157: 'Server moved',
+  //   158: 'Shared Subscriptions not supported',
+  //   159: 'Connection rate exceeded',
+  //   160: 'Maximum connect time',
+  //   161: 'Subscription Identifiers not supported',
+  //   162: 'Wildcard Subscriptions not supported'
+  // }
```
mcollina: why is this block commented here? If it's not needed, remove.

lib/client.js
```diff
@@ -585,14 +653,65 @@ MqttClient.prototype.unsubscribe = function (topic, callback) {
...
+
+  var args = Array.prototype.slice.call(arguments)
```
mcollina: Don't do this. It's super slow in older versions of Node.js (< 6), and not really needed in this case.

lib/client.js
```diff
@@ -874,18 +993,55 @@ MqttClient.prototype._handlePingresp = function () {
...
+    144: 'Topic Name invalid',
+    149: 'Packet too large',
+    151: 'Quota exceeded',
+    153: 'Payload format invalid',
+    154: 'Retain not supported',
+    155: 'QoS not supported',
+    156: 'Use another server',
+    157: 'Server moved',
+    159: 'Connection rate exceeded'
+  }
```
mcollina: this should really go to the top of the file. There is no need to declare this here.

lib/client.js
```diff
@@ -1000,28 +1174,43 @@ MqttClient.prototype._handleAck = function (packet) {
...
       // same thing as puback for QoS 2
     case 'puback':
+      var pubackRC = packet.responseCode
       // Callback - we're done
       delete this.outgoing[mid]
       this.outgoingStore.del(packet, cb)
+      if (pubackRC && pubackRC > 0 && pubackRC !== 16) {
+        err = new Error('Publish error: ' + errors[pubackRC])
+        err.code = pubackRC
+        this.emit('error', err)
```
mcollina: we should call the callback with the error, not emitting one.

lib/client.js
```diff
@@ -1000,28 +1174,43 @@ MqttClient.prototype._handleAck = function (packet) {
...
         qos: 2,
         messageId: mid
       }
+      var pubrecRC = packet.responseCode
 
-      this._sendPacket(response)
+      if (pubrecRC && pubrecRC > 0 && pubrecRC !== 16) {
+        err = new Error('Publish error: ' + errors[pubrecRC])
+        err.code = pubrecRC
+        this.emit('error', err)
```
mcollina: same here

test/server.js
```diff
@@ -8,8 +8,10 @@ var MqttServer
 var MqttSecureServer
 
 function setupConnection (duplex) {
-  var connection = new Connection(duplex)
-  this.emit('client', connection)
+  var that = this
+  var connection = new Connection(duplex, function () {
+    that.emit('client', connection)
+  })
```
mcollina: why this change?

lib/client.js
scarry1992: I`ll add returning 'this' after emitting error

lib/client.js
scarry1992: i`ll add end for connection with reason code, but don`t add reconnect, because I get loop, if do this.

lib/client.js
scarry1992: resolve

lib/client.js
scarry1992: I`ll remove it

lib/client.js
scarry1992: resolve

lib/client.js
scarry1992: rebase to call callback

lib/client.js
scarry1992: rebase to call callback

test/server.js
scarry1992: this changed because logic of initialization of server was been changed. Server process packets by version in connect packet

psorowka:
Hi @scarry1992 thanks for the great effort!

I have a question regarding your design choices. I understand that most of the publish property fields (like `contentType`) are application knowledge and thus must be specified by the outer application when calling `publish`.

However I believe the `topicAlias` feature falls into a different category as it basically is a compression feature that requires some special handling for managing and resolving the alias<->topic map. I think this burden should not be put onto the application but it would be convenient if mqttjs would manage the topic alias mapping auto-magically under the hood.

Currently - if I didn't miss anything in your implementation - I understand that the application must give the `topicAlias` as a parameter to `publish` and also in the `on('message')` handler the application must handle cases when `topic` *or* `topicAlias` *or* both is set and resolve aliases to topics accordingly. Is that correct?

psorowka:
I am also thinking about other convenience features that could reflect some MQTT5 ideas in the client API, e.g.:

- automatically set the `payloadFormatIndicator` flag depending on if the payload parameter is of type `String` or other
- provide additional events that represent specific subscriptions (subscription_id feature) <-- or is that a bad idea from a performance perspective, @mcollina ?

might be more.. but that's just for now.

scarry1992:
@psorowka I’ve got your point and it does make sense. However, I still think that this logic should be left at the discretion of the MQTT.js user. We can also ask @mcollina opinion and if he believes that this functionality should be implemented, I will do that.

@mccollina, MQTT 5 enables communication between the broker and the client using reason codes, which, in my understanding, means that there should be the opportunity to send puback/pubrel packets manually by a user. Did I get it right? More details here: [pubac specs](http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs02/mqtt-v5.0-cs02.html#_Toc514345396) in reason codes section (Implementation specific error).

Please also check my fixes based on the comments in the review.

psorowka:
@scarry1992 what are the arguments towards leaving the alias logic to the user?

My strong opinion is that such kind of communication logic is what a user expects to be handled by a well tested client library instead of heaving to implement that manually in each application.

psorowka:
@scarry1992 as for the PUBACK. As far as I understand, that does not mean that a PUBACK packet should be triggered manually, it's still always the direct response to a PUBLISH packet. As the usual way to receive messages is the event listener, I think it's not possible to set a response from the callback. The only way I can think of would be to offer the ability to add an "interceptor" function that's called before the message event is triggered and may set a reason code.

On the other hand I am really wondering if it's really the intention of the spec to have the client set other reason codes than 0x00 or 0x80.. what should the broker do with the responses...

mcollina:
I'm sorry for the long delay in reviewing this. I've been traveling for most of the last month, with more traveling next month.

@scarry1992 Good work!

> MQTT 5 enables communication between the broker and the client using reason codes, which, in my understanding, means that there should be the opportunity to send puback/pubrel packets manually by a user. Did I get it right? More details here: pubac specs in reason codes section (Implementation specific error).

I agree with you.

@psorowka I would like to keep this implementation minimal, so we can cut a new major and then improve on that one. We can even land it and start cutting -rc releases, and iron bugs along the way.

@scarry1992 @psorowka can you please point me to some docs about topic alias?

scarry1992:
http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs02/mqtt-v5.0-cs02.html#_Toc514345323 - topicAliasMaximum, http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs02/mqtt-v5.0-cs02.html#_Toc514345385 - topicAlias

scarry1992:
I believe that it is necessary to leave the topicAlias property on the user side because, at the moment, this makes the use of this flag not transparent, and can lead to incorrect behavior with different brokers. Most brokers currently do not support or support partially 5 standard and there may not be implemented this feature, and there is no need to completely stop working with this broker. I myself like the idea of bringing this logic to the library, but so far, it seems to me, you can not do it.

psorowka:
Vernemq is already almost completely supporting everyithing of mqtt5, just for testing.

IMHO topic aliases are one of the coolest features as it reduces the package size significantly. 

When releasing this to major we must be very careful that topicaliases are disabled by default (topicAliasMaximum 0) because otherwise a user might receive onMessage events with empty topic strings unexpectedly

scarry1992:
Hi. I did custom handling reason codes feature. But after merging master into my branch, I get broken test "should reconnect to multiple host-ports-protocol combinations if servers is passed". That test was been modified in last commit to master. I can`t understand whats going on. I get different error on thats test and in list buffer wrong binary data. But writer and parser workin as well. Maybe @mcollina can you review code one more time? Old test, before modifiing, was been passed.

scarry1992:
@mcollina Do I need to implement something else?

lib/client.js
```diff
@@ -93,6 +138,8 @@ function MqttClient (streamBuilder, options) {
 
   this.options.clientId = (typeof this.options.clientId === 'string') ? this.options.clientId : defaultId()
 
+  this.options.customHandleAcks = this.options.protocolVersion === 5 ? this.options.customHandleAcks : false
```
mcollina: it seems we are manipulating `this.options`  a lot in this function. Would you mind extracting it into a variable `var options = this.options`?

lib/client.js
```diff
@@ -398,6 +465,15 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
...
   }
 
+  if (this.options.protocolVersion === 5) {
+    packet.properties = opts.properties
+    if ((!this.options.properties && packet.properties && packet.properties.topicAlias) || ((opts.properties && this.options.properties) &&
+      ((opts.properties.topicAlias && this.options.properties.topicAliasMaximum && opts.properties.topicAlias > this.options.properties.topicAliasMaximum) ||
+        (!this.options.properties.topicAliasMaximum && opts.properties.topicAlias)))) {
+      delete packet.properties.topicAlias
+    }
+  }
```
mcollina: Can you please add a comment explaining this block?
Also, can you extract `var options = this.options`  for readability?

lib/client.js
```diff
@@ -427,7 +503,7 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
  * @api public
  * @example client.subscribe('topic');
  * @example client.subscribe('topic', {qos: 1});
- * @example client.subscribe({'topic': 0, 'topic2': 1}, console.log);
+ * @example client.subscribe({'topic': {qos: 0}, 'topic2': {qos: 1}}, console.log);
```
mcollina: Is this previous form now removed?

lib/client.js
```diff
@@ -510,7 +606,12 @@ MqttClient.prototype.subscribe = function () {
     var topics = []
     subs.forEach(function (sub) {
       if (that.options.reconnectPeriod > 0) {
-        that._resubscribeTopics[sub.topic] = sub.qos
+        that._resubscribeTopics[sub.topic] = { qos: sub.qos }
+        if (version === 5) {
+          that._resubscribeTopics[sub.topic].nl = sub.nl || false
+          that._resubscribeTopics[sub.topic].rap = sub.rap || false
+          that._resubscribeTopics[sub.topic].rh = sub.rh || 0
+        }
```
mcollina: we are manipulating a lot `_resubscribeTopics`. Can you please extract it into a separate variable?

Can you also add the docs for `nl` , `rap` and `rh`? If these are not exposed, can you please add a comment somewhere explaining them?

lib/client.js
```diff
@@ -537,21 +638,34 @@ MqttClient.prototype.subscribe = function () {
...
  */
-MqttClient.prototype.unsubscribe = function (topic, callback) {
+MqttClient.prototype.unsubscribe = function () {
   var packet = {
     cmd: 'unsubscribe',
     qos: 1,
     messageId: this._nextId()
   }
   var that = this
+  var args = Array.prototype.slice.call(arguments)
```
mcollina: This is better implemented with:

```js
var args = new Array(arguments.length)
for (var i = 0; i < arguments.length; i++) {
  args[i] = arguments[i]
}
```

because using `Array.prototype.slice` is extremely unperformant in older Node.js versions.

lib/client.js
```diff
@@ -874,18 +1013,26 @@ MqttClient.prototype._handlePingresp = function () {
...
+    }
+    if (packet.properties.serverKeepAlive && this.options.keepalive) {
+      this.options.keepalive = packet.properties.serverKeepAlive
+      this._shiftPingInterval()
+    }
+    if (packet.properties.maximumPacketSize) {
+      if (!this.options.properties) { this.options.properties = {} }
+      this.options.properties.maximumPacketSize = packet.properties.maximumPacketSize
+    }
+  }
```
mcollina: Can you extract `var options = this.options`?

lib/client.js
```diff
@@ -934,23 +1081,49 @@ MqttClient.prototype._handlePublish = function (packet, done) {
...
+          if (err) {
+            return done && done(err)
+          }
+          // send 'puback' if the above 'handleMessage' method executed
+          // successfully.
+          that._sendPacket({cmd: 'puback', messageId: mid}, done)
+        })
+      }
       break
+    }
```
mcollina: All this function becomes very complicated with the `customHandleHack` function. Can you please refactor so that there is _always_ a default for that, and we avoid all those ifs?

lib/client.js
```diff
@@ -934,23 +1081,49 @@ MqttClient.prototype._handlePublish = function (packet, done) {
...
-      this.handleMessage(packet, function (err) {
-        if (err) {
-          return done && done(err)
-        }
-        // send 'puback' if the above 'handleMessage' method executed
-        // successfully.
-        that._sendPacket({cmd: 'puback', messageId: mid}, done)
-      })
+      if (this.options.customHandleAcks) {
+        this.options.customHandleAcks(topic, message, packet, function (code) {
```
mcollina: the signature for the callback of `customHandleAcks` must be `function (err, code) {}`. We should also validate that `code` is valid, or is this done when generating a PUBACK`?

lib/client.js
scarry1992: From spec: If a Client receives a packet whose size exceeds this limit, this is a Protocol Error, the Client uses DISCONNECT with Reason Code 0x95 (Packet too large). If you think, that bad idea, I can remove emitting error, but I think thats right. Just tell me what do you think. I think user must proccess that error himself.

lib/client.js
scarry1992: fixed

lib/client.js
scarry1992: fixed

lib/client.js
scarry1992: right

lib/client.js
scarry1992: fixed, added info in readme

lib/client.js
scarry1992: implemented

lib/client.js
scarry1992: fixed

lib/client.js
scarry1992: refactored, check pls

lib/client.js
scarry1992: code validation added

scarry1992:
I fixed all issues and give answers. Can You check it and release new version?

scarry1992:
After update my branch, I has error in CI "The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 1 during ."

mcollina:
Do you know which brokers are currently compatible with this? I would like to do some manual testing.

My goto broker for this type of things is Mosquitto, and it's not supporting MQTT v5, but it should be ready sooner rather than later https://mosquitto.org/blog/2018/05/press-release/.

Regarding the CI, could you: a) drop testing on Node 4, 4.3.2 (lambda), 5, 7, and 9? Also, add Node 10.

mcollina:
Could you add a note in the README that MQTT v5 support is experimental as it has not been implemented by brokers yet?

scarry1992:
@mcollina yes sure

mcollina:
@scarry1992 can you please rebase your changes or merge master? It's conflicting now.

scarry1992:
@mcollina done)

Issue #810, opened 2018-04-26
messageId is not generated correctly, possible BUG


zlatinejc:
A strange event occurred yesterday on our MQTT server, calling for an invalid messageId. The value was 0. What, right ?

After some digging I found that the ```messageId``` property in [client.js](https://github.com/mqttjs/MQTT.js/blob/7ccda99f5245dc2d34d5903259267d28c007e979/lib/client.js#L123) has two problems:

1 a. 
The random value generated for messageId can be of value 0
```
// MessageIDs starting with 1
this.nextId = Math.floor(Math.random() * 65535)
```
1 b. 
If the value is actually 0 it is not incremented correctly in the [_nextId](https://github.com/mqttjs/MQTT.js/blob/7ccda99f5245dc2d34d5903259267d28c007e979/lib/client.js#L1074) method.
```
var id = this.nextId++
// id is still 0
```

2
Valid range for a messageId is  0 < messageID <= 65535 (max value of an unsigned integer 16 bit), but this value (65535) is never actually reached or used due to this code in [_nextId](https://github.com/mqttjs/MQTT.js/blob/7ccda99f5245dc2d34d5903259267d28c007e979/lib/client.js#L1077) method, therefore we loose an open slot.

```
if (id === 65535) {
    this.nextId = 1
}
```

RangerMauve:
Would you be able to submit a PR and unit tests that fix this?

Degola:
@RangerMauve please see following PR which tackle the mentioned issues:
https://github.com/mqttjs/MQTT.js/pull/841

1. Set minimum of this.nextId initialisation to 1:
```
this.nextId = Math.max(1, Math.floor(Math.random() * 65535))
```

2. Missing slot 65536 is not the case, the previous function was the following:
```
  var id = this.nextId++
  // Ensure 16 bit unsigned int:
  if (id === 65535) {
    this.nextId = 1
  }
  return id
```

Which results into the following behaviour:
1. nextId is 65535
2. id becomes 65535
3. nextId becomes 65536
4. if(id === 65535) sets nextId to 1 while nextId is already 65536

It would be different with var id = ++this.nextId.

Nevertheless, it's  hard to read and not very explicit, id is a copy of nextId while nextId is changed which is not really clear (devil is here definitively in the detail) therefore I added a merge request to change the implementation to with the purpose of easier readability:
```
var id = this.nextId++
if(this.nextId === 65536) {
  this.nextId = 1
}
return id
```

[...]

PR #841, opened 2018-07-16
Solves issue #810

Degola:
1: Set minimum of this.nextId initialisation to 1:
```
this.nextId = Math.max(1, Math.floor(Math.random() * 65535))
```

2. Missing slot 65536 is not the case, the previous function was the following:
```
  var id = this.nextId++
  // Ensure 16 bit unsigned int:
  if (id === 65535) {
    this.nextId = 1
  }
  return id
```

Which results into the following behaviour:
1. nextId is 65535
2. id becomes 65535
3. nextId becomes 65536
4. if(id === 65535) sets nextId to 1 while nextId is already 65536

It would be different with var id = ++this.nextId.

Nevertheless, it's  hard to read and not very explicit, id is a copy of nextId while nextId is changed which is not really clear (devil is here definitively in the detail) therefore I added a merge request to change the implementation to with the purpose of easier readability:
```
var id = this.nextId++
if(this.nextId === 65536) {
  this.nextId = 1
}
return id
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -120,8 +120,11 @@ function MqttClient (streamBuilder, options) {
   this.connackTimer = null
   // Reconnect timer
   this.reconnectTimer = null
-  // MessageIDs starting with 1
-  this.nextId = Math.floor(Math.random() * 65535)
+  /**
+   * MessageIDs starting with 1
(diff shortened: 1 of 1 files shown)


RangerMauve:
This looks good to me! I'll wait for someone else to sign off so we can release a new version. Thank you!

PR #846, opened 2018-07-21
fix: edge case when handlePacket is called after .end()

mvayngrib:


mcollina:
Can you please add a unit test for this case?

mvayngrib:
@mcollina i've updated the PR with a unit test 

however, i noticed that other functionality relies on `_handlePacket` being called while `disconnecting` is true, e.g. this test: "should delay ending up until all inflight messages are delivered"

the original reason for this fix is that i was getting a `puback` when `outgoingStore` was already set to null. How would you prefer to handle this?

mcollina:
I'm a bit lost.
I'd like to retain the current functionality. We might want to delay clearing up the stores until all messages are delivered, if I understand your goal correctly.

mvayngrib:
haha, i'm sure i'm more lost. Basically, in some edge cases, `outgoingStore` is null and `outgoingStore.del` errors

mcollina:
Can you please add a unit test that verifies one of those edge cases?

mvayngrib:
@mcollina not actually sure what causes it, but i'll look out for a repro. Just fyi, i'm using `mqtt@1.11.2` by way of `aws-iot-device-sdk`

robertsLando:

robertsLando:
@mvayngrib could I know if you closed this because it wasn't fixing the issue or because you are not interested in continue with the PR?

mvayngrib:
@robertsLando i won't have time to pursue this in the foreseeable future 😅. It also doesn't seem like this is an issue for anyone else, so maybe it's fine as is

robertsLando:
I prefer to keep this open as it could be a  fix to some edge case issues

robertsLando:
Supersided by #1780

PR #865, opened 2018-08-31
Add cb in removeOutgoingMessage function arg

codincafe:
cb was missing in the removeOutgoingMessage which threw the below:

cb(new Error('Message removed'))
    ^

TypeError: cb is not a function


lib/client.js
```diff
@@ -667,7 +667,7 @@ MqttClient.prototype.end = function (force, cb) {
 MqttClient.prototype.removeOutgoingMessage = function (mid) {
```
mcollina: I think the callback should be defined here.

PR #416, opened 2016-06-03
Gracefully handle unknown message id's in handlePubrel

shylesh107:
This should handle PUBREL messages without throwing an exception for unknown message id's. 
See #415 for more details.

PR #893, opened 2018-11-25
fix: delete completed incoming QOS 2 messages

joeblynch:
I ran into an issue where my store was being filled with incoming QOS 2 messages, even though those messages had completed the QOS 2 handshake. If I understand the last point of `[MQTT-4.3.3-2]` in the MQTT spec correctly, the client should forget incoming QOS 2 messages once the `PUBCOMP` message has been sent.

@mcollina I noticed you also saw this issue when [you commented that this put should be a delete](https://github.com/mqttjs/MQTT.js/pull/416#discussion_r65729673). Do you know of any issues doing so could cause?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1215,18 +1215,14 @@ MqttClient.prototype._handlePubrel = function (packet, callback) {
   var comp = {cmd: 'pubcomp', messageId: mid}
 
   that.incomingStore.get(packet, function (err, pub) {
-    if (!err && pub.cmd !== 'pubrel') {
+    if (!err) {
       that.emit('message', pub.topic, pub.payload, pub)
       that.handleMessage(pub, function (err) {
         if (err) {
           return callback(err)
         }
-        that._sendPacket(comp, function (err) {
-          if (err) {
-            return callback(err)
-          }
-          that.incomingStore.del(packet, callback)
-        })
+        that.incomingStore.del(pub, nop)
+        that._sendPacket(comp, callback)
       })
     } else {
       that._sendPacket(comp, callback)
--- a/lib/client.js
+++ b/lib/client.js
@@ -1217,15 +1217,15 @@ MqttClient.prototype._handlePubrel = function (packet, callback) {
   that.incomingStore.get(packet, function (err, pub) {
     if (!err && pub.cmd !== 'pubrel') {
       that.emit('message', pub.topic, pub.payload, pub)
-      that.incomingStore.del(packet, function (err) {
+      that.handleMessage(pub, function (err) {
         if (err) {
           return callback(err)
         }
-        that.handleMessage(pub, function (err) {
+        that._sendPacket(comp, function (err) {
           if (err) {
             return callback(err)
           }
-          that._sendPacket(comp, callback)
+          that.incomingStore.del(packet, callback)
         })
       })
     } else {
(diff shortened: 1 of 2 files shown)


joeblynch:
Closing because I realized this implementation will result in the message being handled by the client more than once if the `PUBCOMP` fails to send and multiple `PUBREL` are received. Will reopen once I have this resolved and tested.

lib/client.js
```diff
@@ -1215,18 +1215,14 @@ MqttClient.prototype._handlePubrel = function (packet, callback) {
...
-          if (err) {
-            return callback(err)
-          }
-          that._sendPacket(comp, callback)
-        })
+        that.incomingStore.del(pub, nop)
+        that._sendPacket(comp, callback)
       })
     } else {
       that._sendPacket(comp, callback)
```
mcollina: I think we might want to change this error condition. Should we really send pubcomp here?

mcollina:
Can you add a test or some assertions that checks if nothing is left in incomingStore?

lib/client.js
joeblynch: My thought here is that if `del` were to fail, and prevented sending `pubcomp`, the server would respond with another `pubrel`, which wold trigger the QOS 2 message to be handled twice.

The prior implementation prevented this by saving the `pubrel` to the incoming store, and checking for that message in the store when a second `pubrel` is received. However, it also had the effect of leaving the messages in the incoming store, and would result in the message not being reprocessed if the first call to `handleMessage` failed.

I could definitely be misunderstanding the spec here though, so I'd appreciate any input on how to better handle this, if this is incorrect.

mcollina: The guarantee about being processed only once is important for QoS 2.

I think we might want to keep the `put`, but rather call a `del`  after `_sendPacket(comp)` callback.

joeblynch: If I understand correctly, I think that'd look something like this:

```JavaScript
  that.incomingStore.get(packet, function (err, pub) {
    if (!err && pub.cmd !== 'pubrel') {
      that.emit('message', pub.topic, pub.payload, pub)
      that.incomingStore.put(packet, function (err) {
        if (err) {
          return callback(err)
        }
        that.handleMessage(pub, function (err) {
          if (err) {
            return callback(err)
          }
          that._sendPacket(comp, function (err) {
            if (err) {
              return callback(err)
            }
            that.incomingStore.del(packet, callback)
          })
        })
      })
    } else {
      that._sendPacket(comp, callback)
    }
```

The thing I'm unclear on here, should the message be considered processed if the call to `handleMessage` returns an error? With this logic, if `handleMessage` fails and prevents `pubcomp` from being sent, when the server resends `pubrel` the check for the previous `pubrel` message will prevent `handleMessage` from being called again.

Another concern is that if `_sendPacket` were to fail, the message wouldn't be removed from the incoming store when the server resent `pubrel`, although that could probably be resolved by splitting out the check for the `pubrel` and duplicating the send/del logic:

```JavaScript
that.incomingStore.get(packet, function (err, pub) {
    if (!err) {
      if (pub.cmd !== 'pubrel') {
        // ...
      } else {
        that._sendPacket(comp, function (err) {
          if (err) {
            return callback(err)
          }
          that.incomingStore.del(packet, callback)
        })
      }
    } else {
      that._sendPacket(comp, callback)
    }

mcollina: > The thing I'm unclear on here, should the message be considered processed if the call to handleMessage returns an error? With this logic, if handleMessage fails and prevents pubcomp from being sent, when the server resends pubrel the check for the previous pubrel message will prevent handleMessage from being called again.

I think so, yes. When `handleMessage` is called, the responsibility passes down to the application.

> Another concern is that if _sendPacket were to fail, the message wouldn't be removed from the incoming store when the server resent pubrel, although that could probably be resolved by splitting out the check for the pubrel and duplicating the send/del logic

+1 for splitting the logic

Can you add tests for those cases?

Issue #892, opened 2018-11-22
The implementation of Wechat Mini Program is out of data


SyMind:
The implementation of Wechat Mini Program is out of data. In WeChat Mini Program document:

https://developers.weixin.qq.com/miniprogram/en/dev/api/network-socket.html#wxconnectsocketobject

> Before base library 1.7.0, one Mini Program can only have one "WebSocket" link at one time, if a "WebSocket" link already exists, that link will be automatically closed and a "WebSocket" link will be re-created. For base library 1.7.0 and later versions, multiple "WebSokcet" links are supported, a new SocketTask will be returned whenever "wx.connectSocket" is called successfully.

SyMind:
I read code in [](https://github.com/mqttjs/MQTT.js/blob/master/test/browser/wx.js), but i have not find ../helpers/wx.

taoqf:
Sorry i didn't know this.

SyMind:
emmm, you add '../helpers/wx'. How to write unit test for Weixin.

taoqf:
Actually, you could not, there is no environment in nodejs.

taoqf:
Forget about unit tests.

SyMind:
think you. (゜-゜)つロ

PR #898, opened 2018-12-06
fix bug in weixin min program and add support to ali min program

SyMind:
issue #892


mcollina:
Can you please restore the test, and add a similar one for ali?

SyMind:
@mcollina I don't think i can do it. Because weixin and ali min program is a platform that Tencent and Alibaba does not open source, and they donot have tools for unit testing. 

I read code in https://github.com/mqttjs/MQTT.js/blob/master/test/browser/wx.js, but i have not find ../helpers/wx. I wonder guy who add weixin min program support how to do it.

lib/connect/ws.js
```diff
@@ -82,7 +82,9 @@ function buildBuilderBrowser (client, opts) {
       opts.port = parsed.port
     }
   }
-  return createWebSocket(client, opts)
+  var ws = createWebSocket(client, opts)
+  console.log('ws -> ', ws)
```
mcollina: Can you remove this console.log?

SyMind:
@taoqf i need your help.

lib/connect/index.js
```diff
@@ -87,6 +90,7 @@ function connect (brokerUrl, opts) {
           case 'wx':
             opts.protocol = 'wxs'
             break
+          case 'alis':
```
mcollina: I fear this might throw for alis.

SyMind: case 'ali':
            opts.protocol = 'alis'
            break

lib/connect/ws.js
SyMind: yes

taoqf:
@SyMind I might use that, too. I am glad to see this pr. 
@mcollina , Actually I don't think anyone could give a unit test for ali or wx, it could only run in it's [devetools](https://developers.weixin.qq.com/miniprogram/dev/devtools/devtools.html) and in [weixin](https://weixin.qq.com/) or [alipay](https://www.alipay.com/).

mcollina:
@SyMind is this ready to go?

@taoqf can you review as well?

PR #913, opened 2019-01-25
fix bug in weapp

SyMind:
there are some bug in weapp


SyMind:
@RangerMauve 
Last code for weapp is also pulled request by me, is fixed some bug for it.

RangerMauve:
@SyMind Fair enough. :)

@mcollina Do you think we could include this in the next release? I don't think I have publish rights for MQTT.js

mcollina:
I'll see if I can get a new release out asap.

PR #917, opened 2019-02-14
MQTT 5.0 resubscribe behavior fix

scarry1992:


scarry1992:
@mcollina, something go wrong with 10 version of node with test, but on my PC all test is passed on last 10 version. Maybe you can rerun tests for 10 node?

PR #926, opened 2019-03-18
server side disconnect handling

scarry1992:
[Disconnect MQTT 5.0](https://docs.oasis-open.org/mqtt/mqtt/v5.0/cos02/mqtt-v5.0-cos02.html#_Toc1477505). Can be sent from server side.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -351,6 +351,10 @@ MqttClient.prototype._handlePacket = function (packet, done) {
       this._handlePingresp(packet)
       done()
       break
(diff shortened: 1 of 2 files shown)


scarry1992:
Hm, what wrong? I added test.

scarry1992:
@mcollina, do you know what wrong? codcov speaks all new code strings covered by tests..I must write some tests more?

mcollina:
I don't know, and I would not worry.

PR #929, opened 2019-03-21
echo websocket stream error

SyMind:


lib/client.js
```diff
@@ -279,8 +279,10 @@ MqttClient.prototype._setupStream = function () {
 
   this.stream.pipe(writable)
 
-  // Suppress connection errors
-  this.stream.on('error', nop)
+  // Echo connection errors
+  this.stream.on('error', function (error) {
+    that.emit('error', error)
+  })
```
mcollina: Ignoring those errors is done on purpose. I’d be ok in emitting them as “streamError” instead.

lib/connect/wx.js
```diff
@@ -104,6 +104,27 @@ function buildStream (client, opts) {
 
   proxy = buildProxy()
   stream = duplexify.obj()
+  stream._destroy = function (err, cb) {
+    socketTask.close({
+      success: function () {
+        cb && cb(err)
+      }
```
mcollina: You might want to call the callback in case or error as well.

lib/connect/wx.js
```diff
@@ -104,6 +104,27 @@ function buildStream (client, opts) {
...
+
+    var self = this
+    process.nextTick(function () {
+      socketTask.close({
+        fail: function () {
+          self._destroy(new Error())
+        }
+      })
+    })
+  }.bind(stream)
```
mcollina: I don’t think this method should be overridden.

lib/client.js
SyMind: When will it be updated code?

lib/connect/wx.js
SyMind: sorry, it cause by donot merge code from master.

Issue #791, opened 2018-03-09
Memory leak in mqtt client


stephendeyoung:
I think I've discovered a memory leak in mqtt.js. Here is an example script:

https://gist.github.com/stephendeyoung/3180ce2ec527332e0ece6eee573e6ecb

Steps to reproduce:

1. Run script with 100 mqtt clients and using the node.js inspector: `NUMBER_OF_CLIENTS=100 node --inspect mqtt-memory-leak.js`
2. Wait for all mqtt clients to connect
3. Take a heap snapshot with the Chrome devtools
4. Wait 5-10 minutes
5. Take another heap snapshot
6. Wait another 5-10 minutes
7. Take a final heap snapshot

You should see from the snapshots that memory use is growing over time. Here is an example:

https://ufile.io/bdmme

You can even see this for just 1 mqtt client although you have to wait quite a long time before creating snapshots to see the effect.

For my particular use case I wanted to create over 1000 mqtt clients in a node.js script but I kept on running out of memory and it seems like this memory leak is causing the problem.

I am using version 2.16.0 of mqtt.js. Please let me know if I've made a mistake anywhere.

mcollina:
@stephendeyoung unfortunately I will have no time to investigate this issue for the next few weeks. Feel free to send a pull request if you figure it out.

aaamitsingh:
Any update on this issue?

RangerMauve:
Nothing so far, I don't think anyone has looked into it in detail. If you have time, you could help by running a node process and using the new [inspector](https://nodejs.org/en/docs/guides/debugging-getting-started/) feature to take memory snapshots to find where the leak is happening.

stephendeyoung:
@senthil5053 no i didn't unfortunately. Sorry I can't help.

svogl:
I ran into the same issue and had a look around. I see memory use increasing in a minimal example that just receives (frequent) messages in an empty message handler. looking further, I suspect calling the work function with process.nextTick in lib/client.js 

https://github.com/mqttjs/MQTT.js/blob/0379bf53a8ec803b0ac3ccbe94e22467abe57213/lib/client.js#L259

directly calling work() instead results in a much flatter memory consumption line. I don't know which side-effects this would cause when you're reconnecting with many messages in the queues though.

Any ideas on this @stephendeyoung ?

Simon

RangerMauve:
@svogl Which version of node are you using to test this?

svogl:
[...]

PR #931, opened 2019-03-25
perform nextTick work only if work needs to be done

svogl:
This addresses the memory leak discussed in issue #791 
Alongside, it is more efficient as timer objects are created only if it's necessary.


svogl:
great - looking forward to the next minor release on npm 😄

PR #937, opened 2019-04-09
process disconnect packet w/o full destroy the connection

scarry1992:
I don`t see any information about reconnections logic in spetification and I think that behavior more logical in this case. @mcollina, what so you think?
[Disconnect MQTT 5.0](https://docs.oasis-open.org/mqtt/mqtt/v5.0/cos02/mqtt-v5.0-cos02.html#_Toc1477505).

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,12 @@ Emitted when a reconnect starts.
 
 Emitted after a disconnection.
 
+#### Event `'disconnect'`
+
+`function (packet) {}`
+
+Emitted after reciving disconnect packet from brocker. MQTT 5.0 feature.
+
 #### Event `'offline'`
 
 `function () {}`
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ Emitted after a disconnection.
 
 `function (packet) {}`
 
-Emitted after reciving disconnect packet from brocker. MQTT 5.0 feature.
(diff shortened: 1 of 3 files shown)


scarry1992:
@mcollina, do you see it?)

mcollina:
Unfortunately I don’t have time to debug this at the moment.

scarry1992:
@mcollina, what do you think about this PR?

README.md
```diff
@@ -291,6 +291,12 @@ Emitted when a reconnect starts.
 
 Emitted after a disconnection.
 
+#### Event `'disconnect'`
+
+`function (packet) {}`
+
+Emitted after reciving disconnect packet from brocker. MQTT 5.0 feature.
```
mcollina:
```suggestion
Emitted after receiving disconnect packet from broker. MQTT 5.0 feature.
```

scarry1992: @mcollina, Disconnect packet from broker is MQTT 5.0 feature. I removed this.end(true) for working well logic of reconnects. In another way we must reconnect by code.

scarry1992:
@mcollina, disconnect packet from broker is mqtt5.0 feature. So I think we must try to reconnect to broker by usual logic after close connection by broker. Isn`t it?

PR #946, opened 2019-05-03
resubscribe mqtt5 fix

scarry1992:
I fixed resubscribes after disconnect. MQTT 5 has packet specific settings(properties) and resubscribe must proccess with it settings.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1308,19 +1308,18 @@ MqttClient.prototype.getLastMessageId = function () {
  * @api private
  */
 MqttClient.prototype._resubscribe = function (connack) {
-  var self = this
   var _resubscribeTopicsKeys = Object.keys(this._resubscribeTopics)
   if (!this._firstConnection &&
       (this.options.clean || (this.options.protocolVersion === 5 && !connack.sessionPresent)) &&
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1305,12 +1308,23 @@ MqttClient.prototype.getLastMessageId = function () {
...
+  var _resubscribeTopicsKeys = Object.keys(this._resubscribeTopics)
   if (!this._firstConnection &&
       (this.options.clean || (this.options.protocolVersion === 5 && !connack.sessionPresent)) &&
-      Object.keys(this._resubscribeTopics).length > 0) {
+      _resubscribeTopicsKeys.length > 0) {
     if (this.options.resubscribe) {
-      this._resubscribeTopics.resubscribe = true
-      this.subscribe(this._resubscribeTopics)
+      if (this.options.protocolVersion === 5) {
+        _resubscribeTopicsKeys.forEach(function (topicKey) {
```
mcollina: Would you mind using a for loop instead?

scarry1992: Ok, I`ll rewrite to for loop later today.

scarry1992: rebased

PR #954, opened 2019-05-27
fix: set default servername in tls connect

darahayes:
Hey there,

I figured the best way to start this discussion was with a PR.

I have a scenario where I'm trying to talk to an MQTT broker that's behind a proxy. This proxy is performing SSL/TLS passthrough. Having read into [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) I discovered the `servername` option in the options for [tls.connect](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback).

I found that the only way to connect was by explicitly setting `servername` to the same value as `host` (the hostname for the proxy).

I'm wondering does it make any sense to set that as a default for convenience? Feel free to reject in any case 🙂

### DIFF
--- a/lib/connect/tls.js
+++ b/lib/connect/tls.js
@@ -5,6 +5,7 @@ function buildBuilder (mqttClient, opts) {
   var connection
   opts.port = opts.port || 8883
   opts.host = opts.hostname || opts.host || 'localhost'
+  opts.servername = opts.servername || opts.host
 
   opts.rejectUnauthorized = opts.rejectUnauthorized !== false
 
--- a/test/secure_client.js
+++ b/test/secure_client.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var mqtt = require('..')
+var should = require('should')
 var path = require('path')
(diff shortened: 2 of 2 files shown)


mcollina:

darahayes:
@mcollina I've pushed two unit tests. 

1. Test that the client will request the hostname as the servername when no servername is passed to the client
2. Test that this behaviour is overridden if a servername prop is explicitly passed to the client

Let me know if that is okay. Thanks!

mcollina:
Unfortunately this fails in Node v12. I'll revert.

```
(node:38405) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.
      1) should request the hostname as servername by default
      2) should request the servername
```

darahayes:
My bad. The failure here is caused by the test explicitly setting an IP address but I guess it still doesn't make sense to introduce the proposed change because Node 12 users who pass a direct IP address (I guess that is fairly common) would experience this problem also. Thanks for the quick review!

PR #958, opened 2019-06-06
[FIXED] Unsubscribe while topics are in array.

sikkapat79:
In previous commit. The checking statement was checked by using "typeof". It should check by using ```javascript
Arrar.isArray(topic)
```
for check array of topics.

### CODE REVIEW

**Comment by sikkapat79** [comment]
Sorry for about my auto formatting.
But the changing is in line number 685.

**Comment by OmgImAlexis** [comment]
Please change the formatting back and fix the conflicts.

**Comment by mcollina** [comment]

**Comment by OmgImAlexis** [comment]
So this is the whole PR without the whitespace changes. Actually looks reasonable.

https://github.com/mqttjs/MQTT.js/pull/958/files?diff=split&w=1

File: lib/client.js [hunk 1]
```diff
@@ -653,57 +653,57 @@ MqttClient.prototype.subscribe = function () {
...
+		callback = nop
+	}
+
+	if (this._checkDisconnecting(callback)) {
+		return this
+	}
+
+	if (typeof topic === 'string') {
+		packet.unsubscriptions = [topic]
+	} else if (Array.isArray(topic)) {
```
sikkapat79, line 685: Checking with primitive function instead of checking by typeof variable

OmgImAlexis:
LGTM

PR #967, opened 2019-06-26
Retained message config example

MBerka:
Initially, ran into a TypeScript compilation problem where `{retain: true}` would not compile (and provided an unhelpful "does not match the callback type" message) because it was missing the `qos` property and therefore not recognized as the `IClientPublishOptions` type. Wanted to add documentation for this niche, then found that the basic examples did not say much about defining remote broker paths via URL string. Add the responsibility of clearing the retained message on the test broker, and this is is the result.

Happy to adjust for clarity, or write about the TypeScript pitfall elsewhere if there is a better spot. Thanks!

### DIFF
--- a/examples/client/both-with-retain.js
+++ b/examples/client/both-with-retain.js
@@ -1,0 +1,36 @@
+'use strict'
+
+var mqtt = require('../..')
+
+// Subscribe to a custom server with a URL string
+var client = mqtt.connect('mqtt://test.mosquitto.org:1883')
+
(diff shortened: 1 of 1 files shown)


robertsLando:

MBerka:
Not worth it for something that no one looked at for 3 years.

robertsLando:
@MBerka I started maintaining this PR so now I will check all issues/PR :)

PR #969, opened 2019-07-09
The protocols parameter of wx.connectSocket should be Array.

jerecui:
Wechat mini app document url:
https://developers.weixin.qq.com/miniprogram/dev/api/network/websocket/wx.connectSocket.html.

If the protocols value is 'mqtt' instead of ['mqtt'], it will be failed if
you use the android device, the interesting thing is iOS works.

Both android and ios works if the value is an array.


OmgImAlexis:
@mcollina any chance this can be merged?

Issue #987, opened 2019-09-10
Weapp fails on real devices.


dtweedle:
As the title says, right now we are unable to get the client to connect to AWS IoT when previewing using a real device, things work correctly inside of the simulator.

I've carried out a few small tests and I believe that the issue is that the protocols are not being correctly set on the socket connection. [If you look at the documentation](https://developers.weixin.qq.com/miniprogram/en/dev/api/network/websocket/wx.connectSocket.html), the protocol property that you provide to `wx.connectSocket` should be a string array. However in the[ source code ](https://github.com/mqttjs/MQTT.js/blob/master/lib/connect/wx.js#L92)it is simply provided as a string.

I've tested this by manually creating a connection manually using the following:
* No protocol
* Protocol as string
* Protocol as string array

The results seem to backup my hypothesis, when using no protocol both simulator and preview fail to connect, when using a string, the simulator connects correctly but the real device fails, and when using a string array, both the simulator and the real device work correctly.

I noticed the pull request fixxing this issue. This should be merged in, there seems to be a bunch of people having trouble with this and it's a difficult problem to diagnose.

PR #988, opened 2019-09-10
Add protocol string to array

dtweedle:
Fixes issue #987 

The protocol is now added to an array inside the `wx.connectSocket` function call.


dtweedle:
Duplicate

PR #989, opened 2019-09-11
fix: re-subscription of topics after connection loss

rlaranjeira:


ievgennaida:
Hello @rlaranjeira  thanks for the fix.

I am having similar problems that subscriptions are lost when the connection is unstable. 
One of the issues reproduced was that subscriptions are lost when QoS2 is used. 

But this fix is not solving this and even more, messages are received two times when I am testing it with the QoS 2 

I have tried to test this forward the problem I am having: https://github.com/mqttjs/MQTT.js/issues/1001
But issues are still there and duplicate message is recieved:

```Broker: Ready
Broker: client connected: client
Client: connect
Client: message recieved:data. message when connected. qos:2
Broker: Closed
Client: reconnect
Broker: Ready
Client: reconnect
Broker: client connected: client
**Client: message recieved:data. message when connected. qos:2**
```

Could you please describe the problem you are having better? 
How have you reproduced this? 

Thanks in advance!

robertsLando:

PR #998, opened 2019-10-16
docs: describe object-form subscription options usage

joshperry:
It's non-obvious that `options` is not used when using an object to
subscribe. Seems to just be doc rot from when there was only 1 option.


robertsLando:

PR #1003, opened 2019-10-25
Disabled auto reconnect for wrong credentials

diegovasquesmellet:
This PR is basically to disable the auto reconnect if the user typed wrong credentials (errors 4 / 5)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -178,6 +178,10 @@ function MqttClient (streamBuilder, options) {
   this.connackTimer = null
   // Reconnect timer
   this.reconnectTimer = null
+  // reconnect possible
+  // true := try reconnect
+  // false := do NOT reconnect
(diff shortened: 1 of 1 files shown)


diegovasquesmellet:
Hey..
I think it would look good.

Will work on it.

diegovasquesmellet:
We are sending the error like this: `var err = new Error('Connection refused: ' + errors[rc])`
Do you want to create an specific one for wrong credentials? They way it is we are already sending a connection refused + error.

atrovato:
Hello,
on my side, when MQTT could not connect, it retries... but I never get any error.

diegovasquesmellet:
This is funny. It was supposed to emit a connection refused error.

quaddan:
Is there any news?

PR #1038, opened 2020-01-30
fix: remove close listener before next startStreamProcess loop

patrykwegrzyn:
This cause memory leak during fast publishing when leveldb store is used
```VM38 warning.js:27 (node:2090114) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [MqttClient]. Use emitter.setMaxListeners() to increase limit```

### CODE REVIEW

**Comment by rtuin** [comment]
I can't tell much about if this is implemented correctly, but I know it solves the same problem for me.

**Comment by YoDaMa** [comment]
@rtuin can you explain the problem in a bit more detail? There is a memory leak when publishing in general or just when using leveldb store?

**Comment by rtuin** [comment]
@YoDaMa I ran into this issue while writing a simple load generator to run some tests.

* Notice I'm using `async-mqtt`
* The MQTT broker software is VerneMQ
* The Promise.allConcurrent is from: https://gist.github.com/jcouyang/632709f30e12a7879a73e9e132c0d56b

Let me elaborate with a code sample. Hope this helps!

```js
var mqtt = require('async-mqtt');
const fs = require('fs');

const runAsync = async () => {
    var client = await mqtt.connect('mqtts://my-host.domain', {
        key: fs.readFileSync('client.key'),
        cert: fs.readFileSync('client.crt'),
        ca: [ fs.readFileSync('ca.crt') ]
    });
     
    let count = 0;
    const publishes = [];
    while (count++ < 100000) {
        publishes.push(() => {
            return client.publish('my-topic', 'Hello mqtt' + count, {qos: 2});
        });
    }

    await Promise.allConcurrent(1000)(publishes).then((x) => {
        console.log(x);
    });

    return await client.end();
};

runAsync();
```

robertsLando:
@patrykwegrzyn Can you sync with master so I can merge this?

robertsLando:

robertsLando:
Supersided by #1759

PR #1042, opened 2020-02-10
Websocket stream removal

YoDaMa:


lib/connect/ws.js
```diff
@@ -1,6 +1,9 @@
 'use strict'
 
-var websocket = require('websocket-stream')
+import * as dbg from 'debug'
```
mcollina: This should not have landed.

YoDaMa: I forced a commit to the branch, hence why it landed. This PR was not into Master though. Just a carry-over PR into another branch...

mcollina: don't use import :).

PR #1043, opened 2020-02-10
fix various options definition for ts usage

vincent-tr:
Looking at mqtt-packet, 'will' payload should be a buffer

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -80,7 +80,7 @@ export interface IClientOptions extends ISecureClientOptions {
     /**
      * the message to publish
      */
-    payload: Buffer
+    payload: Buffer | string
     /**
      * the QoS
      */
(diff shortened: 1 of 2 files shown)


types/lib/client-options.d.ts
```diff
@@ -80,7 +80,7 @@ export interface IClientOptions extends ISecureClientOptions {
     /**
      * the message to publish
      */
-    payload: string
+    payload: Buffer
```
mcollina: I think it can be both a string or a buffer

vincent-tr:
I've also marked 'qos' as optional on client.publish()

vincent-tr:
Sure, I did not see we have tests on typescript definitions, I will add some.

vincent-tr:
👍 Thanks!

vincent-tr:
Is there any chance to get it released soon

PR #1048, opened 2020-02-12
feat: add support for files/unix sockets

### LINKED ISSUE #1040: Connect to unix socket
Would you consider support for connecting to unix sockets? The use case is a local application wanting to connect to a local MQTT broker without having to create a network.

It looks like `lib/connect/unix.js` would be very similar to `tcp.js`, but with this sort of construct instead:

```
    unixpath = '/var/run/mqtt.socket'

    return net.createConnection(unixpath)
```

ralight:
Closes #1040.


ralight:
Sure thing, could you give me a hint on where to start?

YoDaMa:
https://github.com/mqttjs/MQTT.js/blob/master/test/client.js is the tests for the regular MQTT.js client, and you'll note this test extends the abstact_client.js tests. Following this model but for the specific connection type should be the best option. Instead of a regular HTTPServer though you'll need something specific for the Unix Domain Socket.

robertsLando:
@ralight Could you add the required tests please?

robertsLando:

robertsLando:
This is already possible with standard `mqtt` protocol by using `path` in client options

PR #1055, opened 2020-02-21
feat: support SNI on TLS

YoDaMa:
carryover with tests of #1050

### DIFF
--- a/test/secure_client.js
+++ b/test/secure_client.js
@@ -153,5 +153,34 @@ describe('MqttSecureClient', function () {
         done()
       })
     })
+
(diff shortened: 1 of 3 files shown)


test/secure_client.js
```diff
@@ -153,5 +154,33 @@ describe('MqttSecureClient', function () {
         done()
       })
     })
+
+    it.only('should support SNI on the TLS connection', function (done) {
```
mcollina: This should not have landed as it has a `only` still in the tests.

YoDaMa: yes I'll revert it, I just wasn't getting Travis CI to even identify the branch to run testing.

PR #1059, opened 2020-02-24
chore: migration from Travis CI to Github Actions

YoDaMa:


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,24 @@
+name: Node.js CI
+
+on: [push]
```
mcollina: this needs to be on pr as well.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,26 @@
+name: Node.js CI
+
+on: [pull_request]
```
mcollina: This should include push as well.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    - run: npm run build --if-present
+      if: always()
+    - run: npm test
+      env:
+        CI: true
+        DEBUG: mqttjs*
```
mcollina: The debug statement should likely be removed before landing

package.json
```diff
@@ -22,13 +22,13 @@
...
   "scripts": {
-    "test": "node_modules/.bin/istanbul cover ./node_modules/mocha/bin/_mocha --report lcovonly --",
+    "test": "node_modules/.bin/nyc --reporter=lcov --reporter=text-summary ./node_modules/mocha/bin/_mocha",
     "pretest": "standard | snazzy",
     "tslint": "if [[ \"`node -v`\" != \"v4.3.2\" ]]; then tslint types/**/*.d.ts; fi",
     "typescript-compile-test": "tsc -p test/typescript/tsconfig.json",
     "typescript-compile-execute": "node test/typescript/*.js",
     "typescript-test": "npm run typescript-compile-test && npm run typescript-compile-execute",
-    "browser-build": "rimraf dist/ && mkdirp dist/ && browserify mqtt.js -s mqtt > dist/mqtt.js && uglifyjs < dist/mqtt.js > dist/mqtt.min.js",
+    "browser-build": "rimraf dist/ && mkdirp dist/ && webpack",
```
mcollina: Is this change needed for this PR? Maybe we should focus on the browser on the next PR.

test/abstract_client.js
```diff
@@ -2394,7 +2404,10 @@ module.exports = function (server, config) {
       function check () {
         if (serverPublished && clientCalledBack) {
           client.end()
+          debug('calling done')
           done()
+        } else {
+          debug('serverPublished %s and clientCalledBack %s', serverPublished, clientCalledBack)
         }
```
mcollina: I would not include debug statements in tests. Why were they added?

test/client.js
```diff
@@ -114,6 +115,9 @@ function buildServer (fastFlag) {
 server = buildServer().listen(port)
 
 describe('MqttClient', function () {
+  before(function (done) {
+    setTimeout(done, 10) // essentially set a 100ms buffer between each test
```
mcollina: Why is this needed? The overarching problem is that:

1. there is a shared broker. We should likely remove it.
2. there are a load of setTimeout or intervals both in the tests and in the code. This make the test suite very brittle

.github/workflows/nodejs.yml
YoDaMa: @mcollina on: [push, pull_request] will create redundant builds in the Github Actions. The result will be that every time a push is made on a PR, instead of just the tests running once they will all be run twice. See for instance: https://github.com/mqttjs/MQTT.js/actions/runs/44644983

test/client.js
YoDaMa: @mcollina I am seeing in the tests that oftentimes the results of one test bleed over into another test causing the test suite to fail when individual tests will succeed. This was an attempt to alleviate some of the issues of this bleed over.

package.json
YoDaMa: I can remove this from the PR, but it is broken so the Github Actions will fail.

test/abstract_client.js
YoDaMa: will remove.

mcollina:
push is needed because otherwise it won’t run on master.

Please avoid landing things when I leave a comment before I reply again, thanks.

YoDaMa:
@mcollina I see what you're saying, however the solution does not make sense to do as an array `[push, pull_request]`. Let me see a workaround for the master build...

PR #1060, opened 2020-02-26
refactor: webpack for browserify and nyc for istanbul

YoDaMa:
The decision to switch from browserify to webpack is twofold: 

1) uglify-js is having issues with what looks to be a dependency with ES6 Style Code. 
- The solution to this, removing the dependency of uglify-js, meant looking for an alternative. I considered babel but their plugin, `babel-minify`, is not a stable release yet, so I didn't want to take that on as a dependency. 

2) `webpack` has significantly more traction behind it right now, with 9M weekly downloads to Browserify's 700k. Weekly downloads are not everything but I think indicative. I found `webpack`'s documentation to be easier to follow to configure `minimize` and setting up the proper entry point etc. A good comparison of all JS Bundlers is here: [Choosing the Right JavaScript Bundler in 2020](https://blog.bitsrc.io/choosing-the-right-javascript-bundler-in-2020-f9b1eae0d12b)


YoDaMa:
@RangerMauve if you have thoughts let me know.

RangerMauve:
I've personally had a lot more trouble getting Webpack to play nice in different environments compared to Browserify.

This doesn't seem to be preventing people from choosing to use Browserify to compile this library within their application so it won't make a difference there. 😁

One question: Why the switch from `istanbul` to `nyc`?

YoDaMa:
@RangerMauve good question. 

`nyc` is the new cli tool provided by `istanbuljs`, and essentially the npm package `istanbul` is deprecated. See: https://github.com/gotwarlost/istanbul

package.json
```diff
@@ -103,6 +102,8 @@
     "tslint-config-standard": "^8.0.1",
     "typescript": "^3.2.2",
     "uglify-js": "^3.8.0",
```
anthonyvercolano: >uglify-js": "^3.8.0 [](start = 5, length = 19)

Were you going to remove this?

webpack.config.js
```diff
@@ -0,0 +1,13 @@
...
+  optimization: {
+    minimize: true
+  },
+  mode: 'production',
+  entry: './mqtt.js',
+  output: {
+    filename: 'main.js',
+    path: path.resolve(__dirname, 'dist')
+  }
+}
```
anthonyvercolano: Well the test failures don't have anything to do with the build.  So if that can be fixed you're good to go.

YoDaMa:
closing because after reviewing things I've decided @RangerMauve might be on to something with sticking with Browserify. All the fanfare around Webpack is justified from a web developer perspective creating a bundle with SASS, LESS, SCSS, JPGs, etc... but as this is a node.js project we are not looking at that.

Issue #533, opened 2017-02-03
changing options after client initialization


cordovapolymer:
Is it possible to change options after client initialization?
For example I'd like to change clientId on reconnect.

yohei1126:
@mcollina probably not?

RangerMauve:
It'd be easier to just create a new client instance, tbh.
However you could hack around that by modifying [client.options](https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L70) at runtime, though this is unsupported so don't complain if something you hack on top of it breaks. 😛

cordovapolymer:
@RangerMauve is it possible without modifying client.js?

RangerMauve:
@cordovapolymer You can try to modify the `options` property directly on an instance of an mqtt client.

```javascript
var client = mqtt.connect(something);
client.options.something = "something else";
```

cordovapolymer:
@RangerMauve , unfortunately it is not working this way.

burritoIand:
@cordovapolymer yes this is possible with `client.options.transformWsUrl`, see https://github.com/mqttjs/MQTT.js/issues/1083

YoDaMa:
`transformWsUrl` provides functionality for changing the URL or other options in websockets. For non-websockets, like @cordovapolymer said the solution would be to modify the `client.options` directly.

Issue #1083, opened 2020-05-01
Update documentation for client "reconnectPeriod" option.


burritoIand:
In exploring the use of this library I've been trying to wrap my head around how reconnection works etc.

The documentation in the readme states "milliseconds, interval between two reconnections". Meanwhile I am using mqtt to connect to AWS IoT service and their v2 iot js sdk uses another aws common library called [aws-crt for nodejs](https://github.com/awslabs/aws-crt-nodejs/blob/master/lib/browser/mqtt.ts#L123). In that library you'll notice the default value for `reconnectPeriod` is `0`, whereas the default value in this library is `1000`.

I noticed that when the value is set to `0`, my client will close and then never try to reconnect again. This led me to look at the source code which makes it clear that if the value is 0 then it does not schedule a reconnection.

https://github.com/mqttjs/MQTT.js/blob/d764f02552639e9fed0f2b2039ca86d9941ba0cb/lib/client.js#L940
https://github.com/mqttjs/MQTT.js/blob/d764f02552639e9fed0f2b2039ca86d9941ba0cb/lib/client.js#L904
https://github.com/mqttjs/MQTT.js/blob/d764f02552639e9fed0f2b2039ca86d9941ba0cb/lib/client.js#L651

It's not clear from the documentation that setting `reconnectPeriod: 0` will make the client flush the connection at end and not try to reconnect. Also please correct me if I'm wrong about how that works. If it's true, I feel like the AWS library should also tell users that setting `reconnectPeriod` to 0 means the client will not reconnect upon disconnection.

*Side Note*: I think it'd be very helpful to have a dedicated "Reconnection" section in this documentation towards the top that describes how reconnection works and how to refresh the client option data using transformWsUrl. The transformWsUrl docs are at the bottom and also aren't clear that you _can_ manipulate more than the url in that hook. For instance I needed to update the `options.username` and `options.clientId`.
```
  transformWsUrl(url, options, client) {
    // update client options with latest auth data.
    client.options.username = `${this.prefix}&token=${this.token}`;
    client.options.clientId = `${this.clientId}`;

    return url;
  }

```
related:
https://github.com/mqttjs/MQTT.js/issues/533  - resolved question on updating reconnect data
https://github.com/mqttjs/MQTT.js/issues/602 - questions answered below & will be added to readme reconnection section
https://github.com/mqttjs/MQTT.js/issues/1019 - resolved question on updating reconnect docs
[...]

Issue #1019, opened 2019-12-09
Documentation about Reconnection handling 


fabik111:
Hi,
I had added a callback on the `offline` event that, in case is fired, it  will call the `reconnect()` method.   
But I've experienced some re-connection loop after the first offline event, so I've deduced that there is already a re-connection policy.
Can you improve the documentation about `reconnect()` method and default re-connection attempts?

PHPJourney:
Hi, How do you solution this question？
I cannot use reconnect function. It's like not running

fabik111:
Hi,
I don't use reconnect function, I think there is already an auto re-connection policy, so if you lose connectivity automatically it re-establish the connection. My issue is not about a bug but about bad documentation

PHPJourney:
> Hi,
> I don't use reconnect function, I think there is already an auto re-connection policy, so if you lose connectivity automatically it re-establish the connection. My issue is not about a bug but about bad documentation
I'm not see automatically it re-establish the connection work. My app in wi-fi offline state. wait a moment open it. It's can't reconnect.

fabik111:
I'm sorry but I can't help you, also for me this is unknown

Issue #602, opened 2017-04-24
reconnect documentation


ozomer:
It is not clear from the documentation what the reconnectPeriod option covers and what it does not.
If the first connection attempt does not succeed, will there be a reconnection attempt?
If the first connection attempt succeeds and then fail, will there be a reconnection attempt?
What are the other events that would happen in each scenario, regarding 'error', 'close', 'offline' and 'connect' events? In what order?
Is it required to re-subscribe in each 'connect' event? 
Will a "publish" attempt during a reconnection return an error immediately or be queued? Will a "publish" attempt before the first connection succeed?
Will the "reconnect" event happen for the first connection attempt?

Please add some example scenarios. Thanks!

mcollina:
@ozomer the best way to do write this doc is to look at the code, and fill in the documentation.
I'll be happy to review.

ozomer:
Well, I just thought it would be nice to have a state-machine diagram by the developers themselves.

mcollina:
@ozomer this is done by volunteers, so you (or anybody else) can be one! :)

burritoIand:
@ozomer were you able to clarify any of the reconnection logic? It's definitely lacking and gives me some slight pause in using this, but the wss service I'm using is based on mqtt so...

PR #1088, opened 2020-05-05
Update reconnection docs.

burritoIand:
changes
  - Add "About Reconnection" towards top of doc explaining transformWsUrl and other reconnection behavior.

https://github.com/mqttjs/MQTT.js/issues/1083

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,10 +183,13 @@ Example (update clientId & username on each reconnect):
 
     const connection = await mqtt.connectAsync(<wss url>, {
       ...,
-      transformWsUrl,
+      transformWsUrl: transformUrl,
     });
 
 ```
+Now every time a new WebSocket connection is opened (hopefully not too often),
+we will get a fresh signed url or fresh auth token data.
+
 Note: Currently this hook does _not_ support promises, meaning that in order to
 use the latest auth token, you must have some outside mechanism running that
 handles application-level authentication refreshing so that the websocket
--- a/README.md
+++ b/README.md
@@ -151,6 +151,60 @@ $env:DEBUG='mqttjs*'
 
 ```
 
+<a name="reconnecting"></a>
+## About Reconnection
+
+An important part of any websocket connection is what to do when a connection
+drops off and the client needs to reconnect. MQTT has built-in reconnection
+support that can be configured to behave in ways that suite the application.
+
+#### Refresh Authentication Options / Signed Urls
+
+When an mqtt connection drops and needs to reconnect, it's common to require
+that any authentication associated with the connection is kept current with
+the underlying auth mechanism. For instance some applications may pass an auth
+token with connection options on the initial connection, while other cloud
+services may require a url be signed with each connection.
+
+By the time the reconnect happens in the application lifecycle, the original
+auth data may have expired.
+
+To address this we can use a hook called `transformWsUrl` to manipulate
+either of the connection url or the client options at the time of a reconnect.
+
+Example (update clientId & username on each reconnect):
+```
+    const transformWsUrl = (url, options, client) => {
+      client.options.username = `token=${this.get_current_auth_token()}`;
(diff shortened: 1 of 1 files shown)


burritoIand:
@YoDaMa started to address reconnection clarification in docs (https://github.com/mqttjs/MQTT.js/issues/1083)

I think this could be closed https://github.com/mqttjs/MQTT.js/issues/533

Further I was thinking to address some of the questions in https://github.com/mqttjs/MQTT.js/issues/602 (?) I don't know the library well enough yet to do so without studying it a bit :)

Open to feedback/changes

burritoIand:
I think https://github.com/mqttjs/MQTT.js/issues/1019 could be closed as well as the default-reconnection is explained by usage of `reconnectPeriod`

README.md
```diff
@@ -151,6 +151,60 @@ $env:DEBUG='mqttjs*'
 
 ```
 
+<a name="reconnecting"></a>
+## About Reconnection
+
+An important part of any websocket connection is what to do when a connection
+drops off and the client needs to reconnect. MQTT has built-in reconnection
+support that can be configured to behave in ways that suite the application.
```
YoDaMa, line 159: suite -> suit.

YoDaMa:
@burritoIand this looks to be somewhat WS specific, correct?

README.md
burritoIand: oops

burritoIand:
@YoDaMa correct, updated to mention that

README.md
```diff
@@ -151,6 +151,60 @@ $env:DEBUG='mqttjs*'
...
+    const transformWsUrl = (url, options, client) => {
+      client.options.username = `token=${this.get_current_auth_token()}`;
+      client.options.clientId = `${this.get_updated_clientId()}`;
+
+      return `${this.get_signed_cloud_url(url)`;
+    }
+
+    const connection = await mqtt.connectAsync(<wss url>, {
+      ...,
+      transformWsUrl,
```
YoDaMa, line 186: this should be a key value parameter, so replace with:

```
    const connection = await mqtt.connectAsync(<wss url>, {
      transformWsUrl: transformWsUrl,
    });

```

README.md
```diff
@@ -641,29 +695,6 @@ you can then use mqtt.js in the browser with the same api than node's one.
...
-var awsIotUrlSigner = require('awsIotUrlSigner')
-mqtt.connect('wss://a2ukbzaqo9vbpb.iot.ap-southeast-1.amazonaws.com/mqtt', {
-  transformWsUrl: function (url, options, client) {
-    // It's possible to inspect some state on options(pre parsed url components)
-    // and the client (reconnect state etc)
-    return awsIotUrlSigner(url)
-  }
-})
-
-// Now every time a new WebSocket connection is opened (hopefully not that
```
YoDaMa, line 662: can you add this line to your explanation, i think it is a nice clear and concise way of explaining what the code snippet is doing.

burritoIand: sounds good, added. Do you want to still mention "awsIotUrlSigner"? I removed the specific aws mention since there could be other cloud providers that do the same thing. But, how this came up for me in the first place is that I'm using aws iot with their new custom authentication beta which lets you pass auth data to your own custom backend iot authorizer via the username/password fields etc. I wasn't sure if mentioning specific cloud solutions made sense for this repo (?)

README.md
burritoIand: btw in es6 they introduced property shorthands :) (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer)

README.md
YoDaMa: I mean I am biased since I am at Microsoft... but removing the specific cloud provider from the sample seems like a nice thing to do. I'd like for this repo to remain company agnostic

burritoIand: ok yeah I agree, the cloud providers can always reference in their own docs

PR #1094, opened 2020-05-08
feat: unix sockets

YoDaMa:
Related to: #1040 
This PR is still in the works, but carrys over from @ralight's PR for Unix Domain Sockets support for MQTT.js. I see there is lots of discussion around MQTT.js in Mosquitto that @ralight has participated in, and the resolution was something along the lines of, there is an actionable and useful security gain for having unix domain socket support for a service like mosquitto, so they are looking to add it. 

From the MQTT.js perspective, this should be a minimal change. This PR is not complete quite yet, I think I have added some of the changes to make it work, but I want to add a unix domain sockets sample and also add the tests to support it. 

For the tests, it should be very simple, just adding on a new flavor of the existing MQTT-Connection based server, except in this case listening on a unix domain port. There should be a check that the tests are running on a POSIX OS (i.e. not Windows) since obviously the tests will fail without the support of unix domain sockets on the OS.

### DIFF
--- a/examples/unix-socket/unix-socket-client.js
+++ b/examples/unix-socket/unix-socket-client.js
@@ -1,0 +1,38 @@
+var mqtt = require('../..')
+
+const fs = require('fs')
+const connections = {}
+let client
+
+// prevent duplicate exit messages
+var SHUTDOWN = false;
+
+// Our socket
+const SOCKETFILE = '/tmp/unix.sock';
+
+console.info('Loading interprocess communications test');
+console.info('Socket: %s \n  Process: %s',SOCKETFILE,process.pid);
+
+// Connect to server.
+console.log("Connecting to server.");
+let opts = { protocol: 'file' }
+client = mqtt.connect(SOCKETFILE, opts)
+
+client.subscribe('presence')
+client.publish('presence', 'Hello mqtt')
+
+client.on('message', function (topic, message) {
+  console.log(message.toString())
+})
+
+client.end()
+
+
+function cleanup(){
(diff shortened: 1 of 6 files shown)


jdiamond:
Is `file:` in the URL flexible enough? It might be rare, but what if you want to use web sockets over a unix domain socket instead of plain MQTT? Or mqtts or wss?

Check out how the request library uses unix domain sockets:

https://www.npmjs.com/package/request#unix-domain-sockets

Not sure if that format is standardized and other tools use it or not.

`curl` supports unix sockets, too, but you use a normal URL and a separate `--unix-socket` option to specify the path to the socket file.

README.md
```diff
@@ -237,7 +237,7 @@ Connects to the broker specified by the given url and options and
 returns a [Client](#client).
 
 The URL can be on the following protocols: 'mqtt', 'mqtts', 'tcp',
-'tls', 'ws', 'wss'. The URL can also be an object as returned by
+'tls', 'ws', 'wss', 'file' (e.g. unix socket). The URL can also be an object as returned by
```
OmgImAlexis, line 240: `unix` should be used not `file` as it needs to be a unix socket not just a normal file.

OmgImAlexis:
@jdiamond is right, you can use http/ws with a unix socket just the same as you can a port. So really this should just use http/ws as the protocol and then a socket or port connection should depend on the connection url as the `request` library does.

ralight:
The format that the request library use is also used by nginx, so it's not exactly an uncommon format.

natcl:
Why has this been closed ?
support for unix sockets would be great !

Issue #1008, opened 2019-11-14
Building with Webpack have issues with  #!/usr/bin/env node in mqtt.js


savnik:
When building with webpack on seed.run following webpack error occours:

`Serverless: Bundling with Webpack...
ERROR in /tmp/seed/source/node_modules/async-mqtt/node_modules/mqtt/mqtt.js 1:0
Module parse failed: Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> #!/usr/bin/env node`

Following the webpack issue list here is a similar type of issues: https://github.com/webpack/webpack/issues/4603

They add appropriate loaders to to the webpack configuration.
This could probably be resolved within the MQTT package instead of in each project?

GauravWable:
i am also facing error with webpack

`npm ERR! code ELIFECYCLE
npm ERR! errno 3221226505
npm ERR! ble-collect@1.0.0 start: `webpack && qode ./dist/index.js`
npm ERR! Exit status 3221226505
npm ERR!`

hieutranagi47:
I am facing the same problem with #!/usr/bin/env node. I use webpack to build my node cli tool.

savnik:
@hieutranagi47 I did a dirty fix and forked the mqtt lib and removed shebang. 
This fixed the issue for me temporary.
https://www.npmjs.com/package/mqtt-no-shebang

hieutranagi47:
> @hieutranagi47 I did a dirty fix and forked the mqtt lib and removed shebang.
> This fixed the issue for me temporary. In order to use it, i have put it into a package.
> https://www.npmjs.com/package/mqtt-no-shebang

Thank Savnik so much! I'll try it. :)

savnik:
I see there is a fix in place now. I will test it and if all okay, i will deprecate the temporary fix in https://www.npmjs.com/package/mqtt-no-shebang

PR #1096, opened 2020-05-11
chore: move cli to bin directory

OmgImAlexis:
This hasn't been fully verified but it should fix https://github.com/mqttjs/MQTT.js/issues/1008

### DIFF
--- a/bin/mqtt.js
+++ b/bin/mqtt.js
@@ -1,0 +1,24 @@
+#!/usr/bin/env node
+'use strict'
+
+/*
+ * Copyright (c) 2015-2015 MQTT.js contributors.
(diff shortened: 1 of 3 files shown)


OmgImAlexis:
What isn’t? You kinda need to provide details for us to help.

DavidMG01:
> What isn’t? You kinda need to provide details for us to help.

The mqtt.js into bin directory refers to ./bin/pub and ./bin/sub, but these are in the same directory, I corrected the references to ./pub and ./sub but fail all...

If you install gobally and check may be you see the issue.

OmgImAlexis:
I’ll check in the morning and open a PR.

Issue #1074, opened 2020-04-03
The IClientPublishOptions interface is missing the properties attribute


wilcool:
The IClientPublishOptions interface in the [client-options.d.ts](https://github.com/mqttjs/MQTT.js/blob/master/types/lib/client-options.d.ts#L133-L150) file is missing the properties attribute

![image](https://user-images.githubusercontent.com/28377898/78345993-ff97b080-75d0-11ea-8fe3-ae2caec9bc38.png)

sublimator:
> IClientPublishOptions 

Awwww such Java/C# ish names

YoDaMa:
I'mma say this is a bug... we have no testing for typescript interfaces.... we should probably have testing for the typescript interfaces....

Also this opens up one of the spiciest topics in TypeScript naming conventions, "I" prefixing interfaces or not. See: https://github.com/microsoft/TypeScript-Handbook/issues/121 and also https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript/41967120#41967120

I'm not going to advocate for changing all the interfaces to non "I" prefixed, but we can definitely discuss it.

PR #1103, opened 2020-05-18
Added MQTT 5 properties object to publish options

koopman-mark:


koopman-mark:
Addresses issue #1074

aemaem:
Is this PR just hanging because of the 10.x build?

Global-Ekons:
Its hanging because it is missing its system32 folder.

Global-Ekons:
Error 104

The system32 folder is mia, now it doesn't matter what I try, won't even
reinstall win10, as it doesn't pass Boot.

Boot option doesn't work either, or any other command doesn't work either.


LAPTOP, DELL LATITUDE E6540, 64bit with 16gb ram.
Win10 System was never reinstalled after 2016, as updates did enough, was
though a Win8.1 beforehand.

highcaliberkevin:
I am trying to use the `.publish(...)` method's `responseTopic` and a Mosquitto broker (which has v5 support). I would absolutely love for this to be merged--is there anything that can be done (perhaps I can assist? testing, et al) to get it in there? Thanks!

Issue #1106, opened 2020-05-26
Error 'MODULE_NOT_FOUND'


DavidMG01:
I installed mqtt globally on windows 10, but when i type "mqtt -v" getting the following error:

internal/modules/cjs/loader.js:960
  throw err;
  ^

Require stack:
- C:\Users\DavidMG\AppData\Roaming\npm\node_modules\mqtt\bin\mqtt.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.
    at Function.Module._load (internal/modules/cjs/loader.js:840:27)
    at Module.require (internal/modules/cjs/loader.js:1019:19)
    at require (internal/modules/cjs/helpers.js:77:18)
    at Object.<anonymous> (C:\Users\DavidMG\AppData\Roaming\npm\node_
9)
    at Module._compile (internal/modules/cjs/loader.js:1133:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:
/run_main.js:74:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'C:\\Users\\DavidMG\\AppData\\Roaming\\npm\\node_modules\\mqtt\\bin\\mq
tt.js'
  ]

When you moved mqtt.js to ./bin, the reference to ./bin/pub and ./bin/sub was lost...

derberg:
Same here on Mac, when I downgrade to 4.0.0 all works fine

DavidMG01:
How can I downgrade to 4.0.0.0 or install this specific release??

jdiamond:
```
npm i mqtt@4.0.0
```

DavidMG01:
Thanks a lot!!

PR #1107, opened 2020-05-29
fix: path for bin files

YoDaMa:
This is a fix from PR #1096 , which moved some of the CLI functionality into the bin folder. 

We need to add some form of integration testing so this doesn't happen again...

Addresses: #1106

### DIFF
--- a/bin/mqtt.js
+++ b/bin/mqtt.js
@@ -7,14 +7,19 @@
  *
  * See LICENSE for more information
  */
-
+var path = require('path')
 var commist = require('commist')()
-var helpMe = require('help-me')()
+var helpMe = require('help-me')({
+  // the default
+  dir: path.join(path.dirname(require.main.filename), '/../doc'),
+  // the default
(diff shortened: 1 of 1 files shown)


YoDaMa:
@jdiamond if you have ideas on integration tests for CLI code, I don't have knowledge in that subject.

YoDaMa:
also should resolve https://github.com/asyncapi/website/issues/107

jdiamond:
Maybe a test could be to use `child_process.exec` to run `node ./bin/mqtt.js` and assert that the exit code is 0?

YoDaMa:
@jdiamond I'll add that test before I merge.

OmgImAlexis:
@YoDaMa did you want to merge and add the test after to get this back out and working for users?

YoDaMa:
Yeah let’s do that. I was going to get to the tests but I’ve been struggling to get anything done this week. Shit is heavy.

PR #1108, opened 2020-05-29
feat: websocket-streams to ws

YoDaMa:


YoDaMa:
the errors I'm seeing on Node version 10 are not consistent and are on functionality not related to streams, so I'm going to consider this passing. The flakiness of tests still needs to be improved...

fartagaintuxedo:
After this fix in v4.2 we are getting `Uncaught (in promise) Error: ws does not work in the browser. Browser clients must use the native WebSocket object` in all our applications.

YoDaMa:
This bug is being tracked.

PR #1134, opened 2020-07-30
correct browser detection

nosovk:
currently webpack based projects couldn't import mqtt.js because incorrect browser detection.
It's not a problem for react, where node-shims enabled, but a pain for angular (where enabling node-shims isn't supported officialy)
To fix that issue we need to check if `process` exists (it will be not avalilible in browser)
Also due to shims in webpack we need to check are we in process of bundling, or not.

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -143,6 +143,19 @@ export interface IClientPublishOptions {
    * whether or not mark a message as duplicate
    */
   dup?: boolean
+  /*
+   *  MQTT 5.0 properties object
+   */
+  properties?: {
+    payloadFormatIndicator?: number,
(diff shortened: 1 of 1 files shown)


nosovk:
@YoDaMa nope, please undo, there was an error with failing tests, I've fixed it

YoDaMa:
this wasn't merged @nosovk ?

nosovk:
https://github.com/mqttjs/MQTT.js/pull/1135
seems to be that no.
I was squashing history in parallel with merge.
Seems all is ok, created new PR with test working.

nosovk:
sorry for inconviance, just wasn't ready that branch with failing tests could be merjed.

PR #1135, opened 2020-07-30
correct browser detection

nosovk:
currently webpack based projects couldn't import mqtt.js because incorrect browser detection.
It's not a problem for react, where node-shims enabled, but a pain for angular (where enabling node-shims isn't supported officially)
To fix that issue we need to check if `process` exists (it will be not available in browser)
Also due to shims in webpack we need to check are we in process of bundling, or not.

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -8,7 +8,8 @@ var debug = require('debug')('mqttjs')
 
 var protocols = {}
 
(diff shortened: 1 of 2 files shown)


YoDaMa:
@nosovk is this PR finished?

nosovk:
yes, this one is done. Thanks

PR #1136, opened 2020-07-30
do not use process.nextTick without check that it exists

nosovk:
if we do not add shims to browser `process.nextTick` does not exist, because of missing `process`.
In v8 eventloop sequence is:

```javascript
process.nextTick(callback) // runs first, after current Tick
setTimeout(callback, 0) // runs after nextTick
setImmediate(callback) // runs after all (https://github.com/nodejs/node-v0.x-archive/issues/25788#issuecomment-128869483)
```

That's why where is possible (in destructors) process.nextTick substituted with setTimeout.
Where it's significant - substitution will be done if process.nextTick is missing.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -12,9 +12,10 @@ var reInterval = require('reinterval')
 var validations = require('./validations')
 var xtend = require('xtend')
 var debug = require('debug')('mqttjs:client')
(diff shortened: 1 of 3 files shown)


artesan-1:
Are you sure that any dependant packages don't have the same behavioure?

PR #1138, opened 2020-08-12
release(4.2.0): patch update

YoDaMa:


YoDaMa:
again, seeing these build failures in v10, but locally the websockets tests in v10 are passing, so I will merge.

PR #1145, opened 2020-08-18
fix: browser in ws

YoDaMa:


lib/connect/ws.js
```diff
@@ -1,20 +1,25 @@
...
-var WebSocket = require('ws')
-var debug = require('debug')('mqttjs:ws')
-var urlModule = require('url')
-var WSS_OPTIONS = [
+const WS = require('ws')
+const debug = require('debug')('mqttjs:ws')
+const duplexify = require('duplexify')
+var Buffer = require('safe-buffer').Buffer
+const urlModule = require('url')
+var Transform = require('readable-stream').Transform
```
ysfscream, line 8: Replace `var` to `const` or `let`?

PR #1151, opened 2020-08-28
feat: allow user to disable pre-generated write cache

mwohlert:
This change allows the user of the library to disable the pre-generated write cache generated by mqtt-packet and instead rely on in-flight generated buffers. This helps to preserve memory in low-memory environments, but possibly degrading mqtt-stream write performance.


ynagasaki:
This is an old PR, but I wanted to bump it. I also encountered the same use case and this PR would fix our issue nicely.

In my case, our project swaps all MQTT imports with the `dist/mqtt.min.js` that is included in this project's module. Because of this, we are unable to access the `writeToStream.cacheNumbers` property in the embedded `mqtt-packet` module that is bundled in `mqtt.min.js`.

At the moment, we're unable to remove the `mqtt.dist.js` replacement, so it would really help if this setting was exposed by this package also. Thanks.

robertsLando:
@mwohlert Can you add a unit test for this?

ynagasaki:
Hi @robertsLando, I added a unit test on behalf of @mwohlert

robertsLando (approved):

robertsLando:
@ynagasaki This will be available on next beta. Doing it right now.

PR #1167, opened 2020-09-29
Renovate to nodejs 10

qm3ster:
Just smacked some lowest hanging fruit, you know?

Got rid of a bit of polyfills.


qm3ster:
@YoDaMa any ideas what causes failure on node 10?
Is it `aedes`? Seems like it would be some independent process, because it causes a very long spam of network-related error.

YoDaMa:
thanks for the contribution @qm3ster. 

There's been some issues with the Node 10 build for a minute now. I haven't been able to pin down what's the issue. Let me look at the logs.

YoDaMa (approved):

YoDaMa:
If you run the websocket tests locally on node v10 and they pass, then I can merge this PR.

YoDaMa:
should also fix https://github.com/mqttjs/MQTT.js/issues/1168

qm3ster:
They don't pass, nor do they pass on unmodified master.
It's actually happening in dependencies, it looks like I will have to look at them next:
```
      at DestroyableTransform.Transform.push (node_modules/readable-stream/lib/_stream_transform.js:148:32)
      at Parser.push (node_modules/mqtt-connection/lib/parseStream.js:19:12)
      at Parser._newPacket (node_modules/mqtt-packet/parser.js:669:12)
      at Parser.parse (node_modules/mqtt-packet/parser.js:43:45)
```
However, since this doesn't change the error relative to `master`, I would personally merge the PR.

YoDaMa:
yes.. this has been a difficult issue to pinpoint on the v10 failures. I will merge since you've done the local tests though.

qm3ster:
Some notes on the remaining dependencies:
dependencies only cli uses:
```json5
    "commist": "^1.0.0",
    "concat-stream": "^2.0.0",
    "help-me": "^1.0.1",
    "minimist": "^1.2.5",
    "pump": "^3.0.0",
    "split2": "^3.1.0"
```
three of these depend on the naughty `readable-stream` that is ruining my rollup experience and stuff.
the cli "binaries" should probably be a separate npm module, even in the same repo, that depends on core.
I also have concerns about "WeChat Mini Program" and the likes being in the core library.

dependencies that emulate syntax constructs that could be transpiled with typescript/babel:
```json5
    "inherits": "^2.0.3",
    "xtend": "^4.0.1", // this is not deepmerge, it's just `{...a, ...b, ...c}`
```
    
node 10 supports both class syntax and spread, and for web people transpile their bundle anyway

normal:
```json
    "debug": "^4.1.1",
    "mqtt-packet": "^6.3.2", // <- project's own module, and where nodejs 10 error happens
    "reinterval": "^1.1.0", // <- microdependency, only used in ping, I usually roll my own, no harm no foul I guess, it doesn't have its own transient dependencies.
```

YoDaMa:
I agree on both accounts. I am inclined to offer WeChat folks the ability to fork this repo if they need to edit the core code, but otherwise their ali specific code should not live here. As for the CLI dependencies, a separate packaging would be a good major version bump proposal that I could get behind.

qm3ster:
Perhaps I was hasty about https://github.com/mqttjs/MQTT.js/pull/1167/commits/0ff6c182e4e0f8b3d3163ffa36791b73756e12b0, might revert.
Despite there not being a breaking change since `node` `0.11`, the status quo remains that everyone uses `readable-stream` and not `stream` builtin. https://github.com/nodejs/readable-stream/issues/398

Issue #1193, opened 2020-10-28
TypeError [ERR_INVALID_URL] after updating to 4.2.3


youngcholo:
I'm getting this error after updating to 4.2.3

```
TypeError [ERR_INVALID_URL] [ERR_INVALID_URL]: Invalid URL: [object Object]
    at onParseError (internal/url.js:243:9)
    at new URL (internal/url.js:319:5)
    at Object.connect (C:\Program Files (x86)\NLT\bin\node_modules\mqtt\lib\connect\index.js:64:18)
    at Object.connect (C:\Program Files (x86)\NLT\bin\node_modules\async-mqtt\index.js:139:25)
```
```
mqtt: {
    connection: {
      host: {
        doc: "The URL of the mqtt server.",
        format: "mqtt-url",
        default: "tcp://localhost",
        env: "MQTT",
        arg: "mqtt"
      },
      port: {
        doc: "The port to connect to for the mqtt server.",
        format: "port",
        default: 1883,
        env: "MQTT_PORT"
      }
    }
  }
```
const connectionString = url.parse(conf.get("mqtt").connection.host);
connectionString.port = conf.get("mqtt").connection.port;
mqtt.connect(connectionString, { clientId: this._clientId, clean:false })

this was working yesterday with 4.2.1

AB#8679346

YoDaMa:
@youngcholo thanks for filing the issue. What is the connection string you are providing?

YoDaMa:
is there a minimal repro you can provide?

yurievangelista:
Same is happening here with a pre-signed URL string. 
What I've managed to observe is that the opts parameter in the **streamBuilder** function seems to lose the `href` property somewhere along the way, and I'm left with an empty ("/") path.
I'm using the library in a browser, if that's any help.

youngcholo:
Before 4.2.3 and as specified in the docs,  " The URL can also be an object as returned by URL.parse()"
https://repl.it/join/yhjykmju-youngcholo

YoDaMa:
@youngcholo yeah looks like a bug again with the url parsing. Can you incorporate the changes from the PR linked manually and see if they fix things?

youngcholo:
Apologies for the late response. The issue still persists in 4.2.4.
https://repl.it/@youngcholo/mqtttest424

YoDaMa:
[...]

PR #1194, opened 2020-10-29
fix(ws): add all parts of object to opts

YoDaMa:
this is to fix the regression published in v4.2.3

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -70,7 +70,6 @@ function connect (brokerUrl, opts) {
     opts.hostname = parsed.hostname
     opts.href = parsed.href
     opts.origin = parsed.origin
-    opts.path = parsed.pathname
     opts.pathname = parsed.pathname
     opts.port = Number(parsed.port) || null
     opts.protocol = parsed.protocol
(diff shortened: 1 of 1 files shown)


YoDaMa:
related #1193

lib/connect/index.js
```diff
@@ -62,14 +62,20 @@ function connect (brokerUrl, opts) {
...
     // the URL object is a bit special, so copy individual
     // items to the opts object
-    opts.hostname = parsed.hostname
+    opts.hash = parsed.hash
     opts.host = parsed.host
-    opts.protocol = parsed.protocol
+    opts.hostname = parsed.hostname
+    opts.href = parsed.href
+    opts.origin = parsed.origin
+    opts.pathname = parsed.pathname
```
ricardoffp, line 73: Hi 👋 

I just upgraded the version of MQTT.js I'm using to the latest one (4.2.3).
Now I have an error while connecting because the path is not being included in the URL, even after this PR.
Did a few tests and added an extra `opts.path = parsed.pathname` and now the path is passed down and it connects. Can you take a look?
For example, an URL like: `wss://example.server/mqtt` would try to connect to `wss://example.server`.
Thank you!

PR #1197, opened 2020-11-02
Don't send empty username and password

jackwilsdon:
Since 70a247c changed how we generate options, we now end up sending an empty username and password when they are not set, instead of sending _no_ username and password. This ends up confusing some MQTT servers, which don't properly detect an empty username/password as an anonymous login.

This can be seen most clearly in the connect flags:

### b5b3814 (one commit before 70a247c)
```
Connect Flags: 0x02, QoS Level: At most once delivery (Fire and Forget), Clean Session Flag
    0... .... = User Name Flag: Not set
    .0.. .... = Password Flag: Not set
    ..0. .... = Will Retain: Not set
    ...0 0... = QoS Level: At most once delivery (Fire and Forget) (0)
    .... .0.. = Will Flag: Not set
    .... ..1. = Clean Session Flag: Set
    .... ...0 = (Reserved): Not set
```

### 70a247c
```
Connect Flags: 0xc2, User Name Flag, Password Flag, QoS Level: At most once delivery (Fire and Forget), Clean Session Flag
    1... .... = User Name Flag: Set
    .1.. .... = Password Flag: Set
    ..0. .... = Will Retain: Not set
    ...0 0... = QoS Level: At most once delivery (Fire and Forget) (0)
    .... .0.. = Will Flag: Not set
    .... ..1. = Clean Session Flag: Set
    .... ...0 = (Reserved): Not set
```

Setting the options to null when they're false-y seems to fix this.

### DIFF
--- a/lib/connect/index.js
+++ b/lib/connect/index.js
@@ -73,8 +73,8 @@ function connect (brokerUrl, opts) {
     opts.pathname = parsed.pathname
     opts.port = Number(parsed.port) || null
     opts.protocol = parsed.protocol
-    opts.username = opts.username || parsed.username
-    opts.password = opts.password || parsed.password
+    opts.username = opts.username || parsed.username || null
+    opts.password = opts.password || parsed.password || null
     opts.search = parsed.search
     opts.searchParams = parsed.searchParams
     opts.path = parsed.pathname + parsed.search


ccvca:
Just some more background information, as I just tracked down the same Problem (#1207)

It may be also possible to fix this in mqtt-packet here:

https://github.com/mqttjs/mqtt-packet/blob/2dcc70c88f0b8b146b63f017f816568fdc453ebf/writeToStream.js#L259-L265

```js
  // Username and password
  if (username != null) {
    writeStringOrBuffer(stream, username)
  }
  if (password != null) {
    writeStringOrBuffer(stream, password)
  }
```

zambujal:
My mqtt addon mosquito frezes and then nothing maybe because of this? not even showing errors... tried several updates... so this is just bugs...

Issue #1183, opened 2020-10-19
server's topicAliasMaximum is not relevant to client's topicAliasMaximum


heunghingwan:
hi,
I think this is a bug, topicAliasMaximum in CONNACK is for client to send to server, it should not overwrite CONNECT options, this will be a problem when reconnect.
https://github.com/mqttjs/MQTT.js/blob/70a247c29e0b05ddd8755e7b9c8c41a4c25b431b/lib/client.js#L1174

AB#8607503

YoDaMa:
please create a PR addressing it if it is a bug.

heunghingwan:
fix by #1301 (#1300 for detail)

PR #1200, opened 2020-11-03
Differentiate CONNECT packets properties and CONNACK packets properties

heunghingwan:
Addressing bug in #1183


lib/client.js
```diff
@@ -1171,15 +1171,15 @@ MqttClient.prototype._handleConnack = function (packet) {
...
+      options.properties.serverTopicAliasMaximum = packet.properties.topicAliasMaximum
     }
     if (packet.properties.serverKeepAlive && options.keepalive) {
       options.keepalive = packet.properties.serverKeepAlive
       this._shiftPingInterval()
     }
     if (packet.properties.maximumPacketSize) {
       if (!options.properties) { options.properties = {} }
-      options.properties.maximumPacketSize = packet.properties.maximumPacketSize
+      options.properties.serverMaximumPacketSize = packet.properties.maximumPacketSize
```
YoDaMa, line 1182: you've changed this but where is options.properties.serverMaximumPacketSize referenced?

lib/client.js
```diff
@@ -503,11 +503,11 @@ MqttClient.prototype.publish = function (topic, message, opts, callback) {
   if (options.protocolVersion === 5) {
     packet.properties = opts.properties
     if ((!options.properties && packet.properties && packet.properties.topicAlias) || ((opts.properties && options.properties) &&
-      ((opts.properties.topicAlias && options.properties.topicAliasMaximum && opts.properties.topicAlias > options.properties.topicAliasMaximum) ||
-        (!options.properties.topicAliasMaximum && opts.properties.topicAlias)))) {
+      ((opts.properties.topicAlias && options.properties.serverTopicAliasMaximum && opts.properties.topicAlias > options.properties.serverTopicAliasMaximum) ||
```
YoDaMa, line 506: do you know that this change honors the MQTTv5 Spec definition for topicAliasMaximum: 

34 (0x22) Byte, Identifier of the Topic Alias Maximum.
811 Followed by the Two Byte Integer representing the Topic Alias Maximum value. It is a Protocol Error to
812 include the Topic Alias Maximum value more than once. If the Topic Alias Maximum property is absent,
813 the default value is 0.
814
815 This value indicates the highest value that the Client will accept as a Topic Alias sent by the Server. The
816 Client uses this value to limit the number of Topic Aliases that it is willing to hold on this Connection. The
817 Server MUST NOT send a Topic Alias in a PUBLISH packet to the Client greater than Topic Alias
818 Maximum [MQTT-3.1.2-26]. A value of 0 indicates that the Client does not accept any Topic Aliases on
819 this connection. If Topic Alias Maximum is absent or zero, the Server MUST NOT send any Topic Aliases
820 to the Client [MQTT-3.1.2-27]

YoDaMa:
the option `topicAliasMaximum` should validate that the server does not send a topic value greater than specified. It should default to 0 aka no topic aliases accepted. 

What you're proposing, `serverTopicAliasMaximum` will be for making sure that on publish the MQTT client does not publish a Topic Alias that is greater than the specified Maximum provided by the server on the CONNACK. 

The `serverMaximumPacketSize` also doesn't seem like it is correct.

lib/client.js
heunghingwan: It's not use anywhere, It should be use when client sending message, but I don't know the code enough to do it properly, may be this will better translate it meaning?
```suggestion
      options.serverProperties.maximumPacketSize = packet.properties.maximumPacketSize
```

lib/client.js
heunghingwan: It should honor topicAliasMaximum property in CONNACK packet, not CONNECT packets property sent by the client, as shown below: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901088

YoDaMa:
@heunghingwan I'll check more on this in a bit. I have a bit on my plate rn. Thank you for doing this though! Your contribution is appreciated.

heunghingwan:
After reading the specification, I found that it clearly shows that the server and client have independent `topicAliasMaximum` values:
3951 Topic Alias
3952 Decrease the size of the MQTT packet overhead by allowing the topic name to be abbreviated to a
3953 small integer. The Client and Server independently specify how many topic aliases they allow.
and `maximumPacketSize` values:
3969 Maximum Packet Size
3970 Allow the Client and Server to independently specify the maximum packet size they support. It is an
3971 error for the session partner to send a larger packet.

Those appendices summarize them well, glad I haven't miss them😁

heunghingwan:
fix by #1301 (#1300 for detail)

PR #1204, opened 2020-11-07
fix #1157

### LINKED ISSUE #1157: Bug: Resubscribe with persistent session only working correctly with protocolVersion 5
There is an issue with the automatic resubscription.

If a persistent session is used (```clean: false```) and the session does not exist after a reconnect (```connack.sessionPresent: false```, e.g. because the broker restarted), then NOTHING is resubscribed. 

I think this line here in the _resubscribe function is the problem:

```(this.options.clean || (this.options.protocolVersion === 5 && !connack.sessionPresent))```

Why is this check for protocolVersion 5 there? If I remove it then everything works.

### PR DESCRIPTION
**Pull request by heunghingwan**
`connack.sessionPresent` check should also check MQTT 3.1.1, because it also have `sessionPresent` property.
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718035

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1450,7 +1450,7 @@ MqttClient.prototype._resubscribe = function (connack) {
   debug('_resubscribe')
   var _resubscribeTopicsKeys = Object.keys(this._resubscribeTopics)
   if (!this._firstConnection &&
-      (this.options.clean || (this.options.protocolVersion === 5 && !connack.sessionPresent)) &&
+      (this.options.clean || (this.options.protocolVersion >= 4 && !connack.sessionPresent)) &&
       _resubscribeTopicsKeys.length > 0) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/client.js [hunk 1]
```diff
@@ -1450,7 +1450,7 @@ MqttClient.prototype._resubscribe = function (connack) {
   debug('_resubscribe')
   var _resubscribeTopicsKeys = Object.keys(this._resubscribeTopics)
   if (!this._firstConnection &&
-      (this.options.clean || (this.options.protocolVersion === 5 && !connack.sessionPresent)) &&
+      (this.options.clean || (this.options.protocolVersion >= 4 && !connack.sessionPresent)) &&
```
YoDaMa, line 1453: I don't see how this fixes the bug...

PR #1205, opened 2020-11-08
fix: prevent type error when missing args in `bin/pub`

kytart:
If either topic or message is not specified in the command, it throws "TypeError: Cannot read property 'toString' of undefined" because it's trying to call toString() on undefined. It's not really necessary to call toString() anyway.


YoDaMa:
can you add a test demonstrating this? If you need help writing it, just add a code snippet of the code that reproduces the bug.

kytart:
@YoDaMa sure, I'll write it

YoDaMa:
Let me know if you need assistance.

kytart:
I think I’ll be fine but I usually only have time on the weekends so hopefully this weekend I can do it

kytart:
Actually now that I'm looking at it, I'm not sure how to write the test. The code in `pub.js` is tighly coupled with the mqtt client code so that means I can't really write a unit test. I would have to write a test with an actual MQTT server running and test that the publish happened. But that's not really the point of this test. I'm just trying to test that the command line args were processed correctly.

PR #1209, opened 2020-11-11
fix #1175

### LINKED ISSUE #1175: Wrong condition in eventListenerSupport detection
Upgraded to **mqtt@4.2.1** and got error **Uncaught TypeError: socket.addEventListener is not a function** in Firefox

In file **dist/mqtt.js** on line **1930** there is a condition `const eventListenerSupport = (typeof socket.addEventListener === 'undefined')` which sets value of `eventListenerSupport` to `true` in case of method `addEventListener` is undefined. 

Changing the condition to `const eventListenerSupport = (typeof socket.addEventListener !== 'undefined')` solves the problem. 
Hope this helpes someone.

AB#8504348

nosovk:
use correct comparsion in eventListenerSupport check

### DIFF
--- a/lib/connect/ws.js
+++ b/lib/connect/ws.js
@@ -140,7 +140,7 @@ function browserStreamBuilder (client, opts) {
   }
   proxy.on('close', () => { socket.close() })
 
-  const eventListenerSupport = (typeof socket.addEventListener === 'undefined')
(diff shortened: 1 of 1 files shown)


robertsLando:
cc @mcollina

Issue #1215, opened 2020-11-23
yarn 2: Error: mqtt tried to access duplexify, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.


francisu:
The message below when running 4.2.3 in a yarn 2/pnp environment.

To fix this locally, add the following to your .yarnrc.yml file:

```
packageExtensions:
  mqtt@*:
    dependencies:
      duplexify: ^4.1.1
```

The mqtt library should be modified to add this dependency.

-----------------------

```
Error: mqtt tried to access duplexify, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: duplexify (via "duplexify")
Required by: mqtt@npm:4.2.3 (via /Users/francis/d/IdeaProjects/product/.yarn/cache/mqtt-npm-4.2.3-f99e660993-1c5e2a32ab.zip/node_modules/mqtt/lib/connect/)

Require stack:
- /Users/francis/d/IdeaProjects/product/.yarn/cache/mqtt-npm-4.2.3-f99e660993-1c5e2a32ab.zip/node_modules/mqtt/lib/connect/ws.js
- /Users/francis/d/IdeaProjects/product/.yarn/cache/mqtt-npm-4.2.3-f99e660993-1c5e2a32ab.zip/node_modules/mqtt/lib/connect/index.js
- /Users/francis/d/IdeaProjects/product/.yarn/cache/mqtt-npm-4.2.3-f99e660993-1c5e2a32ab.zip/node_modules/mqtt/mqtt.js
- /Users/francis/d/IdeaProjects/product/.yarn/cache/aws-iot-device-sdk-npm-2.2.6-53d2de1554-60ff3ccbb2.zip/node_modules/aws-iot-device-sdk/device/index.js
- /Users/francis/d/IdeaProjects/product/.yarn/cache/aws-iot-device-sdk-npm-2.2.6-53d2de1554-60ff3ccbb2.zip/node_modules/aws-iot-device-sdk/index.js
- /Users/francis/d/IdeaProjects/product/packages/universal/apolloClient/subscriptionHandshakeLink.ts
- /Users/francis/d/IdeaProjects/product/packages/universal/apolloClient/graphQlClient.ts
- /Users/francis/d/IdeaProjects/product/packages/universal/pipeline/pipelineManager.ts
- /Users/francis/d/IdeaProjects/product/packages/universal/dynamoAccess.ts
[...]

PR #1266, opened 2021-04-08
Add missing 'duplexify' dependency

simnalamburt:
'duplexify' is used by MQTT.js but not present in package.json

###### References
- https://github.com/mqttjs/MQTT.js/blob/37b12cb9/lib/connect/ali.js#L4
- https://github.com/mqttjs/MQTT.js/blob/37b12cb9/lib/connect/wx.js#L4
- https://github.com/mqttjs/MQTT.js/blob/37b12cb9/lib/connect/ws.js#L5
- https://github.com/mqttjs/MQTT.js/issues/1215

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "commist": "^1.0.0",
     "concat-stream": "^2.0.0",
     "debug": "^4.1.1",
+    "duplexify": "^4.1.1",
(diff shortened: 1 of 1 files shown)


robertsLando:
cc @mcollina

PR #1298, opened 2021-07-01
feat: add `quic` support

agilan2001:
Added support for QUIC protocol through the core QUIC module of NodeJS (v15 - pre)

[NodeJS QUIC Documentation](https://nodejs.org/dist/v15.7.0/docs/api/quic.html)


agilan2001:
Added support for QUIC protocol through the core QUIC module of NodeJS (v15 - pre)

[NodeJS QUIC Documentation](https://nodejs.org/dist/v15.7.0/docs/api/quic.html)

YoDaMa:
@agilan2001 also please advise how you want to go about supporting this. Since it is an experimental feature in v15, we cannot just say "we support this feature". Is there a way to mark a feature as experimental in the API?

ysfscream:
@agilan2001 Hi, I am so happy and excited to see your PR!

Because the EMQX already supports MQTT over QUIC, see: https://www.emqx.com/en/blog/getting-started-with-mqtt-over-quic-from-scratch
We want to support connections on the client side as well, but Node.js doesn't seem to support it anymore, so I'm very excited to see your PR, so can you help me how to implement and use it in MQTT.js? I'm currently using it with some errors: it doesn't seem to introduce the module related to QUIC.

image

agilan2001:
@ysfscream QUIC support was present in NodeJS only in v15.x.
This would work only in the manually built NodeJS-v15.x binary with `--experimental-quic` flag enabled.

[QUIC Support in the core package could be tracked in the open PR : https://github.com/nodejs/node/pull/44325]

ysfscream:
@agilan2001 Thanks very much. I see a PR https://github.com/nodejs/node/pull/44325 on the Node.js. Hope to support it quickly and well.

robertsLando:
@agilan2001 Can you add a test for this?

robertsLando:

Issue #1306, opened 2021-07-20
A vulnerability introduced in your package


paimon0715:
Hi ,@YoDaMa @redboltz , I’d like to report a vulnerability in your package:
### Issue Description
   A vulnerability [**CVE-2021-32640**](https://snyk.io/vuln/SNYK-JS-WS-1296835) in package **ws (<5.2.3,>=6.0.0 <6.2.2,>=7.0.0 <7.4.6)** is transitively referenced by **_mqtt@2.18.8_**. We noticed that such vulnerability has been removed since _**mqtt@4.2.0**_.
  
   However, mqtt's popular previous version **_mqtt@2.18.8_**  ([49,659 downloads per week](https://www.npmjs.com/package/mqtt/v/2.18.8?activeTab=versions)) is still transitively referenced by a large amount of latest versions of active and popular downstream projects (about **871 downstream projects**, e.g., @types/aws-iot-device-sdk 2.2.3, iobroker.mqtt 2.4.0, @things-factory/shell 3.4.5, @things-factory/board-ui 3.4.5, @things-factory/auth-base 3.4.5, @ftw-cloud/core@1.5.162, @things-factory/apptool-base@3.4.5, @things-factory/attachment-base@3.4.5, etc.). As such, issue [**CVE-2021-32640**](https://snyk.io/vuln/SNYK-JS-WS-1296835)  can be propagated into these downstream projects and expose security threats to them.
   
   These projects cannot easily upgrade **mqtt** from version **_2.18.8_** to _**4.2.\***_ . For instance, **_mqtt@2.18.8_** is introduced into the above projects via the following package dependency paths:
(1)` @ftw-cloud/core@1.5.162 ➔ graphql-mqtt-subscriptions@1.2.0 ➔ mqtt@2.18.8 ➔ websocket-stream@5.5.2 ➔ ws@3.3.3`
(2) `@things-factory/apptool-base@3.4.5 ➔ @things-factory/shell@3.4.5 ➔ graphql-mqtt-subscriptions@1.2.0 ➔ mqtt@2.18.8 ➔ websocket-stream@5.5.2 ➔ ws@3.3.3`
[...]

PR #1307, opened 2021-07-20
fix: security vulnerability in ws stream

YoDaMa:
This is a patch update to remove a security vulnerability from a old version of the package. As referred by #1306


YoDaMa:
cc @paimon0715

Issue #1233, opened 2021-01-15
Browser detection using global __webpack_require__ is misleading


ArminEberle:
Hi there,

we're using your library from a node standalone application.
It works fine when running our .js files as original.

Anyway when we build the node application using webpack, it seems to me that this has an unwanted impact on the browser detection in lib/connect/index.js and lib/connect/ws.js, effectively leading to connection problems. 

We want to switch to a webpack build in order reduce our container sizes in kubernetes.

This is where it fails then:
```
ReferenceError: WebSocket is not defined
    at createBrowserWebSocket (/.build-tmp/webpack-dist/index.js:1993:16)
    at Object.browserStreamBuilder (/.build-tmp/webpack-dist/index.js:2019:16)
    at MqttClient.wrapper [as streamBuilder] (/.build-tmp/webpack-dist/index.js:2633:36)
    at MqttClient._setupStream (/.build-tmp/webpack-dist/index.js:395:22)
    at new MqttClient (/.build-tmp/webpack-dist/index.js:374:8)
    at Object.connect (/.build-tmp/webpack-dist/index.js:2635:16)
```

The line numbers are not correct of course in this scenario. 

I'm just asking if you already know of this issue and could give me a hint, if not I can create a reproducible demo of the problem or bugfix fork if you want me to.

Best Regards,

Armin

AB#9138553

sunkainj:
![图片](https://user-images.githubusercontent.com/19506712/110569072-f4f57c00-818e-11eb-9f6d-61cffa55a82f.png)
http://localhost:3000/node_modules/readable-stream/lib/_stream_readable.js
var OurUint8Array = global.Uint8Array || function () {};
It seems to be one problem

lopengil:
Hi, is there any update with this issue? I'm having the same problem, when running the js files it works fine but when running the webpack bundle it fails here:

```
ReferenceError: WebSocket is not defined
    at createBrowserWebSocket (...\dist\main.js:1940:16)
    at Object.browserStreamBuilder (...\dist\main.js:1966:16)
    at MqttClient.wrapper [as streamBuilder] (...\dist\main.js:2567:36)
    at MqttClient._setupStream (...\dist\main.js:343:22)
    at new MqttClient (...\dist\main.js:322:8)
    at Function.connect (...\dist\main.js:2569:16)
    at MosquittoClient.connect (...\dist\main.js:17090:42)
    at ...\dist\main.js:16934:25
    at Generator.next (<anonymous>)
[...]

PR #1320, opened 2021-08-24
Fix "IS_BROWSER" check for NodeJS Webpack built application

Symsystem:
Suggestion to fix the issue described [here](https://github.com/mqttjs/MQTT.js/issues/1233).


vishnureddy17:
For context, this was closed because #1571 addresses it.

robertsLando:
Them are quite different by the way as this was removing the webpack check but that would break too much users... the alternative to this would be to add support for a way to force set IS_BROWSER to false (maybe with an env var). But I leave this to the user that opened this PR if interested

PR #1332, opened 2021-09-28
Add support for ALPN TLS extension

marcotuna:
This Pull Request adds support for ALPN.  
With this change it is now possible to use AWS MQTT Broker on Port 443 with Custom Authorization.  
More information at: https://docs.aws.amazon.com/iot/latest/developerguide/protocols.html  

This change is not AWS specific, there's more use cases behind the support of this feature.  

For more references have a look at the NodeJS types repository: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L446

Example:  
```
mqtt.connect("AWS_ACCOUNT_ID-ats.iot.eu-west-1.amazonaws.com", {
  ...
  port: 443,
  ALPNProtocols = ['mqtt']
  ...
})
```

### DIFF
--- a/types/lib/client-options.d.ts
+++ b/types/lib/client-options.d.ts
@@ -131,6 +131,10 @@ export interface ISecureClientOptions {
    */
   ca?: string | string[] | Buffer | Buffer[]
   rejectUnauthorized?: boolean
+  /**
+   * optional alpn's
+   */
+  ALPNProtocols?: string[] | Buffer[] | Uint8Array[] | Buffer | Uint8Array;
 }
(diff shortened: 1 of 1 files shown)


marcotuna:
Hey @YoDaMa. Can you have a look on this?
Do you need more clarifications about it?
The current use case that this is attempting to solve is to support AWS iot without any specific AWS code. The current implementation is generic and might allow different integrations with MQTT Brokers using Reverse Proxies using TLS APLN capabilities.

PR #1352, opened 2021-11-04
Rework examples to be a bit more specific

cruxicheiros:
Firstly: I love this module, it's really helpful and I use it all the time.

I was using this module with TypeScript for the first time today and found the docs a bit confusing. I made these changes because I would have liked to have had this information up front and labelled more clearly:
* How to import the contents of MQTT.js in more than one import style
* A TypeScript example showing off one of the types in MQTT.js

I also added the WeChat and AliPay protocols to the list of the supported protocols because I thought it was a bit weird that they were listed in their own section but not there, and restructured the section to only contain one example of the specific usage to make it clearer what's different there.

Thank you! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ in JavaScript for node.js and the browser.
 * [Command Line Tools](#cli)
 * [API](#api)
 * [Browser](#browser)
-* [Weapp and Ali support](#weapp-alipay)
 * [About QoS](#qos)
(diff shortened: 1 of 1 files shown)


YoDaMa (approved):
lgtm. Thanks for your contribution. Be on the lookout in the coming months for the next generation release of this package. We will be moving to Promise-based API and Typescript, and looking for good user feedback.

PR #1369, opened 2021-12-14
fix(type): add properties type for IClientSubscribeOptions

ysfscream:
Hi, devs! @YoDaMa When I set the MQTT 5.0 properties on subscribe function, I found no properties type defined.

image

Hopefully, this PR will fix this type of error.

Because when I checked the documentation, it supports this configuration.

image


YoDaMa (approved):
lgtm

PR #1400, opened 2022-01-07
add dummy jest test

vishnureddy17:
run `npm install`, then `npm test` and it should run the Jest dummy test.


package.json
```diff
@@ -101,16 +88,15 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.0",
     "global": "^4.4.0",
+    "jest": "^27.4.7",
     "mkdirp": "^1.0.4",
-    "mocha": "^9.1.3",
```
vishnureddy17, line 105: There's a bunch of dependencies like `mocha`, `chai`, and `sinon` that we wouldn't be using if we go with Jest. What's the reason why these are here? Is it just because it was there from the vCurrent `package.json`?

YoDaMa: yes they can be removed if unnecessary.

YoDaMa (approved):
lgtm

PR #1401, opened 2022-01-14
fix drain leak

BertKleewein:
Fix a regression introduced by #1301.  The `cb = cb || nop` line in `_sendPacket` was adding a callback where none existed before.  This changed the behavior of `sendPacket` (without the underscore) by adding a listener to the `drain` event, and you can quickly overflow the max # of listeners if you're trying to publish faster than the stream will allow.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -175,7 +175,7 @@ function sendPacket (client, packet, cb) {
   debug('sendPacket :: writing to stream')
   const result = mqttPacket.writeToStream(packet, client.stream, client.options)
   debug('sendPacket :: writeToStream result %s', result)
-  if (!result && cb) {
+  if (!result && cb && cb !== nop) {
     debug('sendPacket :: handle events on `drain` once through callback.')
     client.stream.once('drain', cb)
   } else if (cb) {
@@ -190,6 +190,8 @@ function flush (queue) {
     Object.keys(queue).forEach(function (messageId) {
       if (typeof queue[messageId].cb === 'function') {
         queue[messageId].cb(new Error('Connection closed'))
+        // This is suspicious.  Why do we only delete this if we have a callbck?
+        // If this is by-design, then adding no as callback would cause this to get deleted unintentionally.
         delete queue[messageId]
       }
     })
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -175,7 +175,7 @@ function sendPacket (client, packet, cb) {
   debug('sendPacket :: writing to stream')
   const result = mqttPacket.writeToStream(packet, client.stream, client.options)
   debug('sendPacket :: writeToStream result %s', result)
-  if (!result && cb) {
+  if (!result && cb && cb !== nop) {
```
BertKleewein, line 178: This code (the way `sendPacket` handles `writeToStream` return values) is built on a number of false assumptions:
1. It assumes that `writeToStream` will only return `false` if the packet was queued and is waiting to send. This is wrong. `writeToStream` can also fail if packet validation fails (in `mqtt-packet/writeToStream.js`) If this happens, `sendPacket` will wait for the stream to drain, then call the callback with no error.
2. If the stream is full, and `writeToStream` returns false, it waits for all packets to send before calling any of their callbacks.  Once all packets are sent, the `drain` event fires, and all callbacks get called at once.  If the sending pattern prevents the stream from draining completely, then the callbacks will never be called (or they might be called long after the packet was actually sent).
3, This assumes that there can only be 1000 packets waiting to send.  After that, the maxListeners on the `drain` event overflows. I'm not sure how well the library handles this error (Does it raise an exception? If so, who catches it?)

A better fix would be to pass the callback into the stream's `write` method so it gets called when the packet is actually sent.  At least this is how `TLSSocket.write` behaves.  I don't know if this applies to all of the stream implementations that mqtt.js supports.  This would require an update to the `mqtt-packet` library to support an additional `callback` parameter to the `writeToStream` function.

BertKleewein: Plus, `packetsend` is emitted _before_ the packet is actually sent (if the packet is ever actually sent).  Is that what we want?

vishnureddy17 (approved):

vishnureddy17:
@redboltz I'm not able to request a review from you for some reason, but pinging you in case you have any feedback here.

lib/client.js
vishnureddy17: Opened an issue on this in #1403

YoDaMa (approved):
lgtm

PR #1404, opened 2022-01-21
pruning

YoDaMa:


.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
```
vishnureddy17, line 1: Do we want to check in vscode files?

package-lock.json
```diff
@@ -0,0 +1,15738 @@
+{
```
vishnureddy17: Do we want to check in `package-lock.json`? I don't think it is being done in the main branch. It's pretty annoying but I also get that its useful for consistency. If I were to vote on it, I'd vote to not check it in.

package.json
```diff
@@ -27,7 +27,22 @@
   "scripts": {
     "build": "tsc"
   },
+  "dependencies": {
+    "duplexify": "^4.1.2",
+    "end-of-stream": "^1.4.4",
+    "mqtt-packet": "^7.1.1",
+    "number-allocator": "^1.0.9",
```
vishnureddy17, line 34: Do we really need `number-allocator`? I feel like we can implement message ID allocation without taking a dependency. This rewrite is a good opportunity to be really intentional about dependencies.

vishnureddy17 (commented):
Left some comments that I think we should discuss, but if you feel they are out of scope for this PR go ahead and merge.

package-lock.json
YoDaMa: ok. my reason for package-lock is it is a really nice feature in Node for CI/CD, since it creates reproducible environments for testing. It adds some "bloat" to PRs but the gain of consistent dependencies for testing is worth it IMHO.

PR #1405, opened 2022-01-24
remove process.on, use strict, and js files. Update typescript nightly->beta

vishnureddy17:


src/client.ts
```diff
@@ -1,5 +1,3 @@
-'use strict'
```
vishnureddy17, line 1: esm is strict by default so no need for this

src/connectionFactory/interfaces/webSocketOptions.js
```diff
@@ -1,2 +0,0 @@
-"use strict";
-exports.__esModule = true;
```
vishnureddy17, line 2: What is this file for? I assume we don't want any JS in our `src` directory

src/index.ts
```diff
@@ -11,18 +11,9 @@ import { UniqueMessageIdProvider } from './uniqueMessageIdProvider.js'
...
 import { URL } from 'url'
 
-process.on("unhandledRejection", (err) => {
-  throw err;
-})
-
-process.on("uncaughtException", (err) => {
-  handle(err);
-})
-
```
vishnureddy17, line 25: We should not have this. If a user imports our package, then every single uncaught exception or unhandled rejection that happens on that Node process will be altered by us.

YoDaMa (approved):
lgtm

PR #1413, opened 2022-02-04
refactor connect

vishnureddy17:
Sends CONNECT packet seemingly without errors, but times out receiving CONNACK. Need to investigate more.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,4 +1,5 @@
 import { IConnackPacket, IConnectPacket, Packet, parser as mqttParser, Parser as MqttParser, writeToStream } from 'mqtt-packet'
+import { write } from './write.js'
 import { ConnectOptions } from './interfaces/connectOptions.js'
 import { Duplex } from 'stream'
 import { EventEmitter } from 'node:events'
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -203,27 +203,28 @@ export class MqttClient extends EventEmitter {
...
-        rej('CONNECTION TIMEOUT')
-      }
-      const connackTimer = setTimeout(connectionTimeout, this._options.connectTimeout)
-      this.on('connack', (connackPacket) => {
-        this.removeListener('connack', res)
-        clearTimeout(connackTimer)
-        res(connackPacket)
+  private _awaitConnack(): Promise<IConnackPacket> {
+    return new Promise((resolve, reject) => {
+      const connackTimeout = setTimeout(
```
YoDaMa, line 208: maybe we should create a `sleep` method that's a promisified `setTimeout` so that our code looks less callbacky here.

YoDaMa (approved):
lgtm

src/client.ts
vishnureddy17: Yeah, I think we're going to have timeouts in other places too so good call

PR #1416, opened 2022-02-08
Fix the typings for the connect function to have the proper overloads

erichardson-lee:
This makes it so that it makes it more readily apparent that you can omit the URL string, and provide the options as the first parameter instead.
This change purely effects the typing in order to make it better represent the underlying code, so it shouldn't cause any breaking changes.


YoDaMa (approved):
lgtm

Issue #1412, opened 2022-02-04
ReferenceError: Buffer is not defined ReactJS


yaminagile:
I have react app which is created by create-react-app and follow the rules that provides here
https://github.com/mqttjs/MQTT.js#react

But when I start my app I am getting following error 

```
constants.js:46 Uncaught ReferenceError: Buffer is not defined
    at Object../node_modules/mqtt/node_modules/mqtt-packet/constants.js (constants.js:46:1)
    at Object.options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:61:1)
    at Object../node_modules/mqtt/node_modules/mqtt-packet/parser.js (parser.js:4:1)
    at Object.options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:61:1)
    at Object../node_modules/mqtt/node_modules/mqtt-packet/mqtt.js (mqtt.js:1:1)
```

I have look around similar to this issue but it didn't help me.

I am using react version 17.0.2 and react-scripts version 5.0.0

I am gone through this issue
https://github.com/mqttjs/MQTT.js/issues/1294

But it didn't help me because I don't have any webpack config file on my react app.

flyfishzy:
I have the same probelm as you, and have submitted a PR.

YoDaMa:
@flyfishzy I see the PR is merged in mqtt-packet. is the problem in our dependency version in mqtt.js?

flyfishzy:
@YoDaMa Yes, we need to upgrade the mqtt-packet version. But there are other similar problems, I'm trying to continue to fix them.

Pavanct:
Is there a workaround for this for time being?

yaminagile:
This works for me
 https://www.npmjs.com/package/paho-mqtt 
 which is compatible with all the latest version of react and react-scripts 
You don't have to downgrade react, react-script or webpack for that.

ambienthack:
I was able to build mqtt for React following these steps:

### inside of `node_modules/mqtt` create `webpack.config.js`:
```
const webpack = require('webpack')

module.exports = {

    entry: "./lib/connect/index.js",

    //mode: "development",

    output: {
        library:  {
            type: "commonjs2"
        },
        filename: "mqtt.browser.js"
    },
    
    plugins: [
        // fix "process is not defined" error:
        // (do "npm install process" before running the build)
[...]

PR #1420, opened 2022-02-14
fix: buffer is not defined in browser

flyfishzy:
fix issue https://github.com/mqttjs/MQTT.js/issues/1412
The Node.js doc recommends to explicitly import or require 'buffer':
> While the Buffer class is available within the global scope, it is still recommended to explicitly reference it via an import or require statement.
https://nodejs.org/api/buffer.html

Similar as https://github.com/mqttjs/mqtt-packet/pull/128

### DIFF
--- a/lib/connect/ali.js
+++ b/lib/connect/ali.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { Buffer } = require('buffer')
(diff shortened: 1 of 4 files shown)


YoDaMa (approved):
lgtm

PR #1429, opened 2022-02-24
add packet ID allocator

vishnureddy17:
We need packet IDs for subscribe, so went ahead and implemented this. I tried to explain the code as best I could in comments, please let me know if something is unclear. 

Needs to be tested.

### DIFF
--- a/src/utils/packetIdAllocator.ts
+++ b/src/utils/packetIdAllocator.ts
@@ -1,0 +1,75 @@
+/* Keeps track of free and used packet ids by storing available IDs as intervals */
+export class PacketIdAllocator {
+  /**
+   * Intervals are stored in reverse order since push/pop from end of a list is
+   * faster than shift/unshift from beginning.
+   */
+  private _intervals: [number, number][] = [[65535, 1]];
+
+  async allocate(): Promise<number> {
+    if (this._intervals.length === 0) {
+      /* TODO: return promise that resolves when a new packet ID is available */
+      throw new Error('No more packet ids available');
+    }
+
+    const lastInterval = this._intervals[this._intervals.length - 1] as [number, number];
+    const [lastIntervalMax, lastIntervalMin] = lastInterval;
+    if (lastIntervalMax === lastIntervalMin) {
+      /* We just used the last packet ID in the interval, so remove it. */
+      this._intervals.pop();
+    } else {
+      /* We are using the lowest value in the interval, so increment the min */
+      ++lastInterval[1];
+    }
+    return lastIntervalMin;
+  }
+
+  release(id: number): void {
+    if (this._intervals.length === 0) {
+      this._intervals.push([id, id]);
+      return;
+    }
+    let lastInterval = this._intervals[this._intervals.length - 1] as [number, number];
+    let lastIntervalMin = lastInterval[1];
+    if (id < lastIntervalMin - 1) {
+      this._intervals.push([id, id]);
+      return;
+    }
+    if (id === lastIntervalMin - 1) {
+      --lastInterval[1]
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


src/utils/packetIdAllocator.ts
```diff
@@ -0,0 +1,79 @@
+/* Keeps track of free and used packet ids by storing available IDs as intervals */
```
YoDaMa, line 1: nit: IDs vs ids. I favor IDs

src/utils/packetIdAllocator.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  private _intervals: [number, number][] = [[65535, 1]];
+
+  async allocate(): Promise<number> {
+    if (this._intervals.length === 0) {
+      /* TODO: return promise that resolves when a new packet ID is available */
+      throw new Error('No more packet IDs available');
+    }
+
+    const lastInterval = this._intervals[this._intervals.length - 1] as [number, number];
```
YoDaMa, line 15: is there a reason you are typecasting here? Why not define the type of lastInterval as [number, number]

src/utils/packetIdAllocator.ts
YoDaMa: also it is not clear what an "interval" is. Can you add better documentation? 

At first glance this looks like a way to make sure that we have unique IDs for Messages. I like the initiative on creating this but do we have performance metrics on it? Is it better than something out there that is already implemented that accomplishes the same goal?

test/test_packet_id_allocator.js
```diff
@@ -0,0 +1,181 @@
...
+test('should end up with one interval after releasing all IDs in random order', async (t) => {
+  t.plan(1)
+  const allocator = new PacketIdAllocator();
+
+  /* Allocate all IDs */
+  while (await allocator.allocate() < maxPacketId);
+
+  /* [1, 2, 3, ..., 65535] */
+  const idsToRelease = Array.from({ length: maxPacketId }, (_, i) => i + 1);
+  /* Shuffle array using Fisher-Yates */
```
YoDaMa, line 27: why shuffle the array?

YoDaMa:
So this is supposed to be an implementation of Unique Packet IDs. Are you using an algo you were aware of that optimizes performance with the intervals? Because reading this code... it seems unnecessarily complicated. I could be wrong of course. Regardless, at first glance I have no idea how it works which is not a great sign.

YoDaMa (changes requested):
at the least, add documentation and better comments explaining what this is doing and why it is necessary.

test/test_packet_id_allocator.js
vishnureddy17: Testing that regardless of the insertion order the final result is correct

src/utils/packetIdAllocator.ts
vishnureddy17: Typescript makes this `[number, number] | undefined` because it's not aware that list is non-empty at the point that we check it. We have a check earlier for the list being empty earlier so it's safe to assume that `list[list.length - 1]` will always be not undefined

vishnureddy17:
Did some basic performance testing to see if this approach makes sense. Methodology: Use up all packet ids. Then time the insertion of all the packet ids in random order.

Results (average over 100 iterations):
Using a Set: 4.47ms
Using the implementation in this PR: 1709.79ms
Using [`NumberAllocator`](https://github.com/redboltz/number-allocator) written by @redboltz : 60.01ms

Seems like `Set` is the fastest, but it uses more memory as there are 65535 numbers stored in the set if all the IDs are free. Might not be a big deal in the grand scheme of things. 

For now, I'll close this PR and use the `Set` approach and we can address it later if it becomes an issue.

PR #1432, opened 2022-02-27
docs: Update TypeScript docs to include installing @types/ws

### LINKED ISSUE #1408: WebSocket types aren't pulled in by client libraries
Hello!

I have TypeScript project that depends on `mqtt`. If I enable the `strict`  compiler option in my project, it refuses to build because of a implicit-any match in `mqtt`.

This is because [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck) is set to `false` by default which means tsc will try to type-check `mqtt` as well. The specific line it fails at is importing a type from `ws` in [types/lib/client-options.d.ts](https://github.com/mqttjs/MQTT.js/blob/master/types/lib/client-options.d.ts#L3):
```
node_modules/mqtt/types/lib/client-options.d.ts:3:31 - error TS7016: Could not find a declaration file for module 'ws'. '/project/node_modules/ws/index.js' implicitly has an 'any' type.
  Try `npm install @types/ws` if it exists or add a new declaration (.d.ts) file containing `declare module 'ws';`

3 import { ClientOptions } from 'ws'
                                ~~~~


Found 1 error.
```

This happens because `@types/ws` is declared as a devDependency so it isn't automatically pulled in to my project. To fix this, `mqtt` would either need to move `@types/ws` to dependencies or at least document that I should install it myself (similar to what's done with `@types/node` [here](https://github.com/mqttjs/MQTT.js/blob/main/README.md?plain=1#L799-L800)).

StenAL:
Also remove an unused import in types/lib/client.d.ts.

Fixes #1408

### DIFF
--- a/README.md
+++ b/README.md
@@ -789,13 +789,16 @@ About data consumption, obviously, QoS 2 > QoS 1 > QoS 0, if that's a concern to
 This repo bundles TypeScript definition files for use in TypeScript projects and to support tools that can read `.d.ts` files.
 
 ### Pre-requisites
-Before you can begin using these TypeScript definitions with your project, you need to make sure your project meets a few of these requirements:
+Before you can begin using these TypeScript definitions with your project, you need to make sure your project meets these requirements:
  * TypeScript >= 2.1
  * Set tsconfig.json: `{"compilerOptions" : {"moduleResolution" : "node"}, ...}`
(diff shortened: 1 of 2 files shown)


YoDaMa (approved):
lgtm

StenAL:
Hey @YoDaMa, can this be merged?

PR #1436, opened 2022-03-02
fix disconnect

YoDaMa:
This changes some of the listener destruction and emission to improve the reliability of the disconnect.


YoDaMa:
need to fix problems with merge.

vishnureddy17:
Is this ready for review? @YoDaMa

YoDaMa:
yes @vishnureddy17

src/client.ts
```diff
@@ -196,19 +197,22 @@ export class MqttClient extends EventEmitter {
 
   private async _destroyClient(force?: boolean) {
     this._clientLogger.trace(`destroying client...`);
-    this.conn.removeAllListeners('error')
+    this.conn.removeAllListeners('error');
+    this.conn.removeAllListeners('close');
+    this.conn.on('close', () => {});
```
vishnureddy17, line 202: Why is this needed? I can see why you wanna do this with the `error` event because Node complains when `error` is emitted and there are no listeners. But that's not the case for `close`.

vishnureddy17 (approved):
Approving and merging despite the one comment I had so I can work on top of this. I'll make note of it so we can discuss on Monday. @YoDaMa

PR #1443, opened 2022-03-10
fix: fix regression from #1401 and allow CI test failures to break gitthub workflow

BertKleewein:


lib/client.js
```diff
@@ -1583,7 +1583,7 @@ MqttClient.prototype._handleAck = function (packet) {
   const that = this
   let err
 
-  if (!cb || cb === nop) {
```
vishnureddy17, line 1586: Can we get an explanation as to what went wrong here?

lib/client.js
```diff
@@ -1583,7 +1583,18 @@ MqttClient.prototype._handleAck = function (packet) {
...
+  // Checking `!cb` happens to work, but it's not technically "correct".
+  //
+  // Why? This code assumes that "no callback" is the same as that "we're not
+  // waiting for responses" (puback, pubrec, pubcomp, suback, or unsuback).
+  //
+  // It would be better to check `if (!this.outgoing[messageId])` here, but
+  // there's no reason to change it and risk (another) regression.
+  //
+  // The only reason this code works is becaues code in MqttClient.publish,
+  // MqttClinet.subsribe, and MqttClient.unsubscribe ensures that we will
```
vishnureddy17, line 1595: nit: typo
```suggestion
  // MqttClient.subscribe, and MqttClient.unsubscribe ensures that we will
```

vishnureddy17 (approved):
looks good, small nit

Apollon77:
Thank you very much! Any ETA when we can expoect a new release?

Apollon77:
Thank you for the info!

PR #1444, opened 2022-03-10
chore: fix test, lint tests, and run prettier everywhere

BertKleewein:


vishnureddy17 (approved):
Thanks, @BertKleewein! LGTM. 

I think @YoDaMa may have had a reason for having the ESLint rules the way they were.

PR #1449, opened 2022-03-14
spike: possible sequencer implementation for vNext-dev

BertKleewein:


src/sequencer.ts
```diff
@@ -0,0 +1,305 @@
...
+export interface Packet {
+  messageId: number | undefined;
+}
+
+// It's called SendPacketFuncton, but it accepts a Message. Maybe we need to rethink this.
+// The more I think about this, the more I think it should accept a `Packet` object, and it
+// should use `Packet.cmd` instead of accepting a separate `PacketType` parameter.
+export type SendPacketFunction = (packetType: PacketType, message: Message) => void;
+
+type InFlightMessageMap = { [key: number]: SequenceMachine };
```
vishnureddy17, line 22: nit: prefer to use `Map` over a regular object

src/sequencer.ts
```diff
@@ -0,0 +1,305 @@
...
+        }
+        this.inFlightMessages[message.messageId as number] = sequenceMachine;
+      // SUBSCRIBE also goes into the inFlightMesages map. CONNECT maybe goes somewhere else?
+    }
+
+    sequenceMachine.start();
+  }
+
+  // TODO: Is there an easier way to Promisify this?
+  runSequence(sequenceType: SequenceType, message: Message, done?: DoneFunction) {
```
vishnureddy17, line 63: could we just get rid of the callback? Supporting both promise and callback patterns can be painful long-term

src/sequencer.ts
```diff
@@ -0,0 +1,305 @@
...
+    packet;
+    logger.info('blah');
+  }
+
+  cancel() {
+    logger.info('blah');
+  }
+}
+
+// TODO: do we want a state for Cancelled?
```
vishnureddy17, line 155: I think we do

src/sequencer.ts
```diff
@@ -0,0 +1,305 @@
...
+      logger.info('blah');
+    }
+  }
+}
+
+abstract class SequenceMachine {
+  message: Message;
+  sendPacketFunction: SendPacketFunction;
+  done: DoneFunction;
+  timeout: NodeJS.Timeout | number = 0;
```
vishnureddy17, line 95: Once this is merged in I'll add a TODO comment about investigating efficient timer usage so every state machine doesn't need it's own timer

vishnureddy17 (approved):
This is excellent. Got some nits. I can address them myself once merged.

PR #1465, opened 2022-03-30
cleanup topic validation

YoDaMa:


src/util/validateTopic.ts
```diff
@@ -1,39 +1,61 @@
...
+    if (levels[i] === '#') {
+      // Validate MLWC at end of topic filter.
+      return i === levels.length - 1;
+    }
+
+    // Level must not contain more than one MLWC or SLWC character.
+    if ((levels[i] as string).indexOf('+') !== -1 || (levels[i] as string).indexOf('#') !== -1) {
+      return false;
     }
   }
```
vishnureddy17, line 42: Haven't verified if this builds but this makes things a bit more readable IMO
```suggestion
  for (const [i, level] of levels.entries()) {
    // If SLWC, MUST occupy entire level.
    if (level === '+') {
      continue;
    }

    if (level === '#') {
      // Validate MLWC at end of topic filter.
      return i === levels.length - 1;
    }

    // Level must not contain more than one MLWC or SLWC character.
    if (level.includes('+') || levels.includes('#')) {
      return false;
    }
  }
```

vishnureddy17 (approved):
Readability nit, otherwise LGTM

PR #1486, opened 2022-05-20
integrate sequencer connect with upper level client

vishnureddy17:


src/client.ts
```diff
@@ -30,56 +24,39 @@ function eosPromisified(stream: NodeJS.ReadableStream | NodeJS.WritableStream):
...
     // assume the options have been validated before instantiating the client.
-    this.connecting = false;
-    this.connected = false;
-    this.errored = false;
-    this.disconnecting = false;
-    this._inflightPackets = new Map();
-
-    // Using this method to clean up the constructor to do options handling
-    logger.trace(`populating internal client options object...`);
-    this._options = {
```
YoDaMa, line 61: did you remove the default options?

src/client.ts
```diff
@@ -30,56 +24,39 @@ function eosPromisified(stream: NodeJS.ReadableStream | NodeJS.WritableStream):
...
+    this._conn.setMaxListeners(1000);
 
     this._incomingPacketParser = mqttParser(this._options);
 
-    // Handle incoming packets this are parsed
-    // NOTE: This is only handling incoming packets from the
-    // readable stream of the conn stream.
-    // we need to make sure that the function called on 'packet' is bound to the context of 'MQTTClient'
-    this._incomingPacketParser.on('packet', this.handleIncomingPacket.bind(this));
+    this._incomingPacketParser.on('packet', this._packetSequencer.handleIncomingPacket.bind(this._packetSequencer));
```
YoDaMa, line 59: maybe dumb question: is handleIncomingPacket bound to _packetSequencer because the context for packets should be _packetSequencer's context?

src/client.ts
vishnureddy17: I moved it up to connect in index.ts because I feel like inserting default options and validating options should be done together

src/client.ts
vishnureddy17: It's because the method `MqttPacketSequencer.handleIncomingPacket()` should have the context of `this._packetSequencer`

PR #1492, opened 2022-06-08
fix: `_storeProcessing` staying true after outStore got emptied

### LINKED ISSUE #1116: New messages not being published after store messages are processed - possible solution
Hi there!

We have an electron app that only publishes QoS 2 messages and I'm using https://github.com/mcollina/mqtt-level-store as persistent store.
Our users reported that a lot of messages are not being published. The internet connection in the warehouse we run the app is not good, so we lose connection quite often. After some investigation, this is what I've found:

- whenever we lose connection, the messages are stored as expected, so after reconnection they are sent;
- however, after the stored messages are sent, all the newer messages are stored immediately. They are only sent after someone restarts the app;
- and then again, after the previously stored messages are sent, the newer messages are immediately stored again.

I've enabled debugging logs and I noticed the log `_storeProcessing enabled`, so I checked the code and I noticed that the flag `_storeProcessing` is still `true` after the store finished processing in both cases (https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L517-L535).

So I've forked this repo, added more logging messages to understand what's happening and this is what I've come up with:

- `_storeProcessing = true` is set in the beginning of `storeDeliver` (https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L1507-L1514);
- if there’s no packet, a listener is set for when it’s readable again, but `_storeProcessing` is never set to `false` again (https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L1518-L1522);
- then, when a new message is published, there's the aforementioned check and the message is stored immediately.

Now, I'm not sure if this is the expected behavior. As a quickfix to our problem, I've added a call to `clearStoreProcessing()` inside the `!packet` check to reset the value of `_storeProcessing`.

https://github.com/gabrielfnogueira/MQTTjs/blob/reset-store-processing/lib/client.js#L1518-L1523

I've been using my fork in the application and in my tests everything's working as (we) expected.

If this is indeed a bug, I'm more than happy to open a Pull Request with my change. If not, I'd like to hear your feedback on it 😄

Genoil:
This PR makes sure that new messages get published after a local store that accumulated some messages while the broker was offline, reconnects. In storeDeliver, this._storeProcessing got set to true before the check if there are actually stored packets left to send. So if there aren't any, this._storeProcessing will still be true on the next publish(), resulting in packets accumulating in this._storeProcessingQueue, but their invokes are never called, because Readable Event 'end' has already happened for the emptied Store and no new messages will be put into the Store as well. 

Fixes #1116

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1806,7 +1806,6 @@ MqttClient.prototype._onConnect = function (packet) {
       if (!outStore) {
         return
       }
-      that._storeProcessing = true
 
       const packet = outStore.read(1)
 
@@ -1817,6 +1816,8 @@ MqttClient.prototype._onConnect = function (packet) {
         outStore.once('readable', storeDeliver)
         return
       }
+
+      that._storeProcessing = true
 
       // Skip already processed store packets
       if (that._packetIdsDuringStoreProcessing[packet.messageId]) {


robertsLando:
@Genoil Could you add a unit test for this please? I'm ok to merge it as soon as a test covers the fix

robertsLando:
@gabrielfnogueira could you confirm this fixes your issue?

gabrielfnogueira:
> @gabrielfnogueira could you confirm this fixes your issue?

yes, this does fixes my issue!

Issue #1474, opened 2022-04-25
setImmediate polyfill is not implementing node 16 standard of setImmediate


filipomar:
While working with the MQTT client on the browser, I noticed callbacks (sent through setImmediate) are missing parameters.

I tracked the problem down to this snippet:

```javascript
const setImmediate = global.setImmediate || function (callback) {
  // works in node v0.8
  nextTick(callback)
}
```

[Here](https://github.com/mqttjs/MQTT.js/blob/main/lib/client.js#L20) is the problematic function mentioned above.

It does not implement the setImmediate as per the documentation [here](https://nodejs.org/dist/latest-v16.x/docs/api/timers.html#setimmediatecallback-args) **with** the callback args.

But using core-js's polyfill is for now a good enough alternative (need to be imported before the mqtt lib):

```
import 'core-js/modules/web.immediate'
```

filipomar:
[PR](https://github.com/mqttjs/MQTT.js/pull/1497) submitted

PR #1497, opened 2022-07-01
fix: error events in a browser environment are emitted without any parameters

filipomar:
In a browser, where the `setImmediate` and `nextTick` polyfills are needed, the Errors being passed such as [this scenario](https://github.com/mqttjs/MQTT.js/blob/3451564363cf0933355ab6d66e73c80fa13379e8/lib/client.js#L726) are ignored.

Issue was originally raised [here](https://github.com/mqttjs/MQTT.js/issues/1474).


robertsLando:
Supersided by #1626 #1628

PR #1509, opened 2022-07-16
fix(types): topic alias controls and password

bretambrose:
* Adds typescript entries for the two top-level topic alias controls in client config: autoUseTopicAlias and autoAssignTopicAlias
* Modifies password entry to properly reflect the fact that password is, by specification, binary data (https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901072).  Leave string in for backwards compatibility.


robertsLando (approved):

robertsLando:
@YoDaMa This could be merged

PR #1519, opened 2022-08-08
fix: topicAliasMaximum under must be under Connect properties

RoanBrand:
It looks the `topicAliasMaximum` is under the wrong object.
This seems to be affecting https://github.com/emqx/MQTTX


robertsLando (changes requested):
@RoanBrand Could you submit a test that covers this to prevent regressions in future?

robertsLando (approved):

Issue #828, opened 2018-05-30
TransformUrl function returns a async promise instead of new url string


mxdi9i7:
I'm trying to request for a new signed url from a server that already signs the url, the operation is async and by returning the promise I wasn't able to have the MQTT socket connected (it was never connected, went straight to offline). 
```
mqtt.connect({
      protocal: 'ws',
      transformWsUrl: () => {
        return renewStreamUrl(endpoint, token, conversationId) //this is an async operation
      }
    }
  )
```
The websocket just starts to reconnect nonstop with a new url.

The expected behavior is: transformUrl callback returns a resolved promise, which in my case is the signed url, but it seems like its only returning a wrapped Promise which causes my mqtt client to reconnect nonstop because it didn't find a valid stream URL

adamlc:
This would be useful. Or a callback :)

ivands:
I need this feature aswell.
Please add

cto-tvd-build:
I need this feature as well!

Please add! 👍

maschulze:
I need this feature as well!

Please add!

burritoIand:
can't you just resolve the promise prior to returning?
```
async transformWsUrl() {
    const url = await renewStreamUrl(...);

    return url;
}
```

onip:
code calling that function is sync, won't be able to handle a promise or an async function which is the same.

burritoIand:
@onip gotcha, wondering if it's bc fundamentally this lib wasn't built with promise support and thus async-mqtt exists? Would need to refactor the code in connect/ws.js perhaps

onip:
the code using `transformWsUrl()` and the functions around it are all sync. I've looked to provide a MR in the past and the change was too big for my spare time.

I've resolved setting up a `setInterval()` to always have a fresh and valid token in order to be able to return it in a sync fashion.

burritoIand:
@onip ok yeah in my app I have a separate authentication jwt/refresh mechanism that runs on an interval based on token expire, then in transformWsUrl I just use the latest value..

xiaoxiangmoe:
In the current situation, I used a synchronous xhr request to transformWsUrl. This does not sound like a good idea.

viridia:
I need this feature as well. I've looked at the code and yes, it would be non-trivial, since all of the various functions that call it would need to be async as well.

My use case is similar to others: The URL includes a credential which has to be fetched by a REST call, and that credential may have changed in the interval since the connection was originally established.

[...]

PR #1522, opened 2022-08-11
feat: reconnect callback

jclab-joseph:
Added cbReconnect in option.
#828 can also be solved

---

- [X] add unit test


robertsLando (approved):

robertsLando (changes requested):
Could you add this to docs?

PR #1534, opened 2022-09-08
fix: tests that relied on order of events

BertKleewein:


robertsLando (commented):
Was this ready for review?

BertKleewein:
@robertsLando  - yes, this is ready for review.

test/abstract_client.js
```diff
@@ -1418,29 +1438,41 @@ module.exports = function (server, config) {
...
       client.once('connect', function () {
-        client.unsubscribe(topic)
+        client.unsubscribe(topic, function (err) {
+          assert.ifError(err)
+          assert(received)
+          client.end(done)
+        })
       })
 
+      // TODO: all of these test changes have to do with calling end() before the callback is complete. Maybe that should be fixed first
```
robertsLando, line 1470: Would you mind to do this?

test/abstract_client.js
```diff
@@ -2983,7 +3019,12 @@ module.exports = function (server, config) {
 
         client.on('connect', function () {
           if (!reconnect) {
-            client.publish('topic', 'payload', { qos: 2 })
+            client.publish('topic', 'payload', { qos: 2 }, function (err) {
+              server2.close()
```
robertsLando, line 3023: Does server.close expects a callback too?

robertsLando (changes requested):

robertsLando:
@BertKleewein Should this fix flaky tests problem? I noticed that sometime tests fail with no reason

robertsLando:
@BertKleewein kindly ping

test/abstract_client.js
BertKleewein: It looks like it does. Do you think I should update the tests to use it? I assume it will help, but I can't guarantee that it won't cause other problems.

BertKleewein:
@robertsLando - it should fix _some_ flaky tests. I don't know if it will fix the flaky tests that you were seeing though. :)

test/abstract_client.js
BertKleewein: Depends on what you mean by "this" :)  Do you mean "make these tests better" or do you mean "fix the problems that happen when you call `client.end` from inside a callback?"

IMHO, one of the ugliest problems in this code (MQTT.js), is the spaghetti flow that happens when you call `client.end`.  I've tried to make sense of the order that things get shut down, and there are too many branches and patches and callbacks inside that code to make me confident in _any_ changes to the shutdown code. 

If the flow of `client.end` does get fixed, it needs to be done very carefully. I assume that customers have worked around its oddities (whether they know it or not), and it just really feels like changing how `client.end` works is opening a large can of worms.

BertKleewein: And I forgot to say that I'm assuming that you want to make these tests better.  Sure, I can do that.

BertKleewein:
@robertsLando  - ready again. I think this second set of changes really improves these tests. I found a few things that might explain general flakiness.

test/abstract_client.js
```diff
@@ -14,6 +14,27 @@ const serverBuilder = require('./server_helpers_for_client_tests').serverBuilder
...
+  * server2        | The name used for mock brokers that are created for an individual test and then destroyed.
+  * serverClient   | An socket on the mock broker. This gets created when your client connects and gets collected when you're done with it.
+  *
+  * Also worth noting:
+  *
+  * `serverClient.disconnect()` does not disconnect that socket. Instead, it sends an MQTT disconnect packet.
+  * If you want to disconnect the socket from the broker side, you probably want to use `serverClient.destroy()`
+  * or `serverClient.stream.destroy()`.
+  *
+  */
```
robertsLando, line 36: Thanks for this! Will be super useful 👍🏼

robertsLando (approved):
Looks good to me! I'm quite sure that respecting all callbacks makes tests much more reliable

test/abstract_client.js
robertsLando: > Depends on what you mean by "this" :) Do you mean "make these tests better" or do you mean "fix the problems that happen when you call client.end from inside a callback?"

Sorry for not being clear. I assumed both, anyway I already checked the `end()` code and it's a real mess so we can keep that for another PR 🙏🏼

test/abstract_client.js
robertsLando: Looks better now :)

PR #1538, opened 2022-09-20
process stale issues from oldest to newest and increase operations / run

ryanwinter:
Process the older issues first as these are the most stale, and also increase the operations / run so we can get through the backlog.


ryanwinter:
@BertKleewein: Can you approve this to more efficiently process the stale issues. So far only we only have [14 stale issues](https://github.com/mqttjs/MQTT.js/labels/stale) because its hitting the limit.

BertKleewein (approved):

PR #1553, opened 2022-11-11
Update browserify, add missing client.off method to mqtt.min.js

### LINKED ISSUE #1552: Error: `client.off is not a function` in mqtt.min.js
The main entry is not compatible with webpack 5, so I switch to the bundled  `mqtt.min.js`, which is built by browserify@16 / events@2,  and the `off` method only exists in [events v3](https://github.com/browserify/events/releases/tag/v3.0.0).

ambar:
Fixes #1552

```bash
browserify 16.5.2 (main branch)
734K mqtt.js
306K mqtt.min.js

browserify 17.0.0
734K mqtt.js
305K mqtt.min.js
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
         "@types/ws": "^7.4.7",
         "aedes": "^0.46.2",
         "airtap": "^4.0.4",
-        "browserify": "^16.5.0",
+        "browserify": "^17.0.0",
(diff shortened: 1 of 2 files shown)


robertsLando:
Could you update to latest?

robertsLando (approved):

PR #1555, opened 2022-11-15
docs: add `mqtt-jsonl-store` to available stores

robertsLando:


robertsLando:
@vishnureddy17 no code changes here BTW giving almost no mqtt store works well could be a nice alternatives for the once looking for a working mqttjs store

PR #1572, opened 2023-01-20
fix: add missing export of UniqueMessageIdProvider and DefaultMessageIdProvider

ogis-ksuzuki:
I tried to build the source code using the UniqueMessageIdProvider with webpack, but it could not be called correctly.

for example)

```js
import * as mqtt from 'mqtt';
let provider = new mqtt.UniqueMessageIdProvider();
```

The reason is that in the webpack build, "browser" field of package.json **"./mqtt.js": "./lib/connect/index.js"** causes the entry point to be "./lib/connect/index.js" and that "index.js" does not contain an export of UniqueMessageIdProvider.

In case of node build, './mqtt.js' is referenced and 'mqtt.js' exports UniqueMessageIdProvider.
Same for DefaultMessageIdProvider.

I added a DefaultMessageIdProvider and UniqueMessageIdProvider export to "index.js" and confirmed that the issue was resolved and the unit tests passed.


robertsLando:
@ogis-ksuzuki Could you fix conflicts please?

ogis-ksuzuki:
@robertsLando Resolved merge conflict.

robertsLando (approved):

Issue #1264, opened 2021-04-01
React Native Support - How to configure?


larry-cherry:
I have been reading over some of the past issues about getting this library to work with React Native. I would like to use this library because it looks like it is the most supported one out there for javascript. I have not found a guide that can be followed to get this project working on React Native. I have read several posts from this repo linked below that talk about some of the difficulties getting it to work. It does appear that people have gotten this project to work with React Native but I am not sure what their project setup looks like. What I am asking is if it is possible to get this repo working with React Native can there be a README file created that walks through the setup process and possibly common issues that arise? I have listed a few of the issues that mention the setup process below but do not know if any of these are relevant. 

- [React-Native support #573](https://github.com/mqttjs/MQTT.js/issues/573)
- [MQTT after version 4.2 does not work in react-native #1242](https://github.com/mqttjs/MQTT.js/issues/1242)
- [Cannot import in React Native #979](https://github.com/mqttjs/MQTT.js/issues/979)

larry-cherry:
I figured I would post some of my findings on getting this library to work with react native. I found you have to set up the buffer globally in your react native app. I did this in my index.js file. The following comments helped me with that configuration [573#issuecomment-428583410](https://github.com/mqttjs/MQTT.js/issues/573#issuecomment-428583410) and [573#issuecomment-467283871](https://github.com/mqttjs/MQTT.js/issues/573#issuecomment-467283871). You also need a way of passing the appropriate net client to the tcp.js file. The require for net is a little quirky so I modified the streambuilder function in the file MQTT.js/lib/connect/tcp.js to accept an optional argument of the net client. That appeared to fix all the issues that I had. I will post a copy of the updated tcp.js file below. If there is a better way to pass the net client please let me know. 

`
'use strict'
var net = require('net');
var debug = require('debug')('mqttjs:tcp')

function streamBuilder (client, opts, netClient) {
  var port, host
  opts.port = opts.port || 1883
  opts.hostname = opts.hostname || opts.host || 'localhost'

  port = opts.port
  host = opts.hostname

  debug('port %d and host %s', port, host)
  if(!netClient){
    netClient = net;
  }
  return netClient.createConnection(port, host);
}

module.exports = streamBuilder
`

inkCrazy:
+1

DannieL8:
+1

inkCrazy:
[...]

Issue #573, opened 2017-03-23
React-Native support


franleplant:
Hi There!

Thanks a lot for your work on this amazing library.

I've been fighting a lot with the browser builds, for several reasons:

- Im using React Native and I need to manually create a dist file
- Im using Typescript and I cannot include that dist file in my source because Typescript will complain no end
- I tried other libraries and frankly this one is the best one
- Playing around with node_modules is no good, there's no easy solution for automating steps inside that directory and it's also a transient directory, I don't want to get too attached to it in the case I need to `rm -rf node_modules`, which is too often

So, that's my problem, and here is my solution, which aims to be as simple as it gets

- Add `dist/mqtt.js` to the source files, so I can easily `require(mqtt/dist/mqtt)`. Names can be discussed but the important thing is to have that file available

There are other variations of this solution but I think this works and it's super simple and you already got the script to generate those dist files, the only remaining thing to do would be to un-ignore `dist` and to fix all the pre-commit checks that fail on those dist files

What do you think?

There might be more fancy solutions and Im open to anything but this requires minimal work and provides maximum satisfaction

franleplant:
Other easy solution is to add a postinstall hook that creates the build every time someone installs the package

mcollina:
Hey @franleplant, I understood your solution but I did not understand the problem you are facing.

You can easily do `require('mqtt/dist/mqtt')`, it should work. It's inside the package in the lastest versions, so there should  be no problem.

franleplant:
Wow, apparently I'm blind, I need 👓 . I did found however this in the gitignore https://github.com/mqttjs/MQTT.js/blob/master/.gitignore#L10 which might have confused me.

So, two more things before closing
- It'd be nice to have this stated in the docs instead of stating that the users need to manually build the browser package (I can help)
- Im getting this error while using the `dist/mqtt`

```
Unable to resolve module `./store` from `Users/flp/code/MyProject/node_modules/mqtt/dist/mqtt.js`
```

This is the way Im importing it

```
const mqtt = require('mqtt/dist/mqtt')
```

The mqtt dist are made with browserify and my project uses webpack and react native, there might be some incompatibilities with the module systems?

Thanks a lot for your help

mcollina:
@franleplant that is highly possible. You probably have to have your webpack build it together with your application

franleplant:
[...]

PR #1574, opened 2023-01-31
feat: react-native support

rrebase:
Add support for React Native by leveraging the `react-native` entrypoint (metro bundler makes use of it), extending the browser behaviour and polyfilling anything that still broke

This does introduce a [tiny](https://bundlephobia.com/package/safe-buffer@5.2.1) package `safe-buffer` to dependencies, which NodeJS doesn't need.

Resolves https://github.com/mqttjs/MQTT.js/issues/1264, https://github.com/mqttjs/MQTT.js/issues/573

### DIFF
--- a/lib/connect/react-native.js
+++ b/lib/connect/react-native.js
@@ -1,0 +1,20 @@
+'use strict'
+
+// React Native might not have a process object, so we need to polyfill it
+if (typeof process === 'undefined') {
+  // eslint-disable-next-line no-global-assign
+  process = {}
+};
+
+// We need set the value of process.title, which mqtt.js uses to determine if it's running inside a browser
+// like environment and thus knows it doesn't have access to NodeJS core modules like net, tls
+process.title = 'browser'
+
+// Legitimate polyfill for process.nextTick, necessary for Duplexify, which is used by mqtt-packet
+process.nextTick = setImmediate
+
+// Polyfill for Buffer, which is present in NodeJS core, but not in the React Native environment
+global.Buffer = global.Buffer || require('safe-buffer').Buffer
+
+// Entry point for a browser like environment
+module.exports = require('./index')
--- a/package.json
+++ b/package.json
@@ -60,6 +60,12 @@
     "tls": false,
     "net": false
   },
+  "react-native": {
+    "./mqtt.js": "./lib/connect/react-native.js",
+    "fs": false,
+    "tls": false,
+    "net": false
+  },
   "dependencies": {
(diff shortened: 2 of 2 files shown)


umangloria:
@rrebase  I tried your branch to get it working on RN but getting this error related to 'net' package. 
Simulator Screen Shot - iPhone 14 - 2023-03-28 at 23 10 33

rrebase:
@umangloria it seems that the shims were not applied (bundler didn't use the `react-native` entry point) 🤔

What version of RN do you have there, is it using the metro bundler?

umangloria:
> @umangloria it seems that the shims were not applied (bundler didn't use the `react-native` entry point) 🤔
> 
> What version of RN do you have there, is it using the metro bundler?

@rrebase  I just created a fresh project by `npx react native init MqttDemo` so it the latest version. Strange enough, I got stuck at this `stream.pipe(` crash. I am out of ideas as I already tried installing that stream package directly in project's `package.json` file. But no luck 😕

umangloria:
I looked further into this and I think it is because on react-native it is not able to find `net` module. I think if it is react native then we should be using `react-native-tcp-socket`. I am not sure how to do that as every time it tries to look for net package from the node.

robertsLando:
@rrebase @umangloria anyone that would like to finish this?

garzfaust:
This version does not work. It is missing the NPM package `url`. When you install it with `npm install url`, it does.

*Attention 1:* The drawback of using MQTT.js in React Native is, that it can only send via WebSocket, but noch via plain TCP. Thus your MQTT broker needs to support WebSocket.

*Attention 2:*: MQTT.js is swallowing errors! If the error number is not known to MQTT.js, it does not emit the error event, thus you will never realise that there was an error using `client.on('error', ...)`. I had the situation, that it would not connect to the broker, without reporting any errors. I had to debug through the MQTT.js code to find out, that there was an error indeed, but it was swallowed. I would doubt that this is good library design. If a error number is undefined, you might report an unknown error number. Reporting no error at all for sure is not good.

### Here are instructions on how to get the React Native supported fork working

At first, install the branch `react-native-layer` of this MQTT.js fork: https://github.com/rrebase/MQTT.js/tree/react-native-layer

`npm install git+https://github.com/rrebase/MQTT.js.git#react-native-layer`

Then install the missing `url` package.

`npm install url`

*Note:* Instead of installing the `url` package, it is also possible to work with this solution here instead: https://github.com/parshap/node-libs-react-native

PR #1601, opened 2023-05-05
fix: weixin min program:trigger onclose event when the server disconnected

Jax2014:


robertsLando:
What does this fixes? Is there a related issue? Can you add a test that covers this?

PR #1614, opened 2023-06-26
chore: release-it and semantic pr title

robertsLando:
This PR adds release-it module to handle releases and semantic PR title checks to improve changelogs.

Mind that in order to make the `release-it` action to work `NPM_TOKEN` secret must be created

Anyway a script `release` has been added to `package.json` that can be used to create releases interactively (much better then doing them manually like we do now)

In order to prevent overriding changelog, the old changelog has been renamed to `CHANGELOG.old.md`

I also did an `npm audit fix` that fixed 6 high security vulnerabilities

### DIFF
--- a/.github/workflows/release-it.yml
+++ b/.github/workflows/release-it.yml
@@ -1,0 +1,54 @@
+# #########################################################################
+# Creates a new release using `workflow_dispatch` event trigger with `type`
+# as input to describe the type of release to create
+name: 'Release-it: Create a new release on demand'
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type. Can be `patch` `minor` or `major`'
+        required: true
+        default: 'patch'
+
+jobs:
+  release-it:
+    if: github.actor == 'robertsLando'
+    runs-on: [ubuntu-latest]
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v3
+        with:
+          ref: 'master'
+          fetch-depth: 0 # fetch all commits history to create the changelog
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3.6.0
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'npm'
+
+      - name: Install dependencies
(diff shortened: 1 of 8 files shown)


.github/workflows/release-it.yml
```diff
@@ -0,0 +1,54 @@
...
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type. Can be `patch` `minor` or `major`'
+        required: true
+        default: 'patch'
+
+jobs:
+  release-it:
+    if: github.actor == 'robertsLando'
```
vishnureddy17, line 16: Is there a way to use `release-it` without hardcoding a GitHub username?

vishnureddy17:
Could you write a brief markdown file describing how to release? I'm approving the PR anyway since my concerns are minor.

vishnureddy17 (dismissed):

.github/workflows/release-it.yml
robertsLando: Yeah this is just to prevent someone else to be able to trigger releases, I can add your username if you want or completely remove this check

robertsLando:
Ok for the markdown file, anyway releasing is as easy as typing "npm run release" and follow the interactive cli: it will firstly ask you the gh token then will do the tests, update changelog, create a release commit with release tag and create the release on gh and npm (will ask for the otp too).

.github/workflows/release-it.yml
vishnureddy17: They won't be able to do anything without write access to this repo or the npm package right? If that's the case, I vote to remove it.

vishnureddy17:
> What about eslint+prettier in place of standard?

I think it should be a different PR, because I imagine changing linter and formatter will change a lot of the code, and I think it's better to have all those changes in its own commit

robertsLando:
Agree!

Would you mind to add the 2 secrets mentioned above?

vishnureddy17:
> Would you mind to add the 2 secrets mentioned above?

I've actually never used GitHub actions before so I will have to do some digging to understand how to make that work.

Is this PR blocked on that?

.github/workflows/release-it.yml
robertsLando: Yeah, ok!

vishnureddy17:
@robertsLando Thanks! Gonna check it out anway since I should probably know this.

Issue #1541, opened 2022-09-30
Input requested: updating the code and porting to TypeScript


BertKleewein:
I'm seeking input on an idea for moving this repo forward. 

My team at Microsoft uses MQTT to communicate with the Azure IoTHub service. __MQTT.js does an excellent job here, but, frankly, it's starting to show its age. It was written when the node.js ecosystem was less mature, and it has evolved as the ecosystem changed. Many (but not all) of these "evolutionary steps" were bolted-on pieces of technology that "added a little more to MQTT.js". The result is code that does what it needs to do, but it’s hard to understand and even harder to change.

@yodama and @vishnureddy17 looked at this problem and started to build an updated version of this library from the ground up. This is a noble pursuit. It’s also extremely easy to underestimate how difficult it is. While MQTT, the transport, is relatively simple, libraries with thousands of dependents, like MQTT.js, are not so simple. The thousands of dependents, or customers, or users, or whatever you want to call yourselves are what make MQTT.js powerful. "Rebuilding from the ground up" means "moving forward by moving away from these customers." I personally do not like this idea.

Instead, we should take the time and take a large evolutionary step to bring the existing MQTT.js code "into the modern age." We should call this MQTT.js version 5.0.0, with the major version change from 4 to 5 showing that this is a breaking change. But we should only break things _when there is a clear advantage_. Changes in API or behavior from version 4.7.3 to version 5 should be justified and documented. Customers who want to "play it safe" can use semantic versioning rules in package.json to make sure they do not get version 5. Other customers who want to "follow along" _shouldn’t_ need to make major changes, but we cannot guarantee this.

We should convert this code to typescript. I have done this as a POC (proof-of-concept). The extra type information has shown me so many places where the code is sloppy and could be tightened up, especially around options and client properties and callback signatures. You can see my POC [here](https://github.com/BertKleewein/MQTT.js/tree/typescript-poc/src)

We should also start using prettier and eslint on the code. I hate arguing about "what format is correct" for pretty-printers. I love being able to write code without worrying about formatting because I know it will get automatically formatted. Unfortunately, this means that every line of code gets changed, and merges from "pre-prettier" code into "post-prettier" code gets ugly. I wish I knew a way to avoid this.

We should start publishing commonjs and ECMAScript side-by-side if we can. It's too bad that there is still a distinction here, but I _think_ we can make both sides happy without picking a side. 

[...]

PR #1615, opened 2023-06-27
chore!: drop support for node 12-14

robertsLando:
BREAKING CHANGE: Dropped support for NodeJS 12-14

### DIFF
--- a/.github/workflows/mqttjs-test.yml
+++ b/.github/workflows/mqttjs-test.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
       fail-fast: false
 
     steps:
--- a/.github/workflows/release-it.yml
+++ b/.github/workflows/release-it.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: [ubuntu-latest]
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [18.x]
     steps:
       - name: Checkout master
         uses: actions/checkout@v3
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "typescript": "^4.5.5"
       },
       "engines": {
-        "node": ">=12.0.0"
+        "node": ">=16.0.0"
       }
     },
     "node_modules/@airtap/browserify-istanbul": {
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "mqtt.js"
   ],
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=16.0.0"
   },
   "browser": {
     "./mqtt.js": "./lib/connect/index.js",


robertsLando:
Is there anything else I should do to drop the support?

vishnureddy17:
Pretty sure that's it.

vishnureddy17:
@robertsLando Since this is a breaking change, I think we should take this opportunity to see if there's any other breaking changes we'd want to make before the next release. One thing I have in mind is removing the WeChat specific stuff, I do not think it belongs in this.

vishnureddy17:
Yep that sounds good.

vishnureddy17:
Yep, but it doesn't have to be in this test beta release unless you want to do it now :)

I think there are references to it in the Readme that would have to be removed also

robertsLando:
Yeah I'm checking them now there are many references around the code

vishnureddy17:
It's not a standard transport protocol and it's instead specific to a proprietary product. I don't think that belongs in an open source project.

> Yeah I'm checking them now there are many references around the code.

Yeah, that's the problem, the proprietary stuff has seeped into lots of areas in the code, which I think is a mistake.

However, this library doesn't have support for custom transports, so maybe we have to tolerate this or add support for custom transport.

robertsLando:
> However, this library doesn't have support for custom transports, so maybe we have to tolerate this or add support for custom transport.

Yeah that's what I wanted to point out, I mean without a custom transport support all users using that protocol would have no alternatives

robertsLando:
Done: https://github.com/mqttjs/MQTT.js/releases/tag/v5.0.0-beta.0

robertsLando:
@vishnureddy17 If that is ok I would wait to merge some other PR marked with `good-to-merge` label and then make the official release. How can we inform users to try using the beta ?

vishnureddy17:
Yes, I agree with waiting on those PRs, there is no rush

To inform users I think replying to [this GitHub issue](https://github.com/mqttjs/MQTT.js/issues/1541) would be a good idea, since there might be interested parties that are subscribed to that issue. Also maybe add a temporary announcement somewhere near the top of the README.

BearCooder:
Hey @robertsLando thank you for the dedication and work! Its great to see that you picked it up and try to manage/maintain mqtt.js!

I want just make a suggestion. Have you think about Github Sponsors for MQTT.js library? https://github.com/sponsors

Just to make it sustainable, the time and effort of the team to contribute to this library. I believe MQTT.js should setup github sponsors so it can accept donations. After all time is precious and I remember this tweet from Matteo Collina - https://twitter.com/matteocollina/status/1083322337292890112?lang=en
Open Source must be sustainable :) 

And I would suggest to also add the beta release note at the top of the README (temporarily) So its the first that users see that a v5 beta to test is available when they open the repository and scroll to the readme

grafik

robertsLando:
Hi @BearCooder, 

I already use GH sponsors in all my projects and I think that would be a good addition for sure. Anyway the project isn't mine so I would like to know @vishnureddy17 opinion on this... how could we manage this? We could enable GH sponsor to the organization but then I dunno who could maintain that

robertsLando:
> Maybe a column in this part of the readme with links to sponsor individual contributors?

Maybe adding a `## Sponsor` section to readme could make it more visible

PR #1622, opened 2023-06-29
feat: option to disable `writeCache` and fix leak in subscriptions

robertsLando:
Move commits from master


robertsLando:
@vishnureddy17 I dunno if there are other cleaner ways to do this, I found some old PR that I merged were pointing to master, I noticed this just now and with this commit I'm moving them to main (squash).

vishnureddy17:
I don't think we should squash this one.

vishnureddy17 (approved):

robertsLando:
@vishnureddy17 I enabled rebase but GH tells me there are conflicts (even if this pr is even with main branch)

vishnureddy17:
In that case, could we squash merge but make sure that the titles of the underlying commits (including the PR numbers) are in the commit description? That way someone can still go through the commit history and find the relevant PRs in the future.

Once we are sure that everything is in main and master doesn't have anything important, I think we should delete master to avoid future confusion.

This is probably really annoying, I really appreciate your handling this :)

robertsLando:
Sure! No problem, I understand that :) I will do it on Monday

robertsLando:
@vishnureddy17 agree about the lock, not sure about deleting it. I would keep it for now, anyway PR merge should fail right now. Not sure if there is an easier way to move all open PR from master to main, maybe someone ha built a tool for that?

PR #1625, opened 2023-06-30
chore: use new `GH_TOKEN` secret

robertsLando:
I have created an organization secret named `GH_TOKEN` with a fine grained organization token to be used in release workflow. It has restricted access only to this organization repo with `content` permissions (create releases and push commits).

This should allow the release workflow to successfully create the releases and bypass the main branch restrictions when pushing commits directly on main branch (`chore(release): xxx` commit)

### DIFF
--- a/.github/workflows/release-it.yml
+++ b/.github/workflows/release-it.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           ref: 'main'
           fetch-depth: 0 # fetch all commits history to create the changelog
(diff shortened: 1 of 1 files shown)


vishnureddy17 (approved):

robertsLando:
@vishnureddy17 Just to inform you, I just did the 5.0.0-beta.2 release using the workflow and it worked now 🎉

PR #1627, opened 2023-07-03
chore: add browser tests to ci

robertsLando:


robertsLando:
@BertKleewein @vishnureddy17  I fixed browser tests in https://github.com/mqttjs/MQTT.js/pull/1628 because them were not working at all (wondering if they ever worked). Now them work locally but seems airtap doesn't like github actions env... 

Based on ther [docs](https://github.com/airtap/airtap#github-actions) it should work but tests always get stucked... I tracked the bug down to a call to https://www.npmjs.com/package/open with the browser url that never returns.

Wondering if you know how this could be fixed

BertKleewein:
I don't think these tests were ever working. I had a private branch long ago that had changes like #1628 but I never followed that all the way through.

This PR didn't work because airtap was configured to test in the default system browser but it needs to run headless to work inside github actions. Updating the airtap config to test with chrome headless makes it pass. 

I noticed two things about this:
1. If add another browser to `.airtap.yml`, the second one will fail about half the time.  I'm guessing someone is leaving the websocket connection open. I would be tempted to blame `server.js` since it's so hacktastically simple, but I don't know this for sure.
2. `test.js` is importing `lib/connect`, but I assume we want to test against `dist/mqtt.js` and/or `dist/mqtt_min.js` since that's what we expect browsers to be using.

robertsLando:
> This PR didn't work because airtap was configured to test in the default system browser but it needs to run headless to work inside github actions. Updating the airtap config to test with chrome headless makes it pass.

Yeah I imagined that, bad that they don't wrote this in their docs BTW, I tried putting headless as first try but it wasn't working on my end, I maybe missed the npm package, I never used airtap so 🤷🏼‍♂️ 

> test.js is importing lib/connect, but I assume we want to test against dist/mqtt.js and/or dist/mqtt_min.js since that's what we expect browsers to be using.

Yeah I think so. Let me do that, thanks so much for the fix, as you can see I did multiple tries to make this working without success 😆

robertsLando:
Cusrious thing:

- If I use `dist/mqtt.min` tests works
- If I use `dist/mqtt` tests doens't work, throws: `MODULE NOT FOUND ../is-browser`

Wondering why...

BertKleewein:
@robertsLando - Renaming the config file from `.airtaprc.yml` to `.airtap.yml` was the missing ingredient.  This is probably why adding `headless: true` didn't work for you. (and probably why I stopped trying to add this to CI earlier. :) )

I don't know why `mqtt.js` doesn't work and `mqtt.min.js` does, but I assume it's related to the way that browserify and terser work.  I saw your draft PR for moving to rollup, and I think this is the way to go. I don't see any value in investigating the odd behavior here if we're going to replace these tools with rollup. I think it's OK to ignore this issue for now.

robertsLando:
Perfect so we are done here! 

About the rollup PR, I'm not a rollup expert and I'm trying to find out the best setup for it. If you have some suggestions there would be appreciated:)

PR #1629, opened 2023-07-06
chore: more code cleanup surrounding client.end callbacks

BertKleewein:
More test changes surrounding `client.end` callbaacks.  One `// BUGBUG:` comment marks a bug that I found with these changes.


test/client.js
```diff
@@ -108,6 +110,9 @@ describe('MqttClient', function () {
 
       client.on('message', function (t, p, packet) {
         if (++count === max) {
+          // BUGBUG: the client.end callback never gets called here
+          // client.end((err) => done(err))
```
robertsLando, line 114: I use mqttjs in lot of my projects and this happend to me too, I always create a MqttClient wrapper library that handles all client stuff, this is how I handle the close usually: https://github.com/zwave-js/zwave-js-ui/blob/master/lib/MqttClient.ts#L154. I think this could happen when for some reason underlig socket is already closed/destroyed or when store cannot be closed for some reason

test/client.js
```diff
@@ -168,10 +173,10 @@ describe('MqttClient', function () {
           unsubscribeCallbackCalled = true
         })
         setTimeout(() => {
-          client.end(() => {
+          client.end((err) => {
             assert.strictEqual(pubCallbackCalled && unsubscribeCallbackCalled, true, 'callbacks not invoked')
             server3.close()
```
robertsLando, line 178: Does `server3.close` expects a callback too?

robertsLando (changes requested):
Well done with this cleanup! I would like to discuss about the "bugbug"

test/client.js
robertsLando: No clue if this could help: https://github.com/mqttjs/MQTT.js/pull/713

BertKleewein: I merged #713 into a local branch and it didn't resolve this issue. It is an interesting change though. 

One of my goals with these test changes is to get the behavior of `client.end` well tested before we make any changes to it. Right now, it's just too hard to reason about how `client.end` might behave and I'm afraid to make any changes to it.

robertsLando: I agree! Tests are essentials if we want to make improvements to the code

PR #1633, opened 2023-07-10
chore!: remove unused deps, convert to ES2015 class

robertsLando:
BREAKING CHANGE: when creating an `MqttClient` instance `new` is now required


robertsLando:
Opinion on this. Many deps are related to `bin/*`: 

- `commists`
- `concat-stream`
- `help-me`
- `minimist`
- `split2`

So this are 5 deps  on a total of 15. I would suggest to create a new dedicated repo  `@mqttjs/cli` and move that code there. What's your opinion on this?

robertsLando:
Next dep to remove is [xtend](https://github.com/Raynos/xtend) that is also archived right now. I would create an util method instead

robertsLando:
cc @vishnureddy17

vishnureddy17:
> So this are 5 deps on a total of 15. I would suggest to create a new dedicated repo @mqttjs/cli and move that code there. What's your opinion on this?

It is not costless. There's some overhead because we have to pay attention to another repo for maintenance. 

I still think it's good on net, reducing the dependencies here would be great!

lib/connect/index.js
```diff
@@ -68,7 +67,7 @@ function connect (brokerUrl, opts) {
       parsed.port = Number(parsed.port)
     }
 
-    opts = xtend(parsed, opts)
```
vishnureddy17, line 71: Good call to get rid of dependencies that are no longer needed due to newer JS features.

lib/client.js
```diff
@@ -247,1672 +245,1670 @@ function nop (error) {
...
-    } else {
-      this.options[k] = options[k]
+    // Defaults
+    for (k in defaultConnectOptions) {
+      if (typeof this.options[k] === 'undefined') {
+        this.options[k] = defaultConnectOptions[k]
+      } else {
+        this.options[k] = options[k]
+      }
     }
```
vishnureddy17, line 264: This might be out of scope for this PR, but it looks like this code can be simplified with spread syntax

robertsLando:
I have created https://github.com/mqttjs/MQTT.js/issues/1634 to keep track of this. Is this good to merge so? 

@vishnureddy17

vishnureddy17:
Still reviewing, almost done.

vishnureddy17 (approved):

Issue #1511, opened 2022-07-19
Inconsistent publish() callback invocations on puback receipt


bretambrose:
This may be intentional but I found it somewhat of a bummer.

If the received puback indicates a failure (by reason code), the user callback will be invoked with the puback packet: https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/client.js#L1614

If the received puback indicates a success, the user callback will be invoked with the original publish packet (happens in the store implementation, where the input parameter 'packet' gets aliased/shadowed by a local with the same name): https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/store.js#L90-L94

The end result is that there is no way to find out a successful puback's ResponseString and UserProperties (which are probably empty but for completeness sake, it would be nice to have theoretical access).

ogis-yamazaki:
I noticed this problem too.

User callback is executed twice if the received PUBACK indicates failure (by reason code).

invoke callback 1st here. 
https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/client.js#L1614

invoke callback 2nd here. 
https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/client.js#L1617
https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/store.js#L94

robertsLando:
Fixed by #1635

PR #1635, opened 2023-07-12
fix: problem with publish callback invoked twice

ogis-yamazaki:
User callback is executed twice if the received PUBACK indicates failure (by reason code). 
Related issue: #1511

The following modification prevents the callback from invoked twice.

- If ``resonCode`` in PUBACK to failure, set the cb variable to NULL after the callback is invoked.
- When deleting a message from the store, do not invoke the callback if cb is null

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1593,7 +1593,7 @@ class MqttClient extends EventEmitter {
     const messageId = packet.messageId
     const type = packet.cmd
     let response = null
-    const cb = this.outgoing[messageId] ? this.outgoing[messageId].cb : null
+    let cb = this.outgoing[messageId] ? this.outgoing[messageId].cb : null
     const that = this
     let err
 
@@ -1626,6 +1626,7 @@ class MqttClient extends EventEmitter {
           err = new Error('Publish error: ' + errors[pubackRC])
           err.code = pubackRC
           cb(err, packet)
+          // prevent invoking callback twice when deleting message from store #1511
           cb = null
         }
         delete this.outgoing[messageId]
@@ -1646,6 +1647,12 @@ class MqttClient extends EventEmitter {
           err = new Error('Publish error: ' + errors[pubrecRC])
           err.code = pubrecRC
           cb(err, packet)
+          // prevent invoking callback twice when deleting message from store #1511
+          cb = null
+          delete this.outgoing[messageId]
+          this.outgoingStore.del(packet, cb)
+          this.messageIdProvider.deallocate(messageId)
+          this._invokeStoreProcessingQueue()
         } else {
           this._sendPacket(response)
         }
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1626,6 +1626,7 @@ class MqttClient extends EventEmitter {
           err = new Error('Publish error: ' + errors[pubackRC])
           err.code = pubackRC
           cb(err, packet)
+          cb = null
```
robertsLando, line 1629:
```suggestion
          // prevent invoking callback twice when deleting message from store #1635
          cb = null
```

robertsLando (changes requested):
Thanks for the PR @ogis-yamazaki ! Would you mind to add a unit test that covers this?

ogis-yamazaki:
@robertsLando 

Added qos 1 and qos2 tests.

When an error occurs in PUBREC, added a process to remove the message from the ``outgoingStore``.

robertsLando (approved):
LGTM

robertsLando (commented):
By checking the code on my side I saw that this: 

```
delete this.outgoing[messageId]
      this.outgoingStore.del(packet, cb)
      this.messageIdProvider.deallocate(messageId)
      this._invokeStoreProcessingQueue()
```

Is used 3 times in the code. Would you mind to create a dedicated function for that?

ogis-yamazaki:
@robertsLando 

> Is used 3 times in the code. Would you mind to create a dedicated function for that?

OK

> I see we also have a method removeOutgoingMessage, it doesn't call _invokeStoreProcessingQueue and messageIdProvider.deallocate maybe that's an error too?

As you said, additional processing was needed.

In ``removeOutgoingMessage``, needs to call messageIdProvider.deallocate.
If not called, the messageId will be in use forever.

``_invokeStoreProcessingQueue`` must also be called.
Normally, ``_invokeStoreProcessingQueue`` is called when a PUBACK or similar event occurs, but ``removeOutgoingMessage`` will not pass through it.
The ``_storeProcessingQueue`` may remain unprocessed.

The ``_invokeStoreProcessingQueue`` function also seems to have an implementation problem.

There is a problem if ``_invokeStoreProcessingQueue`` is called while the store is resending(_storeProcessing = true).
Specifically, Message overtaking may occur.
It is necessary to modify the content of _storeProcessingQueue`` so that it is not processed during resend(_storeProcessing = true).

robertsLando:
Would you mind to also fix those errors or do you prefer to do that in a separeted PR?

ogis-yamazaki:
@robertsLando 

> Would you mind to also fix those errors or do you prefer to do that in a separeted PR?

Fixed in this PR.

robertsLando (approved):
LGTM, now everything is much cleaner. Thanks @ogis-yamazaki 🙏🏼

PR #1636, opened 2023-07-14
refactor: move handlers to lib/handlers

### LINKED ISSUE #1257: MQTT keepalive less than 60s and chrome 88 intensive timer throttling
- it would be great to have some documentation regarding best practices in regards to keepalive and chrome intensive timer throttling
- it is not uncommon for our application tab to be put in the background by our users since they can also spend a fair amount of time in other web applications
- our keepalive is less than 60s since we need to know fairly quickly when the connection with the broker has been severed since we require timely message delivery
- chome 88  only executes timers every minute during intensive timer throttling which means our keepalive does not happen within the configured time frame which in turn causes disconnects
- please advise, thanks

https://docs.google.com/document/d/105Hlbcb1mTV06bT9aDpwBotY13eZiKV5Hl4PW5LlyeE/edit
https://developer.chrome.com/blog/timer-throttling-in-chrome-88/

AB#9481421

robertsLando:
- Created `client.log` function that by default uses `debug` but can be overriden using `options.log`
- Created `client.nop`
- Moved all `client._handle` in `lib/handlers`
- Create `MqttClient.defaultId` static method
- Ensure shiftPingInterval is called every time a packet is written to the stream. Could fix #1257. An alternative could also to shift the ping interval every time a pingresp is received from server

### DIFF
--- a/test/client_mqtt5.js
+++ b/test/client_mqtt5.js
@@ -617,16 +617,16 @@ describe('MQTT 5.0', function () {
 
   it('auth packet', function (done) {
     this.timeout(15000)
-    server.once('client', function (serverClient) {
+    const opts = { host: 'localhost', port: ports.PORTAND115, protocolVersion: 5, properties: { authenticationMethod: 'json' }, authPacket: {} }
+    const client = mqtt.connect(opts)
+    server.once('client', function (c) {
       console.log('server received client')
-      serverClient.on('auth', function (packet) {
+      c.on('auth', function (packet) {
         console.log('serverClient received auth: packet %o', packet)
-        serverClient.end(done)
+        client.end(done)
       })
     })
-    const opts = { host: 'localhost', port: ports.PORTAND115, protocolVersion: 5, properties: { authenticationMethod: 'json' }, authPacket: {} }
     console.log('calling mqtt connect')
-    mqtt.connect(opts)
   })
 
   it('Maximum Packet Size', function (done) {
--- a/README.md
+++ b/README.md
@@ -423,6 +423,7 @@ The arguments are:
   - `resubscribe` : if connection is broken and reconnects,
     subscribed topics are automatically subscribed again (default `true`)
   - `messageIdProvider`: custom messageId provider. when `new UniqueMessageIdProvider()` is set, then non conflict messageId is provided.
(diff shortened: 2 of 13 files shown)


test/client_mqtt5.js
```diff
@@ -617,16 +617,16 @@ describe('MQTT 5.0', function () {
...
+    const client = mqtt.connect(opts)
+    server.once('client', function (c) {
       console.log('server received client')
-      serverClient.on('auth', function (packet) {
+      c.on('auth', function (packet) {
         console.log('serverClient received auth: packet %o', packet)
-        serverClient.end(done)
+        client.end(done)
       })
     })
```
robertsLando, line 628: @BertKleewein By checking all failed actions seems the only test that occasionally keeps failing was this one. Let's see if this change fixes that too

robertsLando (commented):
Check

test/client_mqtt5.js
robertsLando: Nope that doesn't fix the test neighter... I have a feel this is caused by a race condition, the auth packet could be sent even before the connack is received and the `c.on('auth')` listener may be set after the auth is received causing the test to fail

robertsLando:
@vishnureddy17 @BertKleewein Could you tell me your thoughts about this?

Once this is merged I will continue with https://github.com/mqttjs/MQTT.js/issues/1637. Standard doesn't highlight errors like eslint, like undefined vars and others, that's absolutely bad and difficult to see at first sight expecially when refactoring things...

vishnureddy17 (approved):
Looks great, much needed refactor.

robertsLando:
@BertKleewein My idea right now is to make the code cleaner and easier to work with without breaking to much stuff. I understand what you mean and I think we can reach that goal, anyway mocking the client to unit test those handlers wouldn't be that hard right now, but I understand you would like to isolate the access those handlers have, that's possible too but we can keep that for another PR :) 

My next steps will be to remove standard with eslint+prettier (sure this will highlight lot of other bugs in code) and then thinking about moving to Typescript. For the second part what scares me most is rewriting tests... Converting all the files to typescript could require me a day but the tests are a lot. Anyone that could help me with that when I start?

PR #1638, opened 2023-07-18
chore: replace standard with eslint+prettier

### LINKED ISSUE #1496: Manual connect
Is it possible to manually initialize the connection? Eg. create MqttClient and then call .connect method. From what I've read, such a functionality is not possible but I might be mistaken. Is it something you consider in next major update?

### LINKED ISSUE #1637: Replace standard with eslint + prettier

robertsLando:
Fixes #1637

Additional changes:

- `errors` was not defined in `handlers/auth`
- set workflows `DEBUG` env based on if we are running tests with debug mode enabled
- update actions checkout and setup node to v3
- use arrow functions everywhere, drop the need to use `that`
- add `lint` and `lint-fix` npm scripts and added `lint` to ci
- fix `should not reconnect if pingresp is successful` test where client connection was not closed properly
- fix `auth packet` flaky test caused by a race condition (see comment)
- added missing `authPacket` client option
- add `manualConnect` option to allow user manually calling `connect` (previously called `_setupStream`). This trick is used when user want to setup listeners or other thing before doing the connect. This is used in `auth packet` test for example to prevent the race condition. Also Fixes #1496

I also tried to fix the mocha `after` method in `test/client` and `test/unique_message_id_provider_client`. It actually calls `server.close()` without checking if there are open sockets (connections). I did a change to keep a reference of all sockets and close them before server.close and call `done` callback provided by `after` on server close. I also tried to change the code using afterEach and beforeEach so the server is created and closed before and after each test (as I think should be) but that doesn't work too.


robertsLando:
@vishnureddy17 @BertKleewein as expected this has also fix some bugs in code like undefined vars that now are highlighted by eslint. Also everything has a cleaner syntax now using arrow functions

robertsLando:
Ready for review now @BertKleewein @vishnureddy17 🙏🏼

lib/connect/index.js
```diff
@@ -1,27 +1,25 @@
-'use strict'
-
+const url = require('url')
```
vishnureddy17, line 1: Perhaps we could do this everywhere we use a dependency built into node to make things clear. Not sure if this would affect brower compatibility.

```suggestion
const url = require('node:url')
```

lib/connect/ws.js
```diff
@@ -1,257 +1,265 @@
...
+			objectModeMode: pOptions.objectMode,
+		})
+
+		_proxy._write = socketWrite
+		_proxy._flush = socketEnd
+
+		return _proxy
+	}
+
+	function onopen() {
```
vishnureddy17, line 196: nit: camelCase this
Might be out of scope for this PR
```suggestion
	function onOpen() {
```

lib/handlers/publish.js
```diff
@@ -25,96 +25,143 @@ const validReasonCodes = [0, 16, 128, 131, 135, 144, 145, 151, 153]
...
+	const { options } = client
+	if (client.options.protocolVersion === 5) {
+		let alias
+		if (packet.properties) {
+			alias = packet.properties.topicAlias
+		}
+		if (typeof alias !== 'undefined') {
+			if (topic.length === 0) {
+				if (alias > 0 && alias <= 0xffff) {
+					const gotTopic =
```
vishnureddy17, line 43: nit: I feel like the newline here is not really useful. Maybe increase the max line length setting?

lib/client.js
```diff
@@ -50,1444 +56,1613 @@ const socketErrors = [
...
+		}
+
+		const streamErrorHandler = (error) => {
+			this.log('streamErrorHandler :: error', error.message)
+			if (socketErrors.includes(error.code)) {
+				// handle error
+				this.log('streamErrorHandler :: emitting error')
+				this.emit('error', error)
+			} else {
+				this.nop(error)
```
vishnureddy17, line 312: nit: I think changing the name here would make things slightly more clear. Not sure if this was a typo that carried on for a long time or if it was intentional. 
```suggestion
				this.noop(error)
```

README.md
```diff
@@ -10,7 +10,7 @@
 MQTT.js is a client library for the [MQTT](http://mqtt.org/) protocol, written
 in JavaScript for node.js and the browser.
 
-> MQTT [5.0.0-beta.0](https://github.com/mqttjs/MQTT.js/releases/tag/v5.0.0-beta.0) is now available! Try it out and give us feedback! `npm i mqtt@5.0.0-beta.0`
+> MQTT [5.0.0 BETA](https://www.npmjs.com/package/mqtt/v/beta) is now available! Try it out and give us feedback! `npm i mqtt@beta`
```
vishnureddy17, line 13: Maybe we should have a specific place to submit feedback? Perhaps a pinned issue where people can put comments

vishnureddy17 (approved):
Made some suggestions, but feel free to merge without addressing them if you think they are out of scope for this PR.

lib/connect/ws.js
robertsLando: Done, anyway the reason this wasn't camelacase is to follow `socket.onopen` that is not.

lib/connect/index.js
robertsLando: I just did a try, seems browserify doesn't support `ndoe:` protocol yet :( we can keep this maybe for the rollup refactor if we ever find a way to make that work

robertsLando: Reference: https://github.com/browserify/browserify/issues/2019

ljharb: there's no real benefit to using `node:`; just stick with `url`.

lib/handlers/publish.js
robertsLando: Line length increesed to 100 (default is 80)

README.md
robertsLando: https://github.com/mqttjs/MQTT.js/issues/1639

lib/client.js
robertsLando: Done

lib/handlers/publish.js
robertsLando: I think I would revert this as I found out: https://prettier.io/docs/en/options.html#print-width

Issue #1288, opened 2021-06-10
connAckPacket.sessionPresent is not checked for mqtt3.1.1


yyjdelete:
Seems #917 only add check for it when mqtt5.0 is used, but the flag is also available and should also be checked for mqtt3.1
This will make resubscribe and later manually subseibe with the same topic failed if `option.clean` is set to `false` but sever already clean the session when reconnected.
https://github.com/mqttjs/MQTT.js/blob/37b12cb94737acddfe0fa016e013516ba3b9a031/lib/client.js#L1453

YoDaMa:
Hey @yyjdelete thanks for identifying this. could you provide a repro example so I can run it and identify this bug and that the fix handles it?

yyjdelete:
@YoDaMa 
You need an your own MQTT broker(an local mosquitto for me) and stop and restart it manually after see `Please restart mqtt server now`,

```js
var mqtt = require('mqtt')

var cid = 'mqttjs_' + Math.random().toString(16).substr(2, 8)
var i = 0

var test = function(testSessionExpire){
	//test with an local mosquitto or other server
	var client  = mqtt.connect('mqtt://127.0.0.1', {
		clean: false,
		clientId: cid,
		reconnectPeriod: testSessionExpire ? 1000 : 0
	})

	client.on('connect', function () {
		console.log('connect=>' + (++i))
		client.subscribe('presence', function (err) {
			console.log('presence', err)
			if (!err) {
				client.publish('presence', 'Hello mqtt=>' + (+new Date()))
			}
		})
	})

	client.on('message', function (topic, message) {
		// message is Buffer
		console.log(message.toString())
		if (testSessionExpire)
			console.log('Please restart mqtt server now')
		else
			setTimeout(()=>test(false), 1000)
	})
}
test(true)
```

As you can see, subscribe didn't report any error, but the second `Hello mqtt` is missing, seems the subscribe after reconnect is simple ignored, and you can never receive any new message.
I capture with wireshark (testSessionExpire = true and false) and ensure at least mosquitto can properly set `Session Present` flag
```
connect=>1
presence null
Hello mqtt=>1632933523810
Please restart mqtt server now
connect=>2
presence null
```

robertsLando:
[...]

PR #1650, opened 2023-07-26
fix: resubscribe when no session present (#895)

### LINKED ISSUE #895: Client does not resubscribe after reconnect if clean = false
Is there any reason we don't want resubscribe work if clean = false?
I have created a sub client with clean = false, and resubscribe = true, the client will not resubscribe the topics if MQTT aedes broker (no persistence) restart. And, sub-client will no longer receive the message of the topics
https://github.com/mcollina/aedes

https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js
From MQTT.js/lib/client.js

```js
 MqttClient.prototype._resubscribe = function () {
  if (!this._firstConnection &&
      this.options.clean &&
      Object.keys(this._resubscribeTopics).length > 0) {
    if (this.options.resubscribe) {
      this._resubscribeTopics.resubscribe = true
      this.subscribe(this._resubscribeTopics)
    } else {
      this._resubscribeTopics = {}
    }
  }
```
I can provide my sub client code and broker code if necessary

simonnilsson:
The combination **clean=false,  resubscribe=true, sessionPresent=false** should now trigger a resubscribe for MQTT v3.1.1, not only for v5.

Closes #895, #1288

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -2106,7 +2106,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		if (
 			!this._firstConnection &&
 			(this.options.clean ||
-				(this.options.protocolVersion === 5 &&
+				(this.options.protocolVersion >= 4 &&
 					!this.connackPacket.sessionPresent)) &&
 			_resubscribeTopicsKeys.length > 0
 		) {
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -3165,6 +3165,39 @@ export default function abstractTest(server, config) {
 			})
 		})
 
+		it('should resubscribe when clean=false and sessionPresent=false', function _test(done) {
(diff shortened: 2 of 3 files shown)


src/lib/client.ts
```diff
@@ -2106,7 +2106,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		if (
 			!this._firstConnection &&
 			(this.options.clean ||
-				(this.options.protocolVersion === 5 &&
```
robertsLando, line 2109: Could you add a comment here with the links to MQTT 3.1.1 specs that explain this?

robertsLando (changes requested):

src/lib/client.ts
simonnilsson: Done.

robertsLando (approved):

Issue #1655, opened 2023-08-01
Replace airtap with wtr


robertsLando:
Airtap doesn't support esm modules out of the box.

https://modern-web.dev/docs/test-runner/overview/

PR #1658, opened 2023-08-02
chore: replace `browserify` with `vite`

robertsLando:
- Replaced browserify with vite
- Added webpack support in vite example
- Improved vite example by removing the unnecessary module
- Improved `exports` so bundlers takes the correct file when targeting `browser`. (see also browser test now can import from root instead of specify the dist folder)


robertsLando:
I noticed just a problem with this. In the provided binary seems buffer is polifilled 3 times, I think one for mqtt, another for mqtt-packet and maybe another for readable-stream. Anyone has suggestions about this? I think it has always been like this even with browserify

@vishnureddy17 @BertKleewein

robertsLando:
Other problem is that airtap uses browserify under the hoods and it doesn't support [esm](https://github.com/airtap/airtap/issues/332). So also this #1655 may be needed in order to get rid of browserify limitations

vishnureddy17:
Ah bummer. I am unsure on how to get rid of the duplicate polyfills, but it's not a blocker, it it? maybe it can be tracked in a separate issue

robertsLando:
I ended up using vite in place of webpack. Perfromances are much better, works both with `webpack` and `vite` and also browserify seems working well with the produced bundle. I'm just not sure if it will work well in browser when directly imported using cdn.

robertsLando:
Supersided by https://github.com/mqttjs/MQTT.js/pull/1731

PR #1676, opened 2023-08-19
fix: move exported types out of dev dependencies

### LINKED ISSUE #1673: 5.0.3 fails to compile in TS project
I am trying simple thing 
1. NPM install mqtt in a fresh typescript project
2. add import * as mqttClient from 'mqtt'; as suggested by the readme
3. run tsc

It is giving me lot of errors 
1. Missing type definitions for @types/readable-stream and @types/ws
2. Error TS1259 Module '"stream"' can only be default-imported using the 'esModuleInterop' flag for file node_modules/@types/node/stream.d.ts:1425:5 

What am i doing wrong?

markormesher:
The types that are exported for this project use types from `readable-stream` and `ws`, so those types need to be included as proper dependencies that get shipped to consumers. This PR moves those dependencies out of the dev block, and also bumps the Node types to make sure they work with the latest additions to the `readable-stream` types (i.e. using `asyncDispose`).

This ought to fix #1673 (it does for one of my project seeing the same error as that issue).

One downside of this is that consuming projects will need to be on a fairly recent version of `@types/node`, but I don't think that's the worst thing vs. not working at all for TS projects right now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,8 @@
       "version": "5.0.3",
       "license": "MIT",
       "dependencies": {
+        "@types/readable-stream": "^4.0.1",
+        "@types/ws": "^8.5.5",
         "commist": "^3.2.0",
         "concat-stream": "^2.0.0",
         "debug": "^4.3.4",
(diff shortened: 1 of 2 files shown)


robertsLando (approved):

markormesher:
What's the schedule for cutting releases of this library? Would be great to upgrade this in some projects over the weekend.

robertsLando:
I'm on vacation right now so probably next week

andrewperella:
any update on this?

markormesher:
@andrewperella This change should be in 5.0.4 onwards

PR #1683, opened 2023-08-30
docs(readme): update with ALPNProtocols for supporting multiple TLS protocols

ysfscream:
**Title:** Document ALPNProtocols Option for Enhanced TLS Support

**Description:**

This PR adds documentation for the `ALPNProtocols` option in MQTT.js, enhancing clarity for users who wish to utilize multiple TLS protocols on the same port.

**Rationale:**

We've observed that many users were uncertain about how to correctly configure multiple TLS protocols, such as MQTTS and MQTT over WSS, on the same port. While MQTT.js inherently supports this through the `ALPNProtocols` option, it wasn't clearly documented. This omission can lead to confusion and potential misconfiguration.

By clearly detailing the `ALPNProtocols` option, we aim to provide clearer guidance, reduce user uncertainty, and ensure users can fully leverage the library's capabilities without hassle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ Async [`connect`](#connect). Returns a `Promise` that resolves to a `mqtt.Client
 
 The `Client` class wraps a client connection to an
 MQTT broker over an arbitrary transport method (TCP, TLS,
-WebSocket, ecc). 
+WebSocket, ecc).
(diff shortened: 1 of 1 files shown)


ysfscream:
@robertsLando Thanks!

Issue #1686, opened 2023-09-06
v5 type issues


dgg:
Hello,

Our team is migrating our typescript projects from _4.3.7_ to _5.0.4_ and we are hitting some problems that prevent us from upgrading.

We are facing three types of problems:
1. Some types are "missing" in the sense that they are not exported, even though they are used and, in some cases, we can make use of them, but their type cannot be exposed (like returned from a function, for instance.
2. There is some type inconsistency with the runtime behavior. For example, an error that is set to null even though the typing specifies being `Error` or `undefined`.
3. Other type inconsistency appears when some members that can be obtained runtime (and were exposed in previous versions) but are not declared in the types.
4. Other inconsistency would be some functions declare callbacks in which the first error is `Error | null` and some other callbacks' first argument is the more standarized `Error | undefined`

Some examples of each problem:

1.
```javascript
import { connect,MqttClient } from "mqtt"
// missing types
import {/*IConnackPacket */ } from "mqtt"

const client: MqttClient = connect("mqtt://0.0.0.0:1883")
    // cannot type connPacket (missing exported type)
    // "usable" with anonymous func, cannot attach typed function declared somewhere else
    .on("connect", (connPacket/*: IConnackPacket*/) => {
    // ...
    })
```
```javascript
    // cannot type msgPacket (missing exported type)
    .on("message", (topic: string, payload: Buffer, msgPacket /*: IPublishPacket*/){ }
```
2.
```javascript
/* type declaration in mqqtjs */
publish(topic: string, message: string | Buffer, opts?: IClientPublishOptions, callback?: DoneCallback): MqttClient;
export type DoneCallback = (error?: Error) => void;

/* client code: */
client.publish("test/topic", "hello", { qos: 2 }, (pubError?: Error, pubPacket?: unknown) => {
    if (pubError) {
       console.error("Error publishing", pubError)
    } else {
        // null pubError is returned as opposed of what the callback type specifies (undefined)
        console.info("Published", pubPacket, pubError)
    }
}) 
```
```shell
# running example
$ Published { cmd: 'pubrel', qos: 2, messageId: 58612 } null
```
3.
With the same code from point 2), we are able to get an untypeable pubPacket on runtime. That member is not declared in the callback type.
4. Error argument is not consistent: `null` or `undefined` in the case of success. 
```javascript
[...]

PR #1688, opened 2023-09-07
fix: publish/subscribe/unsubscribe types and missing types exports

dgg:
Fixes some of the issues mentioned in #1686.

In particular:
1) exporting missing types from callbacks
3) missing callback argument

I tried to figure out how to fix 2) and 3) (inconsistent error argument declaration in callbacks and runtime/typing mismatch) but it goes deeper into the library (writing to streams) than I am comfortable with.
Maybe it is easier than I think, but I believe it needs some more overview of the project than I have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "mqtt",
   "description": "A library for the MQTT protocol",
-  "version": "5.0.4",
+  "version": "5.0.5",
   "contributors": [
     "Adam Rudd <adamvrr@gmail.com>",
     "Matteo Collina <matteo.collina@gmail.com> (https://github.com/mcollina)",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "mqtt",
   "description": "A library for the MQTT protocol",
-  "version": "5.0.5",
+  "version": "5.0.4",
   "contributors": [
     "Adam Rudd <adamvrr@gmail.com>",
     "Matteo Collina <matteo.collina@gmail.com> (https://github.com/mcollina)",
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -355,6 +355,7 @@ export type ISubscriptionMap = {
 	resubscribe?: boolean
 }
 
+export { IConnackPacket, IDisconnectPacket, IPublishPacket, Packet }
 export type OnConnectCallback = (packet: IConnackPacket) => void
 export type OnDisconnectCallback = (packet: IDisconnectPacket) => void
 export type ClientSubscribeCallback = (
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mqtt",
   "description": "A library for the MQTT protocol",
-  "version": "5.0.4",
```
robertsLando, line 4: Please don't bump the version. I have an automated process that does this on release

robertsLando (changes requested):
It's not very clear to me where in the code we return the packet in publish method... 

Publish calls `_sendPacket that then calls `_writePacket` that simply calls `cb` on end without any arg

package.json
dgg: "umbumped" the version

dgg:
> It's not very clear to me where in the code we return the packet in publish method...
> 
> Publish calls `_sendPacket` that then calls `_writePacket` that simply calls `cb` on end without any arg

Yes, I agree with you it is not very obvious when reading the code, but if you look at the tests `publishing > should fire callback ...` in _test/abstract_client.ts_ it is undeniable that the callback __is__ called with a non-falsy packet argument: either a **publish** `cmd` for QoS 1 or a **pubrel** for QoS 2:

```shell
      ✔ should fire a callback (qos 0)
QOS 1 {
  cmd: 'publish',
  topic: 'a',
  payload: 'b',
  qos: 1,
  retain: false,
  messageId: 30300,
  dup: false
}
      ✔ should fire a callback (qos 1)
QOS 1 ERR {
  cmd: 'publish',
  topic: 'a',
  payload: 'b',
  qos: 1,
  retain: false,
  messageId: 27647,
  dup: false
}
      ✔ should fire a callback (qos 1) on error
QOS 2 { cmd: 'pubrel', qos: 2, messageId: 15779 }
      ✔ should fire a callback (qos 2)
QOS 2 ERR { cmd: 'pubrel', qos: 2, messageId: 12513 }
```

(I console-logged the packet in the callback invoked in the tests).

Besides, that runtime behavior of calling the callback with the packet was the behavior in v4 of the library.

robertsLando:
@dgg Wondering if it just would make more sense to don't return the packet when doing a publish at all. What is the possible use case behind this considering it's not always defined?

dgg:
> @dgg Wondering if it just would make more sense to don't return the packet when doing a publish at all. What is the possible use case behind this considering it's not always 

I do not know about the use case (maybe the one that created the function in the first place or someone that actually uses it can) but in our case we would have to change quite a few places in which we were declaring the argument.

In reality it is a matter of backwards compatibility. There was a function with a callback signature that was not marked as deprecated or anything and it goes away in the next major release (fair, but better warn beforehand) but in reality, the runtime behavior is maintained, only the declaration of the types changed.

dgg:
Btw, I can see that the x20 build is broken. I was going to ask how to fix the issues, but I figured out already.

robertsLando (approved):

robertsLando:
5.0.5 is coming

PR #1696, opened 2023-09-18
feat: custom websocket support

helios57:
We need to add special protocols to connect to our mqtt broker over websocket. With this small addition (which hopefully helps others as well) we can do it very easy. Otherwhise we need to build our own streamBuilder and copy alot code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,28 @@ use the latest auth token, you must have some outside mechanism running that
 handles application-level authentication refreshing so that the websocket
 connection can simply grab the latest valid token or signed url.
 
+#### Customize Websockets with `createWebsocket` (Websocket Only)
+
+When you need to add a custom websocket subprotocol or header to open a connection
+through a proxy with custom authentication this callback allows you to create your own
+instance of a websocket which will be used in the mqtt client.
+
+```js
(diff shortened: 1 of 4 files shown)


robertsLando (changes requested):
Please add a unit test that covers this feature and also mention it in docs

helios57:
Thanks for the fast reply, unfortenately I am not able to run the test on windows and installing everyting on linux took me to long.
I tried to add some tests and can also update the doc but as long as I cant run the tests it's pretty useless and I can't add better tests.

The error seems to be known but I could not find a working solution:
> node_modules/.bin/nyc --reporter=lcov --reporter=text ./node_modules/mocha/bin/_mocha -r ts-node/register test/*.ts --exit

Der Befehl "node_modules" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.

helios57:
Just found something working:
If I change this line it seems to run:
 "unit-test:node": "nyc --reporter=lcov --reporter=text ./node_modules/mocha/bin/_mocha -r ts-node/register test/*.ts --exit",

I will add more tests and update the docu.

helios57:
Thanks for the fast response. Unittest is now added and the docs should be updated.

robertsLando (approved):

PR #1697, opened 2023-09-18
refactor: replace `mocha` with `node:test`

### LINKED ISSUE #1694: [refactor] Rewrite tests using tap/nodejs test runner
This could help making tests less flaky

/cc @mcollina

robertsLando:
Fixes #1694 

/cc @mcollina


mcollina (commented):
good job! tests are failing

robertsLando:
@mcollina Tests are working now BTW them are much slower then before

mcollina:
> A possible way to improve this would be to enable parallel tests. Problem is that abstract_client tests is used in multiple test files and it setup some servers that listen on a port (so this will end up with port conflict). I would like to know if there is a way using process.pid or else to shift ports so there are no conflicts. Otherwise we should use a package to get the first free open port every time listen is called

Indeed. Avoiding reusing those broker instances will fix a lot of the flakyness. It's expected that the tests will take longer. As you said, the solution is to enable parallel testing and use different ports.

robertsLando:
@mcollina I tried multiple approaches when running tests. The first was by using `test/runTests.ts`:

```ts
import { readdirSync } from 'node:fs'
import { run } from 'node:test'
import process from 'node:process'
import { tap } from 'node:test/reporters'
import { basename } from 'node:path'

const files = readdirSync(__dirname)
	.filter((f) => f.endsWith('.ts') && f !== basename(__filename))
	.map((f) => `${__dirname}/${f}`)

run({ files, timeout: 60 * 1000, concurrency: true })
	.compose(tap)
	.pipe(process.stdout)

```

This runs the test concurrently but what I don't like is that if some test fail this is not reported at all, the test passes anyway as I think the exit code is 0 always, I'm not sure how to fix this, maybe by using a custom reporter?

I also tried to use `spec` reporter but I think types are wrong here too, it returns a class constructor and it cannot be used as `dot` or `tap`. The only reporter I found to work is `tap`, the others doesn't work at all...

At the end I used another way to run test using `node -r esbuild-register --test test/*.ts` and that seems more reliable for now.

Apart from this the PR is ready for review.

robertsLando:
I have been able to cut down time to ~2min. Don't know if we can do better here

test/runTests.ts
```diff
@@ -0,0 +1,57 @@
...
+	summary.push(
+		`${data.file} - "${data.name}" (${Math.round(
+			data.details.duration_ms,
+		)}ms)\n${error.toString()} `,
+	)
+})
+
+testStream.on('test:stderr', (data) => {
+	summary.push(`${data.file} - Error:\n${data.message} `)
+})
```
robertsLando, line 41: Dunno why but the summary is missed when running tests using `run` function. It is printed only when running them from a single file. Maybe it's a bug or I dunno, I created my custom summary in this way, also this is needed to correct the exit code in case there are errors

test/helpers/port_list.ts
```diff
@@ -1,51 +1,36 @@
...
-const PORTAND316 = PORT + 316
-const PORTAND326 = PORT + 326
-const PORTAND327 = PORT + 327
-const PORTAND400 = PORT + 400
+/**
+ * Method used to get ports for testing
+ * @param i Index to shift the ports by
+ * @returns
+ */
+export default function getPorts(i = 0) {
```
robertsLando, line 6: trick to be sure ports doesn't overlap, each file must use a different `i` in order to work

test/abstract_store.ts
```diff
@@ -1,25 +1,29 @@
...
 export default function abstractStoreTest(
 	build: (cb: (err?: Error, store?: IStore) => void) => void,
 ) {
 	let store: IStore
 
-	// eslint-disable-next-line
-	beforeEach(function (done) {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-expect-error
+	beforeEach((_ctx, done) => {
```
robertsLando, line 13: This is the type error I mentioned above, use @ts-expect-error so in case this get fixed typescript will tell us

robertsLando:
Ping @mcollina

PR #1705, opened 2023-09-28
fix: import type error

JimmyBjorklund:
node_modules/mqtt/build/lib/shared.d.ts:3:13 - error TS1259: Module '"stream"' can only be default-imported using the 'esModuleInterop' flag

3 import type internal from 'stream';
              ~~~~~~~~

  node_modules/@types/node/stream.d.ts:1726:5
    1726     export = internal;
             ~~~~~~~~~~~~~~~~~~
    This module is declared with 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

Found 1 error in node_modules/mqtt/build/lib/shared.d.ts:3

### DIFF
--- a/src/lib/shared.ts
+++ b/src/lib/shared.ts
@@ -1,5 +1,5 @@
 import type { Packet } from 'mqtt-packet'
-import { Duplex } from 'stream'
+import type { Duplex } from 'stream'
 import type MqttClient from './client'
 import type { IClientOptions } from './client'
 
--- a/src/lib/shared.ts
+++ b/src/lib/shared.ts
@@ -1,5 +1,5 @@
 import type { Packet } from 'mqtt-packet'
-import type internal from 'stream'
+import * as stream from 'stream'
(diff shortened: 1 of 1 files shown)


src/lib/shared.ts
```diff
@@ -1,5 +1,5 @@
 import type { Packet } from 'mqtt-packet'
-import type internal from 'stream'
```
robertsLando, line 2: cannot you import Duplex type instead?

robertsLando (changes requested):

src/lib/shared.ts
```diff
@@ -1,5 +1,5 @@
...
+import { Duplex } from 'stream'
```
robertsLando, line 2:
```suggestion
import type { Duplex } from 'stream'
```

robertsLando (changes requested):

robertsLando (approved):

PR #1715, opened 2023-10-16
fix: garbage at pingTimer

### LINKED ISSUE #1714: Garbage at pingTimer
I found some garbage in the interval from pingTimer. When I log the heap, strings of the methode _checkPing() are retained.
A quick test with this.pingTimer.destroy() seems to fix the problem.

```diff --git a/src/lib/client.ts b/src/lib/client.ts
index 24bffe9..de7e302 100644
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -647,9 +647,9 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
 
-			this.log('close :: clearing ping timer')
+			this.log('close :: destroy ping timer')
 			if (this.pingTimer !== null) {
-				this.pingTimer.clear()
+				this.pingTimer.destroy()
 				this.pingTimer = null
 			}
 
@@ -1768,8 +1768,8 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		}
 
 		if (this.pingTimer !== null) {
-			this.log('_cleanUp :: clearing pingTimer')
-			this.pingTimer.clear()
+			this.log('_cleanUp :: destroy pingTimer')
+			this.pingTimer.destroy()
 			this.pingTimer = null
 		}
 

```

Leubeling:
I found some garbage in the interval from pingTimer. When I log the heap, strings of the methode _checkPing() are retained.
A quick test with this.pingTimer.destroy() seems to fix the problem.

Fixes #1714

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -647,9 +647,9 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
 
-			this.log('close :: clearing ping timer')
+			this.log('close :: destroy ping timer')
 			if (this.pingTimer !== null) {
-				this.pingTimer.clear()
+				this.pingTimer.destroy()
 				this.pingTimer = null
 			}
 
@@ -1768,8 +1768,8 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		}
 
 		if (this.pingTimer !== null) {
-			this.log('_cleanUp :: clearing pingTimer')
-			this.pingTimer.clear()
+			this.log('_cleanUp :: destroy pingTimer')
+			this.pingTimer.destroy()
 			this.pingTimer = null
 		}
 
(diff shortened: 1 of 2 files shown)


robertsLando:
@Leubeling I'm worndering if this could be an issue inside `reInterval` package. Could you check? 

The reason we use that package is that it should be faster then re-creating intervals, no clue if using `destroy` here is correct in that case: https://github.com/4rzael/reInterval/blob/master/index.js#L21

Leubeling:
instead re-creating ??? That makes sense.
But then, instead to set "null", we have to reuse the "pingTimer".

I added another one including tests.

robertsLando:
@Leubeling Re-interval is based on [retimer](https://github.com/mcollina/retimer#how-it-works), essentially instead of deleting and creating a new timer every time we reschedule them it let the old timer trigger and reschedule it with the time missing, most performance benefits are only when the reschedule happens after the timeout as explained in the docs I linked. 

What I told you in my previous answer is that it may be an issue in original [reinterval](https://github.com/4rzael/reInterval) package.

Giviong that module is almost unmaintained and that retimer also added a support for web workers recently, would you mind to try replacing reInterval with retimer? Otherwise I think your initial solution that used destroy would be fine as long as it fixes the memory leak

robertsLando:
Supersided by #1753

Issue #1021, opened 2019-12-16
Client <unknown> disconnected due to protocol error.


pavankumarkl:
I am using MQTT and below are options that I am passing to connect

var options = {
    keepalive: 10,
    clientId: '',
    protocolId: 'MQTT',
    protocolVersion: 3,
    clean: true,
    reconnectPeriod: 20000,
    connectTimeout: 30 * 1000,
    protocol:'mqtt',
    rejectUnauthorized: false

  }
  const client = mqtt.connect('mqtt://192.168.127.13:1883',options); -> my local mosquitto broker

can any one help, why I am getting "Client <unknown> disconnected due to protocol error."

AB#9368700

MShokry:
same here Any update?
```
const us = {
    clientId: 'xiot_' + Math.random().toString(16).substr(2, 8),
    username: mqtt.username,
    password: mqtt.password,
    useSSL: false,
    onSuccess: onAction,
    onFailure: onAction,
    protocolId: 'MQTT',
    protocolVersion: 5,
    rejectUnauthorized: false,
    clean: true,
    reconnectPeriod: 20000,
    connectTimeout: 30 * 1000,
    protocol: 'mqtt',
  };
  let client1 = mqtt1.connect('mqtt://192.168.1.16:1883/',us);
```

fabien-lg:
Same here!

jc-delvalle:
So I had the same issue, new mosquitto server installed on OSX with homebrew and using react-native.

For some reason no native mqtt connection could be made so I used websockets.

You'll need to uncomment from mosquitto.conf

`#port 1883`

and add just below: 

```
listener 8883
protocol websockets 
```

When you tun mosquitto it should say:

```
1583550048: mosquitto version 1.6.8 starting
1583550048: Config loaded from /usr/local/etc/mosquitto/mosquitto.conf.
1583550048: Opening ipv6 listen socket on port 1883.
1583550048: Opening ipv4 listen socket on port 1883.
```

Then connect using: 

`const client = mqtt.connect('ws://localhost:8883', options);`

Hope this helps someone.

naanuswaroop:
I was getting the error "Client disconnected due to protocol error" in the below scenario:
1. Mosquitto broker started without specifying any certs (cafile, certfile and keyfile) in the mosquitto config file.
2. Mosquitto client (pub or sub) tries to connect to the above broker using proper certificates. 

Fix:
[...]

PR #1723, opened 2023-10-24
fix: unambiguously detect web workers

f1234k:
Fixes the issue described [here](https://github.com/mqttjs/MQTT.js/issues/1021#issuecomment-1777498470).


f1234k:
Another, maybe better, approach would be:

`(typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope)`

robertsLando:
@f1234k Why did you deleted your PR?

f1234k:
@robertsLando I was not happy with it. I will submit a new one by Friday.

PR #1724, opened 2023-10-25
fix: crash with React Native

fattomhk:
RN [support Websocket](https://reactnative.dev/docs/network#websocket-support), better to use it directly.

**Sample code for detecting is React native:** 
https://snack.expo.dev/-QQjqDN-8

Discussion: 

- [navigator.product ](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/product)is removed from modern web standard
- React native is not a browser, should we put it in `isBrowser` function?


robertsLando (approved):

robertsLando:
> React native is not a browser, should we put it in isBrowser function?

The `is-browser` env is useful only to prevent loading nodejs related protocols, if react-native doesn't support them that's like a browser so it's ok

robertsLando (approved):

PR #1728, opened 2023-10-29
fix: unambiguously detect web workers

f1234k:
Use the [same method](https://github.com/flexdinesh/browser-or-node/blob/master/src/index.js) that `browser-or-node` NPM package uses to detect web workers. This has been tested against my use case for the issue described [here,](https://github.com/mqttjs/MQTT.js/issues/1021#issuecomment-1777498470) and it works as expected: the Bun runtime is not being (erroneously) identified as a web worker.

P.S: Apologies for cancelling the PR initially, but I didn't feel right about it, and I had to run some more tests against my project.


robertsLando (dismissed):

robertsLando:
@f1234k #1724 also edits this files so you may need to fix conflicts (seems there are also lint issues)

robertsLando (changes requested):
Please remove the bun.lockdb and pacacke-lock.json changes from the PR

f1234k:
@robertsLando sorry for that. Should be fine now.

robertsLando (approved):
No worries! LGTM

PR #1731, opened 2023-11-07
feat: esm version `dist/mqtt.esm.js` and replace `browserify` with `esbuild`

robertsLando:
Replace `browserify` with `esbuild`


robertsLando:
Needs: https://github.com/mqttjs/mqtt-packet/pull/145

PR #1732, opened 2023-11-08
feat(browser): websockets improvements and bundle optimizations

### LINKED ISSUE #876: WebSocket connection errors not catched
When connecting to an invalid url using WebSocket, the error is not dispatched to 'error' event function (`client.on('error')`).

For example, `mqtt.connect('ws://invalid.hostname/')` output in console:
```
WebSocket connection to 'ws://invalid.hostname/' failed: Error in connection establishment: net::ERR_NAME_NOT_RESOLVED
```

However, this event is impossible to catch, including with `try...catch` statements.

### LINKED ISSUE #1647: Try replace `duplexify` with nodejs `stream.duplex`
Suggested by @mcollina:

https://nodejs.org/api/stream.html#streamduplexfromsrc

robertsLando:
- Create an util class `BufferedDuplex` to replace `duplexify` on `ws` `ali` and `wx`
- Partially fixes #876 (will keep this for another PR)
- Fix [bug](https://github.com/mqttjs/MQTT.js/pull/1732#discussion_r1387762620) on `ws` `_writev` function
- Fix [bug](https://github.com/mqttjs/MQTT.js/pull/1732/files#r1388055417) on socket write, seems that when option`browserBufferSize` is reached (defaults to 512kB) this not only wasn't blocking the writes to socket but was also causing duplicated messages. For sure there were some open issues caused by this
- Improve types in `ws`. Use `window.WebSocket` for Browser and `ws` module types for Node
- Significatly reduces bundle size by ~24% (90kB)
- Improved browser tests to subscribe to different topics to prevent false positives when running concurrently

Fixes #1647


robertsLando:
@mcollina We actually use `duplexify` in `ws` `wsx` and `ali` protocols. If I understood it correctly, on browsers we do this:

1. Create the websocket `const socket = createBrowserWebSocket(client, opts)`. This is the connection to our broker
2. Create a proxy `const proxy = buildProxy(opts, socketWriteBrowser, socketEndBrowser)`. Proxy is a Transform stream used to ensure data written to socket is a Buffer and not a string (when opts.objectMode is not true).
3. If the socket is open we return the `proxy` as `stream`, if not we create a duplex using duplexify and when the socket is open we set proxy as duplex readable and writeable. The only reason I found to do this is that duplexify, when no readable and writeable are set, buffers all writes and flush them when `setWritable` is called.

If all my suppositions are correct I would like to know how in your opinion I could use readable-stream `Duplex` to replace duplexify. In my idea I could simply create a util class like:


```ts
export class BufferedDuplex extends Duplex {
	public socket: WebSocket

	public proxy: Transform

	public eventListenerSupport: boolean

	public socketOpen: boolean

	private isReadyPromise: Promise<void>

	private resolveReady: () => void

	constructor(opts: IClientOptions, proxy: Transform, socket: WebSocket) {
		super({
			objectMode: true,
		})
		this.proxy = proxy
		this.socket = socket
		this.eventListenerSupport =
			typeof socket.addEventListener !== 'undefined'

		if (!opts.objectMode) {
			this._writev = writev
		}

		this.socketOpen = false

		this.isReadyPromise = new Promise((resolve) => {
			this.resolveReady = resolve
		})

		function onOpen() {
			this.socketOpen = true
			this.resolveReady()
			this.emit('connect')
		}

		if (this.eventListenerSupport) {
			socket.addEventListener('open', onOpen)
		} else {
			socket.onopen = onOpen
		}

		this.on('end', () => {
			this.proxy.end()
		})
	}

	_read(size?: number): void {
		this.proxy.read(size)
	}

	async _write(chunk: any, encoding: string, cb: (err?: Error) => void) {
		if (!this.socketOpen) {
			// wait for socket to open
			await this.isReadyPromise
		}

		this.proxy.write(chunk, encoding, cb)
	}

	_final(callback: (error?: Error) => void): void {
		this.proxy.end(callback)
	}
}
```

src/lib/connect/ws.ts
```diff
@@ -220,41 +224,27 @@ const browserStreamBuilder: StreamBuilder = (client, opts) => {
...
+		else data = Buffer.from(data as string, 'utf8')
 		proxy.push(data)
 	}
 
-	// this is to be enabled only if objectMode is false
-	function writev(chunks: any, cb: (err?: Error) => void) {
-		const buffers = new Array(chunks.length)
-		for (let i = 0; i < chunks.length; i++) {
-			if (typeof chunks[i].chunk === 'string') {
-				buffers[i] = Buffer.from(chunks[i], 'utf8')
```
robertsLando, line 249: I also noticed this should have been `chunks[i].chunk`. Wondering how this cold ever be working before. Corrected [here](https://github.com/mqttjs/MQTT.js/pull/1732/files#diff-c807a1a5cf80c9f0c9fd037f1173b6e0beefe3672e805ff1392ef3be225858c7R15)

src/lib/connect/ws.ts
```diff
@@ -1,12 +1,15 @@
 import { StreamBuilder } from '../shared'
-
 import { Buffer } from 'buffer'
-import WS, { ClientOptions } from 'ws'
+import WebSocket, {
+	ClientOptions,
+	ErrorEvent,
+	MessageEvent,
+} from 'isomorphic-ws'
```
robertsLando, line 7: This is a nice package that automatically uses ws on node and global.Websocket on browser. This makes types more clear

robertsLando:
cc @vishnureddy17 if you also would like to give a look at this your feedback would be welcome :)

src/lib/connect/ws.ts
```diff
@@ -220,41 +224,27 @@ const browserStreamBuilder: StreamBuilder = (client, opts) => {
...
-				buffers[i] = Buffer.from(chunks[i], 'utf8')
-			} else {
-				buffers[i] = chunks[i].chunk
-			}
-		}
-
-		this._write(Buffer.concat(buffers), 'binary', cb)
-	}
-
 	function socketWriteBrowser(
```
robertsLando, line 248: Wondering if there should be a return after setTimeout [here](https://github.com/mqttjs/MQTT.js/pull/1732/files#diff-cc6fb47ee82c471804f56a0629299e3e03381c08ce37d2b3053c259badf7fa64L265) ?

src/lib/connect/ws.ts
```diff
@@ -166,11 +169,15 @@ const browserStreamBuilder: StreamBuilder = (client, opts) => {
...
+	// the websocket connection
 	const socket = createBrowserWebSocket(client, opts)
+
+	// the proxy is a transform stream that forwards data to the socket
+	// it ensures that data write to socket is a buffer
 	const proxy = buildProxy(opts, socketWriteBrowser, socketEndBrowser)
 
 	if (!opts.objectMode) {
-		proxy._writev = writev
+		proxy._writev = writev.bind(proxy)
```
robertsLando, line 180: Wondering how this could have work before without the `bind`, maybe Transform automatically handles the bind the `_writev` is set?

vishnureddy17:
I like the changes. Unfortunately, my knowledge on Streams in JS is poor, so I don't think my review will be of much use.

I share your concerns regarding:
- lack of `bind` on `writev` before
- the `chunks[i].chunk` issue
- return after the `setTimeout` in `socketWriteBrowser`

src/lib/connect/ws.ts
```diff
@@ -263,13 +257,15 @@ const browserStreamBuilder: StreamBuilder = (client, opts) => {
 		if (socket.bufferedAmount > bufferSize) {
 			// throttle data until buffered amount is reduced.
 			setTimeout(socketWriteBrowser, bufferTimeout, chunk, enc, next)
+			return
```
robertsLando, line 260: Wondering if there should be a return after setTimeout [here](https://github.com/mqttjs/MQTT.js/pull/1732/files#diff-cc6fb47ee82c471804f56a0629299e3e03381c08ce37d2b3053c259badf7fa64L265) ?

Seems it should: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/bufferedAmount

robertsLando:
Seems that the reason why socket error in browser are not triggering client error is that the error emitted doesn't contain a code that is recognized as a socket error code, in fact it is a generic window [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event), seems that we could get a code in close event but that [will always be 1006](https://stackoverflow.com/a/28396165) (see [reference](https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.1). The fix could be to emit the error with a static code, maybe ECONREFUSED so that will trigger the error on client and close the connection, this will remove the need to call destroy on stream.

getlarge (commented):
As i have no knowledge of the MQTTjs codebase, i only reviewed BufferedDuplex and its usage in ws.ts.
I'll keep on reading ...but here is my firs thought.

src/lib/BufferedDuplex.ts
```diff
@@ -0,0 +1,90 @@
...
+		this.emit('connect')
+		this.isSocketOpen = true
+		this.resolveReady()
+	}
+
+	_read(size?: number): void {
+		this.proxy.read(size)
+	}
+
+	async _write(chunk: any, encoding: string, cb: (err?: Error) => void) {
```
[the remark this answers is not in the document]
robertsLando: That's the first way I implemented it (I mean by using a queue) but preferred this solution as it doesn't require another variable. Not sure if making the method async makes any difference, the underlying code handling stream expects it to be sync so will never await it anyway (also unhandled rejections are impossible)

robertsLando:
@mcollina kindly ping

mcollina (changes requested):

src/lib/BufferedDuplex.ts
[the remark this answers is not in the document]
robertsLando: @mcollina Ok, fixed now 🙏🏼

robertsLando:
@mcollina ping

mcollina (approved):
lgtm

PR #1747, opened 2023-11-29
fix: adjusted global handling to fix web worker environment issue

hibrandonevans:
This pull request addresses the issue reported in #1748 , where our package faced compatibility issues within the Web Worker environment due to incorrect global scope handling. Specifically, the problem was that `global` was getting replaced with `window` during the packaging process.

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -45,7 +45,7 @@ const nextTick = process
 	  }
 
 const setImmediate =
-	global.setImmediate ||
+	globalThis.setImmediate ||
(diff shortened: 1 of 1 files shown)


robertsLando:
@hibrandonevans AFAIK globalThis is just an alias for global, I don't understand how this could fix the problem, did you try to run esbuild to see if this change fixes your problem? I think this should be fixed on esbuild side

robertsLando:
Supersided by https://github.com/mqttjs/MQTT.js/pull/1749

PR #1757, opened 2023-12-05
fix: improve worker tests

robertsLando:


robertsLando:
cc @vishnureddy17 now also web workers are covered by tests 🎉

Issue #1756, opened 2023-12-05
Backport #1752 to 4.x stream


hardillb:
Thank you for rapid response to #1751.

Would you accept a PR to back port this to v4.3.7? I have a unfortunate requirement to support NodeJS 14 (long story about embedded hardware) for a bit longer but could really do with being able to see all the errors that are getting swallowed in the current implementation.

Thanks.

hardillb:
@robertsLando if you can create a local branch based on tag v4.3.7 I can open a PR against it.

https://github.com/hardillb/MQTT.js/tree/4.3.7-backport-1752

Thanks.

robertsLando:
I created `v4` branch

hardillb:
closed via #1758

PR #1752, opened 2023-11-30
fix: catch all socket errors

robertsLando:
Fixes #1751

PR #1758, opened 2023-12-05
Backport #1752

hardillb:
Backport to version that supports NodeJS 14

As discussed in https://github.com/mqttjs/MQTT.js/issues/1756

Here is a backport of the #1752 fix to the v4 codebase

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -32,14 +32,6 @@ const defaultConnectOptions = {
   resubscribe: true
 }
 
-const socketErrors = [
-  'ECONNREFUSED',
-  'EADDRINUSE',
-  'ECONNRESET',
-  'ENOTFOUND'
-]
-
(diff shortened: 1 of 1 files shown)


robertsLando (approved):

robertsLando:
https://www.npmjs.com/package/mqtt/v/4.3.8

robertsLando:
Out of curiosity, is this for node-red?

hardillb:
This is for the day job, the [FlowFuse](https://flowfuse.com/) Device Agent which is a tool for managing Node-RED instances on Edge Devices. We use MQTT as the C&C pipeline for pushing flows to the Node-RED instance and reporting state as well as a way for the flows to communicate between instances and the cloud. We also run a HTTP Proxy tunnel over the MQTT link for accessing the Node-RED editor.

PR #1780, opened 2024-01-23
fix: ignore packets after `end()`

robertsLando:
Ref #846


robertsLando:
I think this is just wrong as in case there are packets in outgoing store them would be discarded

PR #1808, opened 2024-03-04
fix: some others flaky tests

robertsLando:


robertsLando:
@mcollina I recently started to randomly get this error on tests: https://github.com/mqttjs/MQTT.js/actions/runs/8143942807/job/22256983402?pr=1808#step:6:188

>       'test did not finish before its parent and was cancelled'

robertsLando:
@mcollina It seems this happens very randomly, initially I thought that it was caused by a leaked handler but I have been able to reproduce it locally by running tests on a single file and it  fails after few seconds (when it happens). It makes me thinking about a possible race condition in tests but I have no clue where to look/debug

mcollina (approved):
lgtm

robertsLando:
@mcollina the issue is not fixed btw, I still get that error randomly...

mcollina:
The tests are not properly isolated due to them running against the same broker and in the same process. porting them to tap or node:test would fix the latter, and a major refactoring would fix the former. Otherwise this is a goose chase.

robertsLando:
> The tests are not properly isolated due to them running against the same broker and in the same process. porting them to tap or node:test would fix the latter

1. Each test is using a different broker on a different port
2. Tests are already using `node:test` suite

robertsLando:
@mcollina Ping

robertsLando:
@mcollina I will merge this for now as it makes some improvements I wish to include on next release, if you find some time to check the code and find a possible reason why that error happens let me know 🙏🏼

Issue #1806, opened 2024-02-29
[feat]: Add option to choose between worker-timers and native timers


olso-nordsec:
### MQTTjs Version

5.3.6

### Broker

rabbitmq with cowboy

### Environment

Browser

Firefox 123.0, running inside MV3 browser extension (service worker -> the one without `window`)

### Description

I think it got broken by this https://github.com/mqttjs/MQTT.js/pull/1779

Our server closes the connection after 60s, because there are no pings

Then new connection is opened

Our keep alive is set to 30s

We've always used 30s

### Minimal Reproduction

```ts
 this.mqttClient = mqtt.connect(credentials.endpoint, {
      log: (...args) => console.log('[mqtt]', new Date().toISOString(), ...args),
      protocolId: 'MQIsdp',
      protocolVersion: 3,
      path: '/mqtt',
      clean: true,
      resubscribe: true,
      connectTimeout: 15 * 1000,
      reconnectPeriod: 30 * 1000,
      keepalive: 30,
      transformWsUrl: (url, options): string => {
        /* eslint-disable no-param-reassign */
        options.clientId = this.credentials.ncAppUserUid;
        options.username = this.credentials.username;
        options.password = this.credentials.password;
        /* eslint-enable no-param-reassign */

        return this.credentials.endpoint;
      },
    });
```

### Debug logs

```txt
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.protocol wss instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.protocolVersion 3 instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.username undefined instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.keepalive 30 instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.reconnectPeriod 30000 instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.rejectUnauthorized undefined instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: options.properties.topicAliasMaximum undefined instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: clientId mqttjs_cb55d853 instrument.js:117
[mqtt] 2024-02-29T10:23:40.642Z MqttClient :: setting up stream instrument.js:117
[...]

PR #1818, opened 2024-03-15
feat: `timerVariant` option to choose between native and worker timers

olso-nordsec:
Closes https://github.com/mqttjs/MQTT.js/issues/1806


src/lib/PingTimer.ts
```diff
@@ -1,28 +1,36 @@
...
 		this.keepalive = keepalive * 1000
 		this.checkPing = checkPing
+		this.variant = variant
 		this.reschedule()
 	}
 
 	clear() {
 		if (this.timer) {
-			timers.clear(this.timer)
+			getTimer(this.variant).clear(this.timer)
```
robertsLando, line 26: @olso-nordsec Instead of calling `getTimer` everytime here I would set a private property `timer` on constructor with the value of getTimer based on the variant passed (the varian may be removed from pingtimer props then)

robertsLando (changes requested):

olso-nordsec:
@robertsLando what about the codecov issue, `worker` case wont work in node environment 🤔 

do you have a suggestion how to test it? or just dummy test case

robertsLando:
@olso-nordsec unfortunately codecov doesn't consider the coverage of tests running on browser (with `wrt`, check `test:browser` script ), ignore that no worries. I think there is also a way to tell codecov to ignore some lines from coverage but I never digged into that

src/lib/PingTimer.ts
olso-nordsec: changed

robertsLando:
@olso-nordsec FYI: https://github.com/mqttjs/MQTT.js/pull/1819

robertsLando (approved):
LGTM

robertsLando (changes requested):
Tests are failing now, fix them (I think last change did something wrong)

In order to only trigger pingtimer tests you can use `node -r esbuild-register --test test/pingTimer.ts`

robertsLando:
@olso-nordsec The problem is that test rely on `timer` private var that now is named `timerId` so you should replace `pingTimer['timer']` with `pingTimer['timerId']` in `test/pingTimer.ts`

robertsLando:
@olso-nordsec news here?

olso-nordsec:
@robertsLando it was weekend, will look at it today 🙏

robertsLando:
@olso-nordsec Yeah no worries, thanks 🙏🏼

robertsLando (approved):
LGTM. Thanks

PR #1842, opened 2024-04-15
docs: react native example

MaximoLiberata:


robertsLando (changes requested):
Could you also update the TOC? There should be an action on vscode to do that automatically

MaximoLiberata:
> Could you also update the TOC? There should be an action on vscode to do that automatically

Could you explain me a little more about that? I don't know how i can do that

robertsLando:
You can do it manually or use this vscode plugin named "Markdown all in one":

MaximoLiberata:
I updated the TOC

robertsLando (approved):

PR #1855, opened 2024-04-30
fix: keepalive issues

### LINKED ISSUE #1727: After mqtt client is reconnected, it is unable to continue publishing message
Hi there,

I am using the latest version of mqttjs (5.1.3) and this issue happens in old versions as well.

In my app, I received the offline event even when the mosquitto MQTT broker is up and running. After the mqttjs client goes through the process of offline -> closed -> reconnect -> connected. The mqttjs client can no longer publish messages again.

Is there a workaround?

app log:
image

mqtt broker log:
image

robertsLando:
Actually keepalive is shifted every time a packet is written to stream. This is wrong as this doesn't mean the stream is up and in cases where there is a constant publish the keepalive never triggers as is keeps being shifted without closing the client.

With this PR we ensure that the keep alive timer is shifted only when we receive control packets as defined in specs: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Ref363645900

Fixes #1727

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -2106,8 +2106,16 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.options.keepalive &&
 			this.options.reschedulePings
 		) {
-			this.pingTimer.reschedule()
+			this._reschedulePing()
 		}
+	}
+
+	/**
+	 * Mostly needed for test purposes
+	 */
+	private _reschedulePing() {
+		this.log('_reschedulePing :: rescheduling ping')
+		this.pingTimer.reschedule()
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/lib/client.ts
```diff
@@ -1924,9 +1918,6 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 
 		this.emit('packetsend', packet)
 
-		// When writing a packet, reschedule the ping timer
-		this._shiftPingInterval()
```
robertsLando, line 1928: The root cause of the problem

mcollina (dismissed):
lgtm

PR #1856, opened 2024-05-01
fix(electron): detect electron context

axi92:
In the electron context without context isolation all protocols work like in node.
This is tested in the main as well as in the render context.

This PR evolved from that discussion: https://github.com/mqttjs/MQTT.js/discussions/1854#discussion-6585043

### DIFF
--- a/src/lib/is-browser.ts
+++ b/src/lib/is-browser.ts
@@ -1,5 +1,22 @@
-const isStandardBrowserEnv = () =>
-	typeof window !== 'undefined' && typeof window.document !== 'undefined'
+const isStandardBrowserEnv = () => {
+	if (typeof window !== 'undefined') {
+		return (
+			window.navigator.userAgent.toLowerCase().indexOf(' electron/') >
+				-1 &&
+			Object.prototype.hasOwnProperty.call(process.versions, 'electron')
+		)
+		// const userAgent = window.navigator.userAgent.toLowerCase()
+		// if (
+		// 	userAgent.indexOf(' electron/') > -1 ||
+		// 	Object.prototype.hasOwnProperty.call(process.versions, 'electron')
+		// ) {
+		// 	return false
+		// }
+	}
+	return (
+		typeof window !== 'undefined' && typeof window.document !== 'undefined'
+	)
+}
 
 const isWebWorkerEnv = () =>
 	Boolean(
--- a/src/lib/is-browser.ts
+++ b/src/lib/is-browser.ts
@@ -1,16 +1,28 @@
 const isStandardBrowserEnv = () => {
+	// window is only defined when it is a browser
 	if (typeof window !== 'undefined') {
-		if (
-			navigator.userAgent.toLowerCase().indexOf(' electron/') > -1 ||
-			Object.prototype.hasOwnProperty.call(process.versions, 'electron')
-		) {
+		// Is the process an electron application
+		const electronMainCheck = Object.prototype.hasOwnProperty.call(
+			process.versions,
+			'electron',
+		)
+		// In case of electron the userAgent contains a string formated like: 'Electron/<version>'
+		// we can search for that to detect if it is an electron application
+		const electronRenderCheck =
+			navigator.userAgent.toLowerCase().indexOf(' electron/') > -1
+		if (electronMainCheck && electronRenderCheck) {
+			// Both electron checks are only true if the following webPreferences are set in the main electron BrowserWindow()
+			//   webPreferences: {
+			//     sandbox: false,
+			//     nodeIntegration: true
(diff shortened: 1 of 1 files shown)


axi92:
Does somebody got an idea why the commit message is not compliant?
It does follow this schema: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope
Or did I miss something?

src/lib/is-browser.ts
```diff
@@ -1,5 +1,17 @@
-const isStandardBrowserEnv = () =>
-	typeof window !== 'undefined' && typeof window.document !== 'undefined'
+const isStandardBrowserEnv = () => {
+	if (typeof window !== 'undefined') {
+		if (
+			navigator.userAgent.toLowerCase().indexOf(' electron/') > -1 &&
+			Object.prototype.hasOwnProperty.call(process.versions, 'electron')
```
robertsLando, line 5: I think there should be an `||` here and not an `&&` ?

robertsLando (changes requested):

src/lib/is-browser.ts
robertsLando: Also I'm not sure it is correct to return false in case `navigator.userAgent.toLowerCase().indexOf(' electron/') > -1`.

Based on https://github.com/electron/electron/issues/2288#issuecomment-611231970 seems that means we are in the browser env (rendered script), but `process.versions.hasOwnProperty('electron')` means we are in nodejs env (main script)

robertsLando: @axi92 I would like to know here how to detect when context isolation is true, maybe it was correct to keep the `&&` now that I think about this because seems this happens only when there is no context isolation and so the isBrowser returns true because window is defined... I think reverting back to `&&` is the correct way. Please add a comment that explains this above the code you added

src/lib/is-browser.ts
```diff
@@ -1,5 +1,17 @@
-const isStandardBrowserEnv = () =>
-	typeof window !== 'undefined' && typeof window.document !== 'undefined'
+const isStandardBrowserEnv = () => {
+	if (typeof window !== 'undefined') {
+		if (
+			navigator.userAgent.toLowerCase().indexOf(' electron/') > -1 ||
+			Object.prototype.hasOwnProperty.call(process.versions, 'electron')
+		) {
+			return false
+		}
```
robertsLando, line 8: @axi92 Maybe it was correct to keep the `&&` now that I think about this again, because seems this happens only when there is no context isolation: in that case the isBrowser returns true because window is defined but there is also nodejs available.

 I think reverting back to `&&` is the correct way. Please add a comment that explains this above the code you added to make it clear to everyone looking at this.

robertsLando (changes requested):

src/lib/is-browser.ts
axi92: When context isolation is true electron cant use require in the render.js so there is no way to load the mqtt lib, or am I wrong?
I checked different combinations for the `webPreferences` but only one configuration is working:
```
        webPreferences: {
            sandbox: false,
            nodeIntegration: true,
            contextIsolation: false, 
        },
```
If this config is not set in that way I can't use `require()` in the render.js.
In main.js it is always working because that is always node context.

src/lib/is-browser.ts
```diff
@@ -1,5 +1,29 @@
...
+			navigator.userAgent.toLowerCase().indexOf(' electron/') > -1
+		if (electronMainCheck && electronRenderCheck) {
+			// Both electron checks are only true if the following webPreferences are set in the main electron BrowserWindow()
+			//   webPreferences: {
+			//     sandbox: false,
+			//     nodeIntegration: true
+			//     contextIsolation: false
+			// }
+			return false
+		}
```
robertsLando, line 21:
```suggestion
		// check if we are in electron `renderer`
		const electronRenderCheck =
			navigator?.userAgent?.toLowerCase().indexOf(' electron/') > -1
		if (electronRendererCheck && process?.versions) {
				const electronMainCheck = Object.prototype.hasOwnProperty.call(
					process.versions,
					'electron',
				)
			// Both electron checks are only true if the following webPreferences are set in the main electron BrowserWindow()
			//   webPreferences: {
			//     sandbox: false,
			//     nodeIntegration: true
			//     contextIsolation: false
			// }
			return !electronMainCheck
		}
```

robertsLando (changes requested):

robertsLando:
@axi92 fix lint issues `npm run lint-fix` and there is also a typo mybe

robertsLando (approved):

GuoSirius:
browser environment check  #1877

PR #1865, opened 2024-05-13
feat: keepalive manager

### LINKED ISSUE #1859: [feat] Keepalive improvements
Actually keepalive works this way:

A timer is setup, when timer is elapsed it trigger a function to check if last received control message was more then our keepalive time, if ok it schedule a new check and sends a pingreq otherwise closes the connection and triggers a reconnect. Every time we receive a control message we shift the timer.

Specs say: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Figure_3.5_Keep

> If the Keep Alive value is non-zero and the Server does not receive a Control Packet from the Client within one and a half times the Keep Alive time period, it MUST disconnect the Network Connection to the Client as if the network had failed [MQTT-3.1.2-24].

In order to improve the actual implementation my idea is:

1. Create a single timer that triggers every keepalive / 3 seconds
2. Every time a packet is RECEIVED we shift the timer
3. Every time the timer triggers we increese a counter, when it reaches value 3 it triggers the close and reconnect. When the counter reaches value 2 we send a pingreq packet
4. The counter is resetted on every shift

In this way we would be sure that the connection is closed exaclty after 1.5*keepalive seconds without any packet

cc @mcollina opinions/suggestions on this?

robertsLando:
Fixes #1859


src/lib/KeepaliveManager.ts
```diff
@@ -0,0 +1,89 @@
...
+	get keepaliveTimeoutTimestamp() {
+		return this._keepaliveTimeoutTimestamp
+	}
+
+	set keepalive(value: number) {
+		if (
+			// eslint-disable-next-line no-restricted-globals
+			isNaN(value) ||
+			value < 0 ||
+			value * 1000 > 2147483647
```
Apollon77, line 30: whyh not multiply with 1000 bevore using it ... now you multiply 2 times

src/lib/KeepaliveManager.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	reschedule() {
+		if (this.destroyed) {
+			return
+		}
+
+		this.clear()
+		this.counter = 0
+		// https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Figure_3.5_Keep
+		this._keepaliveTimeoutTimestamp = Date.now() + this._keepalive * 1.5
```
Apollon77, line 71: Math.ceil ? on keepalive multiply result?

src/lib/KeepaliveManager.ts
```diff
@@ -0,0 +1,89 @@
...
+
+			this.counter += 1
+
+			// after keepalive seconds, send a pingreq
+			if (this.counter === 2) {
+				this.client.sendPing()
+			} else if (this.counter > 2) {
+				this.client.onKeepaliveTimeout()
+			}
+		}, this._keepalive / 2)
```
Apollon77, line 87: floor/ceil?

src/lib/KeepaliveManager.ts
```diff
@@ -0,0 +1,89 @@
...
+	/** Timestamp of next keepalive timeout */
+	get keepaliveTimeoutTimestamp() {
+		return this._keepaliveTimeoutTimestamp
+	}
+
+	set keepalive(value: number) {
+		if (
+			// eslint-disable-next-line no-restricted-globals
+			isNaN(value) ||
+			value < 0 ||
```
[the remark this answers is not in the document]
robertsLando: 0 is a valid keepalive in mqtt protocol, it means no keepalive. Also keepalive must be in seconds so it should be an integer not a float, I could add a check for this

src/lib/KeepaliveManager.ts
robertsLando: because I don't want to set it before validating it, the alternative is to write `value = value * 1000` on the above line, I can do that

src/lib/KeepaliveManager.ts
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		this.clear()
+		this.counter = 0
+
+		// https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Figure_3.5_Keep
+		const keepAliveTimeout = Math.ceil(this._keepalive * 1.5)
+
+		this._keepaliveTimeoutTimestamp = Date.now() + keepAliveTimeout
+		this._intervalEvery = this._keepalive / 2
```
Apollon77, line 89: Math.ceil? ;)

robertsLando: Wops :)

PR #1874, opened 2024-05-24
feat: add `unixSocket` option and `+unix` suffix support to protocol

### LINKED ISSUE #1873: [Bug]: Broker URL path is dropped in `connect()`, connections fail
### MQTTjs Version

5.5.0

### Broker

Eclipse, Hive, MQTTHQ, EMQX

### Environment

Browser

### Description

A release after 5.3.5 strips paths from broker URLs.

For example, the `/mqtt` in these broker URLs is dropped and the socket connection is made to the root path `/`:
`wss://mqtt.eclipseprojects.io/mqtt`
`wss://broker.emqx.io:8084/mqtt`
`wss://broker.hivemq.com:8884/mqtt`
`wss://public.mqtthq.com:8084/mqtt`

These connections fail as the path is needed.

### Minimal Reproduction

`mqtt.connect(url)` where `url` contains a path, e.g.:

`wss://mqtt.eclipseprojects.io/mqtt`
`wss://broker.emqx.io:8084/mqtt`
`wss://broker.hivemq.com:8884/mqtt`
`wss://public.mqtthq.com:8084/mqtt`

### Debug logs

`mqtt.esm.js:5 WebSocket connection to 'wss://mqtt.eclipseprojects.io/' failed`

robertsLando:
Fixes #1873

### DIFF
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -96,6 +96,7 @@ function connect(
 
 		opts = { ...parsedOptions, ...opts }
 
+		// when parsing an url expect the protocol to be set
 		if (!opts.protocol) {
 			throw new Error('Missing protocol')
 		}
@@ -150,6 +151,9 @@ function connect(
 
 	if (!protocols[opts.protocol]) {
 		const isSecure = ['mqtts', 'wss'].indexOf(opts.protocol) !== -1
+		// returns the first available protocol based on available protocols (that depends on environment)
+		// if no protocol is specified this will return mqtt on node and ws on browser
+		// if secure it will return mqtts on node and wss on browser
 		opts.protocol = [
 			'mqtt',
 			'mqtts',


ytimenkov:
I finally managed to look a bit at url parsing...

I think it is broken in nodejs: https://github.com/nodejs/node-v0.x-archive/commit/a94ffdaec57d9999485b11183f8da10aec15ff8d#diff-3755baff71802c66753e9be67b9b7c0e219dd0b61fda072a0009f807a3
Master: https://github.com/nodejs/node/blob/main/lib/url.js#L430

It hallucinates (trendy word now) a path part of URL for ipv6 addresses.

```
> url.parse("mqtt://localhost")
Url {
  protocol: 'mqtt:',
  slashes: true,
  auth: null,
  host: 'localhost',
  port: null,
  hostname: 'localhost',
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: 'mqtt://localhost'
}
> url.parse("mqtt://127.0.0.1")
Url {
  protocol: 'mqtt:',
  slashes: true,
  auth: null,
  host: '127.0.0.1',
  port: null,
  hostname: '127.0.0.1',
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: 'mqtt://127.0.0.1'
}
url.parse("mqtt://[::1]")
Url {
  protocol: 'mqtt:',
  slashes: true,
  auth: null,
  host: '[::1]',
  port: null,
  hostname: '::1',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'mqtt://[::1]/'
}
```

So I agree that stripping `path` for non-web-socket urls (`ws://`, `wss:/`) makes sense.

But I would still be very cautious about what is sent to `createConnection` function... The validation / simulation of overloads looks very weak to me (https://github.com/nodejs/node/blob/main/lib/net.js#L287): Any string parameter which doesn't convert to number results in Unix address (as if path was specified).
So adding a trailing slash can result in Unix socket being used (e.g. for `mqtt://localhost/`).

Therefore I would move choice between port/host vs path closer to where `createConnection` is called...

robertsLando:
@ytimenkov Thanks for your research, I think so that the only reliable way to fix this is add a boolean option like `unixSocket`  that if true considers the path in tcp connect otherwise it doesn't. What about this?

ytimenkov:
Not sure how boolean option will help.

Curl's option is a string, because you have 2 paths: the first one is address (goes into `Host:` header), and the second one is in the protocol (goes into the request line).

I wonder if curl sets `Host:` to path-address or to `localhost` placeholder... :thinking:. (or to whatever host which was specified in the URL).

This placeholder reminds me about another project Ansible - I think it also recognizes "localhost" and executes tasks in place rather than going via ssh.

This is if you want to provide complete feature-parity between inet(6) and local address for all protocols (kid of hook into the name resolver).

A smaller change would be sacrificing functionality (ws over unix sockt) and use something like "mqtt+unix". This is close to a boolean option you suggested and what is already implemented - for !ws, check if protocol ends with "+unix" and then take either path part or port-host.

With or without that url parsing inconsistency in nodejs, mqtt protocol doesn't care about path, does it? So I think it's good to be a bit more explicit and have clear distinction between these 2 paths, rather than "pass-through everything and hope it works somehow".

How did you see the boolean option?

robertsLando:
> How did you see the boolean option?

I don't like it as I would prefer not adding other options but let's think about the possible cases:

- `mqtt/mqtts://` protocol: should parse the path only if the boolean flag is true (false by default)
- `ws/wss://` protocol: should always parse the path

I think that those cases should cover all the possible conflicts while parsing the url.

> But I would still be very cautious about what is sent to createConnection function... The validation / simulation of overloads looks very weak to me ([nodejs/node@main/lib/net.js#L287](https://github.com/nodejs/node/blob/main/lib/net.js?rgh-link-date=2024-05-27T03%3A43%3A35Z#L287)): Any string parameter which doesn't convert to number results in Unix address (as if path was specified).

About this, this is only valid when calling the create connection with two args, in our case the first argument is an object with all the options set (both for [tcp](https://github.com/mqttjs/MQTT.js/blob/e6da804f513be64fd245f6640a7079325a009174/src/lib/connect/tcp.ts#L19) and [tls](https://github.com/mqttjs/MQTT.js/blob/e6da804f513be64fd245f6640a7079325a009174/src/lib/connect/tls.ts#L27))

Issue #1877, opened 2024-06-04
[Bug]: Electron browser environment check


GuoSirius:
### MQTTjs Version

5.7.0

### Broker

RabbitMQ

### Environment

Electron Renderer & Browser

### Description

Electron Renderer process still exist `electron` keyword, cannot check browser environment

### Minimal Reproduction
![image](https://github.com/mqttjs/MQTT.js/assets/48358523/c91b299c-65a5-4bc5-84ce-bdf0c113db4d)

### Debug logs
![image](https://github.com/mqttjs/MQTT.js/assets/48358523/074dd2f1-14f3-42f4-915e-f27ab2b536c4)

GuoSirius:
> @GuoSirius Could you submit a PR to fix the issue please?

```javascript
if (process.type === 'renderer') {
  console.log('current  Electron Renderer');
} else if (typeof process.electron !== 'undefined') {
  console.log('current Electron Main');
}
```

GuoSirius:
In fact, this test should not be placed here for judgment.

Because the rendering process is also a browser environment, since it is a browser environment, it should go the browser logic.

Otherwise, the ws module may be mistakenly loaded in the browser environment of the rendering process, and the ws module cannot run in the browser environment.

If you do need to use the mqtt protocol in an electron application, it is recommended to put the connection establishment logic in the main process, and then use the interprocess communication mechanism to interact between the row rendering process and the main process.

The rendering process must have a window and a document, but not necessarily a Node.js environment, depending on the configuration at instantiation time.

With the main rendering process's responsibilities clearly divided, the browser environment's detection mechanism can maintain the previous standard judgment logic.

GuoSirius:
The Electron main process owns all Node environments and can establish connections according to the logic of non-browser environments.

The rendering process belongs to the browser environment and should follow the logic of the browser environment to establish the connection.

Just because the rendering process has more process, Node, and other accessible environments than the normal browser environment, it should not be attributed to the Node environment.

In this cross-end application, it is up to the user to decide which way to establish the connection. According to the needs, the connection can be established in the main process or the rendering process, rather than the application library to force detection to determine the protocol.

According to the above convention, the is-browser.ts file is purely used to detect the browser environment, regardless of whether there are other factors, such as the main process or the rendering process.

robertsLando:
[...]

PR #1879, opened 2024-06-04
fix: electron main and renderer process

GuoSirius:
#1877

### DIFF
--- a/src/lib/is-browser.ts
+++ b/src/lib/is-browser.ts
@@ -1,26 +1,14 @@
 const isStandardBrowserEnv = () => {
 	// window is only defined when it is a browser
 	if (typeof window !== 'undefined') {
-		// Is the process an electron application
-		// check if we are in electron `renderer`
-		const electronRenderCheck =
-			typeof navigator !== 'undefined' &&
-			navigator.userAgent?.toLowerCase().indexOf(' electron/') > -1
-		if (electronRenderCheck && process?.versions) {
-			const electronMainCheck = Object.prototype.hasOwnProperty.call(
-				process.versions,
-				'electron',
-			)
-			// Both electron checks are only true if the following webPreferences are set in the main electron BrowserWindow()
-			//   webPreferences: {
-			//     sandbox: false,
-			//     nodeIntegration: true
-			//     contextIsolation: false
-			// }
-			return !electronMainCheck
+		if (typeof process !== 'undefined') {
+			if (process.type === 'renderer') return true
+			else if (typeof process.electron !== 'undefined') return false
 		}
+
 		return typeof window.document !== 'undefined'
 	}
+
 	// return false if nothing is detected
 	return false
 }


robertsLando:
@aki92 could you take a look at this to be sure it doesn't break you? 

Ref: https://github.com/mqttjs/MQTT.js/pull/1856

GuoSirius:
https://www.electronjs.org/docs/latest/api/process#processtype-readonly

image

image

axi92:
> @axi92 could you take a look at this to be sure it doesn't break you?
> 
> Ref: #1856

I can check later today. Thanks for mentioning me!

axi92:
`npm run build` returns errors.
@GuoSirius  were you able to build your code?

GuoSirius:
It's all right now

robertsLando (commented):
Fix lint issues: `npm run lint-fix`

axi92:
I checked the current branch and it breaks for me.
It uses ws instead of mqtt in the browser context of electron.

What was your vision of an simple electron app?
An automated test or a test that is manually executed in an environment with a display?

robertsLando:
@axi92 An automated test of course with a dedicated GH workflow

GuoSirius:
In fact, this test should not be placed here for judgment.

Because the rendering process is also a browser environment, since it is a browser environment, it should go the browser logic.

Otherwise, the ws module may be mistakenly loaded in the browser environment of the rendering process, and the ws module cannot run in the browser environment.

If you do need to use the mqtt protocol in an electron application, it is recommended to put the connection establishment logic in the main process, and then use the interprocess communication mechanism to interact between the row rendering process and the main process.

The rendering process must have a window and a document, but not necessarily a Node.js environment, depending on the configuration at instantiation time.

With the main rendering process's responsibilities clearly divided, the browser environment's detection mechanism can maintain the previous standard judgment logic.

robertsLando (changes requested):
Lint is failing, I think you have to exclude electron-tests directory from tsconfig

robertsLando:
> If you do need to use the mqtt protocol in an electron application, it is recommended to put the connection establishment logic in the main process, and then use the interprocess communication mechanism to interact between the row rendering process and the main process.

I'm not an electron user but I think it makes sense. Also mqtt-explorer (an electron application that uses mqtt) seems to do it this way

https://github.com/thomasnordquist/MQTT-Explorer/tree/f539e03c7ebc44ab274159ba370b14fcd269b414?tab=readme-ov-file#develop

axi92:
If you use electron in its best practice mode that is the case, yes I agree with you.

On the other hand if you do ignore that on purpose because you choose to, I don't know if it is a good option to deny the user that option.

The workaround we used was to patch mqtt after `npm i`

robertsLando:
@axi92 I'm not blocking this, I just would like to be sure it works for all scenarios :)

GuoSirius:
The Electron main process owns all Node environments and can establish connections according to the logic of non-browser environments.

The rendering process belongs to the browser environment and should follow the logic of the browser environment to establish the connection.

Just because the rendering process has more process, Node, and other accessible environments than the normal browser environment, it should not be attributed to the Node environment.

In this cross-end application, it is up to the user to decide which way to establish the connection. According to the needs, the connection can be established in the main process or the rendering process, rather than the application library to force detection to determine the protocol.

According to the above convention, the is-browser.ts file is purely used to detect the browser environment, regardless of whether there are other factors, such as the main process or the rendering process.

robertsLando:
@GuoSirius I understand, so is there a way to allow both you and @axi92 being able to use mqtt with electron as you want? I don't want to break anyone I just would like to allow everybody using this lib as they prefer, maybe a way to override the is_browser check could work? with an env var maybe. Giving that your solution seems the most correct one and I agree with you we could use yours to detect the electron rendering process and consider it browser and in order to also allow @axi92 using it like he want add an env var to allow users forcing the value of is_browser. What do you think?

cc @axi92

axi92:
Override is a good option! Either with options or an ENV var as long as it is possible to override the check I am fine with it :+1:

robertsLando:
Ok so @GuoSirius would you like to add the override with the env var to this PR please?

robertsLando:
Could this be closed now that we have `forceNativeWebSocket` option?

ericlogic:
Any updates for this issue?

forceNativeWebSocket seems can't solve this.

I can not connect to broker with tcp or mqtt protocol in a node integrated render process.

ericlogic:
When I specify to use the TPC or MQTT protocol for connection, it automatically adjusts to use the WS protocol instead. It doesn't seem reasonable.

In my opinion, if the caller specifies a protocol, the library should simply use the specified protocol to establish connection. If the current environment doesn't support it, just throw an exception for troubleshooting.

Environment detection may be useful for some users, but this mechanism cannot achieve 100% accuracy. Therefore, I believe that an option is needed to tell the library directly bypass environment detection and work solely in the way specified by caller.

robertsLando:
@ericlogic forceNativeWebsockets is used to force MQTTjs to use browser's Websocket instead of the ws module, your problem seems related due to MQTTjs detecting your env as a browser when it is not, would you like to submit a PR to fix this?

ericlogic:
@robertsLando Yes, you are right.

I did some research and it's a little complicated.

First, isStandardBrowserEnv method isn't work as expected. It is executed when is-browser.ts is imported, at that time, "process" object is an empty object. But it works OK at runtime.

After that, when I walked around the isBrower check, The library throwed an Error.
```
Uncaught Error: Node.js net module is not supported by JSPM core outside of Node.js
    at Object.Me (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:13438)
    at Object.pS (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:13484)
    at t.s [as streamBuilder] (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:13752)
    at t.connect (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:11531)
    at new t (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:11511)
    at Object.Pg (webpack-internal:///./node_modules/mqtt/dist/mqtt.esm.js:13754)
```

The error is throwed in the mqtt.esm.js but I can't find any related code in source code. It seem that there is a lot of work to do.

robertsLando:
That error comes from this line:

https://github.com/mqttjs/MQTT.js/blob/d7553e3dce72c9a74aef44c13bfcfcf5328d8d49/src/lib/connect/index.ts#L146

When isBrowser is false the `[tcp](https://github.com/mqttjs/MQTT.js/blob/d7553e3dce72c9a74aef44c13bfcfcf5328d8d49/src/lib/connect/tcp.ts)` module is loaded and that imports nodejs `net` module that as stated on your error it's not present

Issue #1886, opened 2024-06-11
[Bug]: Suback Error Codes are ignored and interpreted as QoS


tsteinholz:
### MQTTjs Version

5.6.1

### Broker

VerneMQ

### Environment

NodeJS

### Description

# Expeted MQTT Behavior

> The Payload contains a list of Reason Codes. Each Reason Code corresponds to a Topic Filter in the SUBSCRIBE packet being acknowledged. The order of Reason Codes in the SUBACK packet MUST match the order of Topic Filters in the SUBSCRIBE packet [MQTT-3.9.3-1].

|Value|Hex|Reason Code name|Description|
|--- |--- |--- |--- |
|0|0x00|Granted QoS 0|The subscription is accepted and the maximum QoS sent will be QoS 0. This might be a lower QoS than was requested.|
|1|0x01|Granted QoS 1|The subscription is accepted and the maximum QoS sent will be QoS 1. This might be a lower QoS than was requested.|
|2|0x02|Granted QoS 2|The subscription is accepted and any received QoS will be sent to this subscription.|
|128|0x80|Unspecified error|The subscription is not accepted and the Server either does not wish to reveal the reason or none of the other Reason Codes apply.|
|131|0x83|Implementation specific error|The SUBSCRIBE is valid but the Server does not accept it.|
|135|0x87|Not authorized|The Client is not authorized to make this subscription.|
|143|0x8F|Topic Filter invalid|The Topic Filter is correctly formed but is not allowed for this Client.|
|145|0x91|Packet Identifier in use|The specified Packet Identifier is already in use.|
|151|0x97|Quota exceeded|An implementation or administrative imposed limit has been exceeded.|
|158|0x9E|Shared Subscriptions not supported|The Server does not support Shared Subscriptions for this Client.|
|161|0xA1|Subscription Identifiers not supported|The Server does not support Subscription Identifiers; the subscription is not accepted.|
|162|0xA2|Wildcard Subscriptions not supported|The Server does not support Wildcard Subscriptions; the subscription is not accepted.|

> The Server sending a SUBACK packet MUST use one of the Subscribe Reason Codes for each Topic Filter received [MQTT-3.9.3-2].
 
> Non-normative comment
>
> There is always one Reason Code for each Topic Filter in the corresponding SUBSCRIBE packet. If the Reason Code is not specific to a Topic Filters (such as 0x91 (Packet Identifier in use)) it is set for each Topic Filter.

[...]

PR #1887, opened 2024-06-13
fix: suback Error Codes Handling

tsteinholz:
- Add error handling for failed suback requests
  - specified constant the suback return code value
  - copied error object used in other packet handlers, adjusted for this packet
  - added error callback method used in other packet handlers
  - passed error object to the callback method

Now the error creates the expected Exception with detail and is available from all error callbacks, consistent with the rest of the library.

### Subscribing to Unauthorized topic

```
error: Error: Subscribe error: Not authorized
    at handleAck (/Users/tsteinholz/Development/MQTT.js/build/lib/handlers/ack.js:102:27)
    at handle (/Users/tsteinholz/Development/MQTT.js/build/lib/handlers/index.js:36:31)
    at work (/Users/tsteinholz/Development/MQTT.js/build/lib/client.js:221:40)
    at writable._write (/Users/tsteinholz/Development/MQTT.js/build/lib/client.js:246:13)
    at writeOrBuffer (/Users/tsteinholz/Development/MQTT.js/node_modules/readable-stream/lib/internal/streams/writable.js:334:12)
    at _write (/Users/tsteinholz/Development/MQTT.js/node_modules/readable-stream/lib/internal/streams/writable.js:283:10)
    at Writable.write (/Users/tsteinholz/Development/MQTT.js/node_modules/readable-stream/lib/internal/streams/writable.js:286:10)
    at TLSSocket.ondata (node:internal/streams/readable:1007:22)
    at TLSSocket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12) {
  code: 135
}
granted: [
  {
    topic: 'topic/user/does/NOT/have/access/to,
    qos: 1,
    nl: false,
    rap: false,
    rh: 0,
    properties: undefined
  }
]
```

### Subscribing to authorized topic

```
error: undefined
granted: [
  {
    topic: 'topic/user/has/access/to',
    qos: 1,
    nl: false,
    rap: false,
    rh: 0,
    properties: undefined
  }
]
```

### DIFF
--- a/src/lib/handlers/ack.ts
+++ b/src/lib/handlers/ack.ts
@@ -54,7 +54,7 @@ const handleAck: PacketHandler = (client, packet) => {
 	const type = packet.cmd
 	let response = null
 	const cb = client.outgoing[messageId] ? client.outgoing[messageId].cb : null
-	let err
+	let err = null
 
 	// Checking `!cb` happens to work, but it's not technically "correct".
 	//
@@ -129,9 +129,6 @@ const handleAck: PacketHandler = (client, packet) => {
 							delete client['_resubscribeTopics'][topic]
 						})
 					}
-					client['_removeOutgoingAndStoreMessage'](messageId, () => {
-						cb(err, packet)
-					})
 				}
 			}
 			delete client.messageIdToTopic[messageId]
--- a/src/lib/handlers/ack.ts
+++ b/src/lib/handlers/ack.ts
@@ -117,7 +117,11 @@ const handleAck: PacketHandler = (client, packet) => {
 			client.messageIdProvider.deallocate(messageId)
 			const granted = packet.granted as number[]
 			for (let grantedI = 0; grantedI < granted.length; grantedI++) {
-				if ((granted[grantedI] & 0x80) !== 0) {
+				const subackRC = granted[grantedI]
+				if ((subackRC & 0x80) !== 0) {
+					err = new Error(`Subscribe error: ${ReasonCodes[subackRC]}`)
+					err.code = subackRC
+
 					// suback with Failure status
 					const topics = client.messageIdToTopic[messageId]
 					if (topics) {
@@ -125,11 +129,14 @@ const handleAck: PacketHandler = (client, packet) => {
 							delete client['_resubscribeTopics'][topic]
 						})
 					}
+					client['_removeOutgoingAndStoreMessage'](messageId, () => {
+						cb(err, packet)
+					})
 				}
 			}
 			delete client.messageIdToTopic[messageId]
 			client['_invokeStoreProcessingQueue']()
-			cb(null, packet)
+			cb(err, packet)
 			break
 		}
 		case 'unsuback': {
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -3997,7 +3997,7 @@ export default function abstractTest(server, config, ports) {
 						if (subErr) {
 							assert.strictEqual(
 								subErr.message,
-								'Error: Subscribe error: Not authorized',
+								'Subscribe error: Not authorized',
 							)
 							return done(endErr)
 						}
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -2,7 +2,7 @@
  * Testing dependencies
  */
 import { assert } from 'chai'
-import sinon, { SinonSpy } from 'sinon'
+import sinon from 'sinon'
 import fs from 'fs'
 import levelStore from 'mqtt-level-store'
 import Store from '../src/lib/store'
@@ -474,7 +474,7 @@ export default function abstractTest(server, config, ports) {
 			// fake a port
 			const client = connect({ reconnectPeriod: 20, port: 4557 })
 
-			client.on('error', () => { })
+			client.on('error', () => {})
 
 			client.on('offline', () => {
 				client.end(true, done)
@@ -1389,7 +1389,7 @@ export default function abstractTest(server, config, ports) {
 
 		it(
 			'should silently ignore errors thrown by `handleMessage` and return when no callback is passed ' +
-			'into `handlePublish` method',
+				'into `handlePublish` method',
 			function _test(t, done) {
 				const client = connect()
 
@@ -1643,7 +1643,7 @@ export default function abstractTest(server, config, ports) {
 
 		it(
 			'should silently ignore errors thrown by `handleMessage` and return when no callback is passed ' +
-			'into `handlePubrel` method',
+				'into `handlePubrel` method',
 			function _test(t, done) {
 				const store = new Store()
 				const client = connect({ incomingStore: store })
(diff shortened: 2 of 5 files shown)


src/lib/handlers/ack.ts
```diff
@@ -117,19 +117,26 @@ const handleAck: PacketHandler = (client, packet) => {
...
 					// suback with Failure status
 					const topics = client.messageIdToTopic[messageId]
 					if (topics) {
 						topics.forEach((topic) => {
 							delete client['_resubscribeTopics'][topic]
 						})
 					}
+					client['_removeOutgoingAndStoreMessage'](messageId, () => {
+						cb(err, packet)
+					})
```
robertsLando, line 134: In this way cb is called twice once here and once later

robertsLando (changes requested):
Also please add a unit test for this

tsteinholz:
The linter seems to have a problem with passing the error object to the callback

> Error:   132:58  error  Function declared in a loop contains unsafe references to variable(s) 'err'  @typescript-eslint/no-loop-func
https://github.com/mqttjs/MQTT.js/actions/runs/9501359758/job/26187059098?pr=1887

Which might be why this error was not passed in the first place.

I am not sure why this is an error, given error is defined [at the beginning of the method](https://github.com/mqttjs/MQTT.js/blob/main/src/lib/handlers/ack.ts#L57). Do we need to explicitly define this as `null` instead of leaving it implicitly `undefined`?

src/lib/handlers/ack.ts
tsteinholz: removed

robertsLando (changes requested):
Looks good now, please add a test for this and then I will merge

tsteinholz:
@robertsLando It seems like there is test coverage for the immediate subscribe method and the validation of the topic strings. This may be redundant as there is [a return code for the MQTT Broker to do this validation and reject the subscription request as well.
](https://github.com/mqttjs/MQTT.js/blob/main/src/lib/handlers/ack.ts#L29).

I think I found similar tests in `describe('topic validations when subscribing'` to base the suback return code validations on, however, it does not appear there is any ACL enforced on this demo broker.

```js
		it('should return an error (via callbacks) for topic it does not have access to', function _test(t, done) {
			const client = connect()
			client.subscribe('$SYS/#', (subErr) => {
				client.end(true, (endErr) => {
					if (subErr) {
						return done(endErr)
					}
					done(new Error('Suback errors do NOT work'))
				})
			})
		})
```

Since the topic filters are validated on the client side and don't get sent to the broker, I cannot set an invalid topic string to initate an error subnack from the broker. Other possible suback errors are listed here: https://github.com/mqttjs/MQTT.js/issues/1886

What is the recommended way to emulate the suback packet or set an Access Control List (ACL) for the test broker?

```
Packet {
  cmd: 'suback',
  retain: false,
  qos: 0,
  dup: false,
  length: 3,
  topic: null,
  payload: null,
  granted: [ 0 ],
  messageId: 29845
}
2024-06-13T16:06:38.820Z mqttjs connecting to an MQTT broker...
2024-06-13T16:06:38.820Z mqttjs calling streambuilder for mqtt
    ✖ should return an error (via callbacks) for topic it does not have access to (1.584458ms)
      Error: Suback errors do NOT work
          at <anonymous> (/Users/tsteinholz/Development/MQTT.js/test/abstract_client.ts:549:11)
          at <anonymous> (/Users/tsteinholz/Development/MQTT.js/src/lib/client.ts:1454:7)
          at Store.close (/Users/tsteinholz/Development/MQTT.js/src/lib/store.ts:163:4)
          at <anonymous> (/Users/tsteinholz/Development/MQTT.js/src/lib/client.ts:1446:24)
          at Store.close (/Users/tsteinholz/Development/MQTT.js/src/lib/store.ts:163:4)
          at closeStores (/Users/tsteinholz/Development/MQTT.js/src/lib/client.ts:1445:23)
          at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
```

tsteinholz:
I was able to find what I was looking for from the `describe('with alternate server client'` test cases

robertsLando (changes requested):
I think you should also check that the `error` event to be emitted as it was not before

robertsLando (changes requested):
@tsteinholz It's good now just remember to fix lint issues `npm run lint-fix`

test/abstract_client.ts
```diff
@@ -3975,6 +3975,37 @@ export default function abstractTest(server, config, ports) {
...
+						})
+					})
+				})
+
+				client.subscribe('$SYS/#', (subErr) => {
+					client.end(true, (endErr) => {
+						if (subErr) {
+							assert.strictEqual(
+								subErr.message,
+								'Error: Subscribe error: Not authorized',
```
robertsLando, line 4000:
```suggestion
								'Subscribe error: Not authorized',
```

robertsLando (changes requested):

test/abstract_client.ts
robertsLando: The test fails otherwise

tsteinholz: 0kay, I was having trouble running this test locally. Thanks

tsteinholz:
Seems like the new unit test is passing:
* https://github.com/mqttjs/MQTT.js/actions/runs/9548768742/job/26317563967?pr=1887#step:6:390
* https://github.com/mqttjs/MQTT.js/actions/runs/9548768742/job/26317564394?pr=1887#step:6:366

Not exactly sure why the unit test suite are timing out
* https://github.com/mqttjs/MQTT.js/actions/runs/9548768742/job/26317563967?pr=1887#step:6:877
* https://github.com/mqttjs/MQTT.js/actions/runs/9548768742/job/26317564394?pr=1887#step:6:798

Seems like there is some side-effects added in the test suite. Is there a timeout configured in the test and adding the new test is making it take too long to execute?

tsteinholz:
Increased timeout from 60 seconds / 1 minute to 90 seconds / 1 minute and 30 seconds.

robertsLando:
@tsteinholz it's not a timeout problem it's tests that are flacky, could you remove the change to timeout please? When I notice them are failing because of their "flackyness" I sumply re-run them

robertsLando:
@tsteinholz test was wrong, because it was running on both v5 and v3.1 tests and that reason codes are only supported in v5

tsteinholz:
@robertsLando While reason codes are only supported in the v5 specification. The v3 specification still returns an error, it's just always a "Unspecified Error".

The v3 test would just need to be validating that 0x80 returns an "Unspecified Error".

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718068

robertsLando:
@tsteinholz Yeah I know that but if I return 135 in granted the mqtt parser doens't like the packet and rejects it with protocol 3/4 so I simply prefer to run this in mqtt v5 tests

PR #1895, opened 2024-06-27
feat: add compatibility with txiki.js

EmixamPP:
[txiki.js](https://github.com/saghul/txiki.js) is a small JavaScript runtime compatible with most of the browser api; very nice for small embedded devices.

However, `MQTT.js` wrongly assume a `Node.js` compatibility because there is no `document` variable in the global space. Although, if it was using the browser ws, the lib would be compatible with it.

After discussion https://github.com/saghul/txiki.js/issues/557 with the creator of `txiki.js`, we both think that it should be the job to add the compatibility.

That's what I've added in the present PR.

Small remark, I also think that, in the future, it may be good to revisit the way `MQTT.js` determine if it can use native websocket or `Node.js` one through the `ws` module. Like checking if `WebSocket` exists in the global space, instead of trying to determine the runtime name or whatsoever sophisticated tricks.

### DIFF
--- a/src/lib/is-browser.ts
+++ b/src/lib/is-browser.ts
@@ -25,6 +25,8 @@ const isStandardBrowserEnv = () => {
 	return false
 }
 
+const isTxikijsEnv = () => navigator.userAgent === "txiki.js";
+
 const isWebWorkerEnv = () =>
 	Boolean(
 		// eslint-disable-next-line no-restricted-globals
@@ -37,7 +39,7 @@ const isReactNativeEnv = () =>
 	typeof navigator !== 'undefined' && navigator.product === 'ReactNative'
 
 const isBrowser =
-	isStandardBrowserEnv() || isWebWorkerEnv() || isReactNativeEnv()
+	isStandardBrowserEnv() || isWebWorkerEnv() || isReactNativeEnv() || isTxikijsEnv()
 
 export const isWebWorker = isWebWorkerEnv()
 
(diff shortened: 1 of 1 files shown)


robertsLando:
>  Like checking if WebSocket exists in the global space, instead of trying to determine the runtime name or whatsoever sophisticated tricks.

Problem is that NodeJS is introducing Websocket support so when that will happen that check would wrongly assume the env to be node and not browser. I have thought many times how this could be improved but I never found a better way to do this, if you find one please tell me or open a PR :)

robertsLando (commented):
Please fix lint issue `npm run lint-fix`

EmixamPP:
Oops I would need to git force, I committed with my wrong email address.

EmixamPP:
> I have thought many times how this could be improved but I never found a better way to do this, if you find one please tell me or open a PR :)

What the maintainer of txiki.js propose:
> AFAICT navigator.userAgent is the right way to test for this, as recommended by the WinterCG: https://common-min-api.proposal.wintercg.org/
>
> More specifically: https://common-min-api.proposal.wintercg.org/#navigator-useragent-requirements
>
> txikijs, Bun and Deno set it. ~Node doesn't (currently)~. Node does as of version 21.

robertsLando:
@EmixamPP Thanks for your link, problem is that I think we need to postpone this till all our envs supports ` navigator.userAgent`

robertsLando:
@EmixamPP It is ok, what I meant is just that method cannot be used (yet) as I don't think it would support ALL our cases and old frameworks version that may don't have `navigator` support

EmixamPP:
Thanks for your extremely fast review and merge!

Issue #1577, opened 2023-02-10
The mqtt connection will be disconnected after locking the computer screen for 5 minutes. Is there any solution?


jmxiang:
The mqtt connection will be disconnected after locking the computer screen for 5 minutes. Is there any solution?

jmxiang:
The browser label will also be disconnected after being inactive for a period of time

eye-c:
are you setting a keepalive value?
http://www.steves-internet-guide.com/mqtt-keep-alive-by-example/

jmxiang:
> 你在设置保活值吗？ http://www.steves-internet-guide.com/mqtt-keep-alive-by-example/

yes, keepalive: 5,

haoguangchuan:
I also encountered this and tried to set the keepalive parameter, but it didn't work.

By looking through the materials and reading the MQTT source code, I feel like I've found the problem.

When browser tabs are switched to background or minimized, the browser delays or even stops the execution of setInterval. MQTT sends heartbeat packets through the **reinterval** library, which is implemented through setInterval.

When the browser is minimized or the TAB is switched to the background, the heartbeat sent by the MQTT to the server is delayed or stopped, and the abnormal heartbeat causes the MQTT to go offline. You don't reconnect until the page is activated again.

Solution: It is recommended to replace reinterval with **work-timer**, which will avoid the above situation. Even if the browser is switched to the background or minimized, the timer can still run normally.

simonbuchan:
> Solution: It is recommended to replace reinterval with **work-timer**, which will avoid the above situation. Even if the browser is switched to the background or minimized, the timer can still run normally.

I assume you mean `worker-timer`, which currently states:

> Chrome, Firefox and maybe others throttle the frequency at which they invoke those timers to a maximum of once per second in such a situation.

And some googling seems to support this as the situation. That certainly sounds like it should be sufficient rate for a keep-alive!

LiuZiqi-dual:
Does anyone have a perfect solution? It seems that this problem still exists so far

robertsLando:
Duplicate of #1257, please follow updates there

PR #1896, opened 2024-06-28
feat: allow to pass custom timer for keepalive manager

sanduluca:
I was testing the library with react native on an Android and an iOS device. I found out that the timers dont work in background on [Android devices](https://stackoverflow.com/questions/51997866/react-native-setinterval-getting-paused-when-app-is-in-background). On iOS everything seems to be ok.

With this PR we allow users to pass any timer solution. In my case I want to use https://github.com/ocetnik/react-native-background-timer which is working when app is in background too.

#1257 
#1577


robertsLando (changes requested):
Thanks for your PR! Looks good to me! Could you add a very basic test that covers this in keep alive manager tests? Just to be sure this is doesn't get broken by future updates :)

robertsLando (approved):
LGTM

PR #1902, opened 2024-07-16
fix: connect after client.end not working

### LINKED ISSUE #1897: [Bug]: disconnecting flag is not reset after Client.end is finished - Error: client disconnecting
### MQTTjs Version

5.7.3

### Broker

mosquitto

### Environment

NodeJS

### Description

While calling the publish method on a connected client the _Error: client disconnecting_ is thrown by the lib.
The reason for this is that the Client.disconnecting flag is not reset to false after the Client.end method is finished.

Workaround: calling the reconnect method instead of the connect method, cause the Client.disconnecting flag is reset to false there

### Minimal Reproduction

- create a Client with mqtt.connect method using the option manualConnect
- call the Client method connect
- after connect event is triggered, call the Client end method
- after close event is triggered, call the Client connect method
- after connect event is triggered, call the Client publish method
- -> _Error: client disconnecting_ is thrown

### Debug logs

Error: client disconnecting
    at MqttClient._checkDisconnecting (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/client.js:764:36)
    at MqttClient.publish (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/client.js:330:18)
    at MqttClient.<anonymous> (/home/scBot/kaeser-utils/ipc/test.js:29:20)
    at Object.onceWrapper (node:events:628:26)
    at MqttClient.emit (node:events:525:35)
    at MqttClient._onConnect (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/client.js:1078:18)
    at handleConnack (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/handlers/connack.js:39:29)
    at handle (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/handlers/index.js:44:35)
    at work (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/client.js:221:40)
    at Writable.writable._write (/home/scBot/kaeser-utils/node_modules/mqtt/build/lib/client.js:246:13)

MaximoLiberata:
Fixes #1897

### DIFF
--- a/test/client.ts
+++ b/test/client.ts
@@ -615,6 +615,85 @@ describe('MqttClient', () => {
 		},
 	)
 
+	describe('connect manually', () => {
+		it(
+			'should not throw an error when publish after second connect',
+			{
+				timeout: 10000,
+			},
+			function _test(t, done) {
+				let wasConnected = false
+				let closeInterval = false
+
+				const publishInterval = (timeout: number) => {
+					const method = () => {
+						return new Promise<void>((resolve, reject) => {
+							client.publish(
+								'test',
+								'test',
+								{
+									qos: 1,
+								},
+								(err) => {
+									const isError =
+										err?.message.toLocaleLowerCase() ===
+										'client disconnecting'
+
+									assert.isFalse(
+										isError,
+										'client disconnecting',
+									)
+
+									if (isError) {
+										closeInterval = true
+										done()
+									}
+								},
+							)
+						})
+					}
+
+					setTimeout(() => {
+						if (closeInterval === false) {
+							method().then(() => {
+								publishInterval(timeout)
+							})
+						}
+					}, timeout)
+				}
+
+				client = mqtt.connect(config)
+
+				client
+					.on('connect', () => {
+						closeInterval = false
+
+						publishInterval(200)
+
+						setTimeout(() => {
+							closeInterval = true
+
+							if (wasConnected) {
+								done()
+							}
+						}, 900)
+
+						if (wasConnected === false) {
+							setTimeout(() => {
+								wasConnected = true
+								client.end()
+							}, 1300)
+						}
+					})
+					.on('close', () => {
+						setTimeout(() => {
+							client.connect()
+						}, 1500)
+					})
+			},
+		)
+	})
+
 	describe('async methods', () => {
 		it(
 			'connect-subscribe-unsubscribe-end',
--- a/test/client.ts
+++ b/test/client.ts
@@ -621,75 +621,166 @@ describe('MqttClient', () => {
 			{
 				timeout: 10000,
 			},
-			function _test(t, done) {
-				let wasConnected = false
-				let closeInterval = false
+			function _test(t) {
+				return new Promise<void>((resolveParent) => {
+					let countConnects = 0
+					let closeInterval = false
 
-				const publishInterval = (timeout: number) => {
-					const method = () => {
-						return new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


test/client.ts
```diff
@@ -615,6 +615,85 @@ describe('MqttClient', () => {
...
+						}
+					})
+					.on('close', () => {
+						setTimeout(() => {
+							client.connect()
+						}, 1500)
+					})
+			},
+		)
+	})
```
robertsLando, line 695: Could you rewrite this test using fake timers or make it async?

robertsLando (changes requested):

test/client.ts
MaximoLiberata: I moved it into an async method. Check it, please.

MaximoLiberata:
> @MaximoLiberata your change would solve the issue exception issue, but not that after end method finished the public disconnecting flag is still true.

@TimT1919 I think I covered that issue, check this PR. Also, the `disconnecting` variable is only possible to set it as `false` when you are the actor who call the `connect` method (call connect method manually), this happens because the reconnect behavior take a look for `disconnecting` variable before try connect again.

robertsLando (commented):
Sorry @MaximoLiberata I may have explained it wrongly, there is no problem if test is written using callbacks (most of tests are using callbacks), what I wanted to do is to prevent using setTimeout, them are not reliable in tests and also make test last longer (as you have to wait the timeout to trigger. 

I usually fix this by not using timeouts at all but rely on events instead (when possible) or use async/await or use sinon fake timers (you can see some tests using them)

Thanks for your help 🙏🏼

TimT1919:
@MaximoLiberata Integrated your changes to my project and could remove my workaround for this issue. So the changes seam to resolve the issue. Thanks a lot 🙏🏼

MaximoLiberata:
> Sorry @MaximoLiberata I may have explained it wrongly, there is no problem if test is written using callbacks (most of tests are using callbacks), what I wanted to do is to prevent using setTimeout, them are not reliable in tests and also make test last longer (as you have to wait the timeout to trigger.
> 
> I usually fix this by not using timeouts at all but rely on events instead (when possible) or use async/await or use sinon fake timers (you can see some tests using them)
> 
> Thanks for your help 🙏🏼

You are absoluty correct, thanks. Removing setTimeout reduce time and tests won't hang up. Also, I saw if we don't close promises (async/await) that we created in the test, the test will hang up for a moment o permanent.

test/client.ts
```diff
@@ -615,6 +615,167 @@ describe('MqttClient', () => {
...
+			{
+				timeout: 10000,
+			},
+			async function _test(t) {
+				const fail = await new Promise<boolean>((resolveParent) => {
+					let countConnects = 0
+					const clock = useFakeTimers({
+						shouldClearNativeTimers: true,
+						toFake: ['setTimeout'],
+					})
```
robertsLando, line 630: Usually it would be better to do this inside `before()` and `after()` (to put inside the test fn)

MaximoLiberata: ¿Open the clock in before y close it in after?

robertsLando: See example here: 
https://github.com/mqttjs/MQTT.js/blob/cd0b044b89e8413e99e055529d523eac6d4ddcd4/test/abstract_client.ts#L2150-L2156

Sorry only after is needed here no before

robertsLando (approved):

PR #1906, opened 2024-07-19
fix: tets hang up

MaximoLiberata:
Fixes (https://github.com/mqttjs/MQTT.js/issues/1897#issuecomment-2231365999).

When you try to execute the tests maybe one of them will fail because it hanged up and never ends. This PR fix that bug in almost all cases.

### DIFF
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -3202,15 +3202,30 @@ export default function abstractTest(server, config, ports) {
 		})
 
 		it('should emit connack timeout error', function _test(t, done) {
-			const client = connect({
-				connectTimeout: 0,
-				reconnectPeriod: 5000,
+			// Use fake timers to simulate the timeout. The setTimeout inside the client connection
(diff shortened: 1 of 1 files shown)


MaximoLiberata:
@robertsLando try to execute this PR many times and give me feedback if it working right, please.

robertsLando:
Thanks for this PR @MaximoLiberata ! I hope you are right, unfortunately I will not be on my pc till Monday but I will definitely look at this ASAP! I checked the changes and it feels strange only this test was causing the issue, my feel was that there were many of them with this same problem. Did they always succeed now or they just fail less?

robertsLando (approved):
Thanks! 🙏🏼  LGTM

robertsLando:
@MaximoLiberata Tests seems much more reliable now! Thanks again 🙏🏼

MaximoLiberata:
I'm glad to hear! And thank you for checking 🙏🏼.

PR #1907, opened 2024-07-20
wechat/ali also not a browser

shushenghong:


robertsLando:
?

shushenghong:
in wechat environment, `IS_BROWSER` check also return false, so if we want use `wx` `wxs`, should inside this condition

robertsLando:
the change you did makes no sense

Issue #1905, opened 2024-07-19
[Bug]: bundler hard code `navigator` in `mqtt.esm`


EmixamPP:
### MQTTjs Version

5.8.0

### Broker

mosquitto

### Environment

Browser

### Description

The variable `navigator`, used to determine some env, like tkiki.js https://github.com/mqttjs/MQTT.js/pull/1895, is replaced by a content determined at bundle time, namely: `P={deviceMemory:8,hardwareConcurrency:8,language:"en-US"}` if you look at `mqtt/dist/mqtt.esm`. And because of that, it is impossible to determine the runtime using this technique (thus the WebSocket lib choice is wrong).

### Minimal Reproduction

1. Open `mqtt/dist/mqtt.esm`
2. Search for `txiki.js` and see that the code from #1895, `navigator.userAgent==="txiki.js"`  has been transformed in `P.userAgent==="txiki.js"`
3. Search for `P=` and see that the value assigned is `P={deviceMemory:8,hardwareConcurrency:8,language:"en-US"}`
4. Can be confirmed at runtime by adding `console.log(P)&&` before `P.userAgent==="txiki.js"`

### Debug logs

-

robertsLando:
Would you like to submit a PR to fix this issue?

EmixamPP:
After investigation, the issue is because of 
https://github.com/mqttjs/MQTT.js/blob/11bb9bdb4535b7b89a4961f2cdd0f201f91b0092/esbuild.js#L31
from
https://github.com/mqttjs/MQTT.js/blob/11bb9bdb4535b7b89a4961f2cdd0f201f91b0092/esbuild.js#L20-L33

If I remove that line, navigator is untouched, and so it works as expected, but I guess I cannot PR that because of #1789.

robertsLando:
Wondering if we can have a different solution for wechat, maybe if they add a navigator polifilly before importing mqtt it could work?

EmixamPP:
So if I understand correctly, you're not against a PR that removes out-of-the-box support for WeChat?

Here is what I propose:
I could PR something similar to what is proposed in https://github.com/mqttjs/MQTT.js/issues/1900 or the current `timeVariant` option, with a new `forceNativeWebScoket` option, that will by-pass the automatic `IS_BROWSER` call. (Consequently, I will need to move the import of the modules inside the `connect` function).
[...]

Issue #1900, opened 2024-07-10
Remove is-browser.ts detection util in favor of config option


someengineer-nordsec:
**Is your feature request related to a problem? Please describe.**
- It seems this detection adds unnecessary complexity
- There have been https://github.com/mqttjs/MQTT.js/commits/main/src/lib/is-browser.ts quite a few fixes for it, there is lot of edge cases
- Chrome extension detection is bit flaky (some of this pain has been alleviated by https://github.com/mqttjs/MQTT.js/pull/1818)
- https://github.com/mqttjs/MQTT.js/issues/1709

**Describe the solution you'd like**
Configurable by required option

```
mqtt.connect(credentials.endpoint, {
  environment: 'browser', // 'react-native' 'electron' 'node'
})
```

**Describe alternatives you've considered**

**Additional context**

robertsLando:
@olso-nordsec problem is that actually we also need that in order to know which modules to load and which not.

If you have something in mind feel free to open a PR

PR #1910, opened 2024-07-25
feat: add `forceNativeWebSocket` client option

EmixamPP:
Solve #1905 (and could also partially solve #1900).

- add a `forceNativeWebSocket` option to the client, if set to true, by-pass the `isBrowser` detection by [src/lib/is-browser.ts](src/lib/is-browser.ts). This allows to easily solve the `ws does not work in the browser` exception for non-standards env/runtime.
- remove the build option `navigator` of the polyfill `esbuild-plugin-polyfill-node`, since it defines an arbitrary value for the `navigator` variable, breaking many usages at runtime of that variable, like `navigator.userAgent` or `navigator.product` when using the `.esm` dist, which then also break the env detection. Instructions for `WeChat` have been updated.
- Remove auto-detection for `txiki.js`, which is now well-supported if `forceNativeWebSocket` is used (no need to complicate the code for niche use case).
- unify the import name for `IS_BROWSER` to `isBrowser` (does not facilitate the task when debugging browser detection).


README.md
```diff
@@ -454,7 +454,7 @@ The arguments are:
       - `messageExpiryInterval`: value is the lifetime of the Will Message in seconds and is sent as the Publication Expiry Interval when the Server publishes the Will Message `number`,
       - `contentType`: describing the content of the Will Message `string`,
       - `responseTopic`: String which is used as the Topic Name for a response message `string`,
-      - `correlationData`: The Correlation Data is used by the sender of the Request Message to identify which request the Response Message is for when it is received `binary`,
+      - `correlationData`: The Correlation Data is used b`y the sender of the Request Message to identify which request the Response Message is for when it is received `binary`,
```
robertsLando, line 457: typo?

src/lib/connect/index.ts
```diff
@@ -152,6 +133,28 @@ function connect(
...
+	} else {
+		protocols.ws = require('./ws').browserStreamBuilder
+		protocols.wss = require('./ws').browserStreamBuilder
+
+		protocols.wx = require('./wx').default
+		protocols.wxs = require('./wx').default
+
+		protocols.ali = require('./ali').default
+		protocols.alis = require('./ali').default
+	}
```
robertsLando, line 156: I would do this only once. Keep protocols outside the function and put it to null by default, if null load the protocols, skip otherwise

robertsLando (changes requested):
Please remove also `test-store` folder

src/lib/client.ts
```diff
@@ -512,7 +516,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		} else {
 			this.log(
 				'MqttClient :: environment',
-				isBrowser ? 'browser' : 'node',
+				options.forceNativeWebSocket || isBrowser ? 'browser' : 'node',
```
robertsLando, line 519: instead of having this `||` codewise I would make isBrowser a function that accepts client options as argument (the argument can be optional and if not present it will simply return the isBrowser value)

robertsLando (changes requested):

src/lib/connect/index.ts
robertsLando: I don't think there could be any valid reason to change that option when creating different clients, do you?

EmixamPP: No indeed, but this could be confusing to have an object option influencing the global space. I will make sure to document that in the README then.

src/lib/client.ts
EmixamPP: I am wondering then if this make sense that this specific variable define the env globally. I meant, it makes sense to use that specific variable to determine which websocket lib to use, but that's it. Even if, we could then determine the env just based on that choice. Maybe the name of `forceNativeWebSocket` was a bad idea from my side.

Same for:
https://github.com/mqttjs/MQTT.js/blob/6b278dca5a5b82b07835344f3c129ddd5b73e6e8/src/lib/get-timer.ts#L23-L39
`get-timer` also use `isBrowser`, should I also rely on the client's options, namely `forceNativeWebSocket`? I could instead pass to the function all the clients options instead of just the `timerVariant` value.

Btw, if we use `forceNativeWebSocket` inside `isBrowser`, then `timeVariant` may also be used inside `isBrowser`.

So in the end, we could have instead an `environment: "auto" | "browser" | "node" | "react-native | "electron"` option as proposed in #1900, give it to `isBrowser` or nothing for "auto". 
This will give the same result, but using a more generic approach, e.g. we could mark as deprecated `timerVariant`. 
However, this will give less flexibility, because if you are under a specific env, but you do not want the choice imposed by us for the underling libs used; I am mainly thinking about a recent node env that want to use native WebSocket.

--- 
1. Should I generalize to an `environment` option?
2. Should I use `forceNativeWebSocket` *only* for websocket choice related, thus not inside `isBrowser` (and in the snippet of your review)?
3. `environment` + `forceNativeWebSocket` options?
4. Also use `forceNativeWebSocket` to influence `get-timer`, or use `environment`?
5. ?

robertsLando: Giving that we already have an option to allow users specifically choose the timer variant IMO it doens't make so much sense here to use. To avoid confusion we could keep exporting the `isBrowser` const and also export a `shouldUseNativeWs` function with the client options as argument as said before.

Using environment is not a good idea as it doens't work well in browsers so I think your solution works better in this case.

So I would go for solution 2

robertsLando (approved):
LGTM

robertsLando:
@EmixamPP Version 5.9.0 published 🚀

EmixamPP:
Fast! Thank you!

PR #1911, opened 2024-07-26
fix(test): close open connections

MaximoLiberata:
Improve the logic of close open connections when you instance a `serverBuilder` class or open a new connection using `client`. Call method `assert` interrupts the test and the method `after` will never called generating that the connections haven't closed yet will still open and other unit test will throw a error like `Error: listen EADDRINUSE: address already in use`.

This PR also remove the temp folders and files generated by this [`test`](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) when failed.

### DIFF
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -3202,15 +3202,30 @@ export default function abstractTest(server, config, ports) {
 		})
 
 		it('should emit connack timeout error', function _test(t, done) {
-			const client = connect({
-				connectTimeout: 0,
-				reconnectPeriod: 5000,
+			// Use fake timers to simulate the timeout. The setTimeout inside the client connection
+			// will inactive by other tests (maybe) causing this test never ends.
+			const clock = sinon.useFakeTimers({
+				...fakeTimersOptions,
+				toFake: ['setTimeout'],
 			})
 
-			client.on('error', (err) => {
-				assert.equal(err.message, 'connack timeout')
-				client.end(true, done)
+			const connectTimeout = 10
+
+			t.after(() => {
+				clock.restore()
 			})
+
+			const client = connect({
+				connectTimeout,
+				reconnectPeriod: 5000,
+			})
+				.on('connect', () => {
+					clock.tick(connectTimeout)
+				})
+				.on('error', (err) => {
+					assert.equal(err.message, 'connack timeout')
+					client.end(true, done)
+				})
 		})
 
 		it(
--- a/test/abstract_client.ts
+++ b/test/abstract_client.ts
@@ -8,6 +8,7 @@ import levelStore from 'mqtt-level-store'
 import Store from '../src/lib/store'
 import serverBuilder from './server_helpers_for_client_tests'
 import handlePubrel from '../src/lib/handlers/pubrel'
+import CleanMethod from './helpers/clean_method'
 import handle from '../src/lib/handlers/index'
 import handlePublish from '../src/lib/handlers/publish'
 import mqtt, {
@@ -47,55 +48,6 @@ const fakeTimersOptions = {
 	shouldClearNativeTimers: true,
 }
 
-class CleanBeforeMethod {
-	#methods: {
-		method: Promise<any> | ((...args: any[]) => Promise<any>)
-		args: any[]
-	}[]
-
-	constructor() {
-		this.#methods = []
-	}
-
-	add<T extends (...args: any[]) => Promise<any>>(
-		method: Promise<any> | T,
-		...args: Parameters<T>
-	) {
-		this.#methods.push({ method, args })
-	}
-
-	async executeAll(options?: {
-		/**
-		 * If `true`, all methods will be removed after execution.
-		 */
-		removeAll?: boolean
-	}) {
-		if (this.#methods.length === 0) {
-			return
-		}
-
-		const results = await Promise.allSettled(
-			this.#methods.map(({ method, args }) => {
-				if (method instanceof Promise) {
-					return method
-				}
-				return method(...args)
-			}),
-		)
-
-		if (options?.removeAll) {
-			this.#methods = []
-		}
-
-		for (const result of results) {
-			if (result.status === 'rejected') {
-				if (result.reason instanceof Error) throw result.reason
-				else throw new Error(result.reason)
-			}
-		}
-	}
-}
-
 export default function abstractTest(server, config, ports) {
 	const version = config.protocolVersion || 4
 
@@ -107,9 +59,20 @@ export default function abstractTest(server, config, ports) {
 		return mqtt.connect(opts)
 	}
 
-	const cleanBeforeMethod = new CleanBeforeMethod()
-	const beforeEachExec = () =>
-		cleanBeforeMethod.executeAll({ removeAll: true })
+	const cleanMethod = new CleanMethod()
+
+	async function beforeEachExec() {
+		await cleanMethod.closeClientAndServer()
+		await cleanMethod.executeAllMethods()
+	}
+
+	async function afterExec() {
+		await cleanMethod.closeClientAndServer()
+		await cleanMethod.executeAllMethods()
+		cleanMethod.reset()
+	}
+
+	after(afterExec)
 
 	describe('closing', () => {
 		it('should emit close if stream closes', function _test(t, done) {
@@ -655,7 +618,7 @@ export default function abstractTest(server, config, ports) {
 
 	describe('offline messages', () => {
 		beforeEach(beforeEachExec)
-		after(beforeEachExec)
+		after(afterExec)
 
 		it('should queue message until connected', function _test(t, done) {
 			const client = connect()
@@ -701,26 +664,6 @@ export default function abstractTest(server, config, ports) {
 		})
 
 		it('should not interrupt messages', function _test(t, done) {
-			cleanBeforeMethod.add(async () => {
-				if (client) {
-					await new Promise<void>((resolve, reject) => {
-						client.end(true, (err) => {
-							if (err) reject(err)
-							else resolve()
-						})
-					})
-				}
-
-				if (server2?.listening) {
-					await new Promise<void>((resolve, reject) => {
-						server2.close((err) => {
-							if (err) reject(err)
-							else resolve()
-						})
-					})
-				}
-			})
-
 			let client: mqtt.MqttClient | null = null
 			let publishCount = 0
 			const incomingStore = new mqtt.Store({ clean: false })
@@ -764,6 +707,8 @@ export default function abstractTest(server, config, ports) {
 				})
 			})
 
+			cleanMethod.setServer(server2)
+
 			server2.listen(ports.PORTAND50, () => {
 				client = connect({
 					port: ports.PORTAND50,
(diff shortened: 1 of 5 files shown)


MaximoLiberata:
@robertsLando the last test we need to fix to run all test as success all the time is this [`test`](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717). After executing thousand times the test mentioned, sometimes it will fail with always the same error below: 

```
✖ should not overtake the messages stored in the level-db-store
      { message: "expected 'payload3' to equal 'payload2'", showDiff: true, actual: 'payload3', expected: 'payload2', operator: 'strictEqual' }
```

Does it necessary to prioritize the order in this [`test`](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717)?

robertsLando:
I also see many times that a test_store folder remains created and empty after failed tests and it could be because of that... my feel is that for some reasons maybe the store folder is the same as the test are spawned in parallel and this creates the error? Maybe adding a more random test folder name could mitigate the issue?

MaximoLiberata:
Creating unique test folder's names for parallel processes will mitigate the issues related of using resource shared, example: in case of a folder was removed before finishing a test that was using the same folder or file. But the [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) is using `fs.mkdtempSync` and this method creates unique temp folder's names. The test will never generate an issue of remaining files for this reason.

I trust more in the correlational issues:
1. If one test fails the following tests will hang or fail, too.
2. If one test didn't successfully close the opened connections like `client` or `server`, then they will still open and the following tests will hang or fail because they use the same `client` or `server` (maybe the same `port`).

Right now the `after` method only will called when the test finish successfully, otherwise it will never called. To avoid correlational issues I prefer use both `beforeEach` and `after` in `describe` scope, then before each test wait for closing opened connections and deleting temp files.

Also, I don't know if the [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) needs to work in this way publish3 and publish2 separated, and not publish3 inside publish2. Can I move it without affect the logic?

robertsLando (changes requested):

robertsLando:
> Right now the after method only will called when the test finish successfully, otherwise it will never called

I didn't know about that, are you 100% sure of this? I dunno if this should be considered a bug or not on nodejs test suite, I would expext `after` to be called always after a test even when it fails as it is used to cleanup resources. I should check how others work (like jest, mocha, ava etc..)

> Also, I don't know if the [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) needs to work in this way publish3 and publish2 separated, and not publish3 inside publish2. Can I move it without affect the logic?

I didn't wrote that test so sincerly I'm not 100% sure about what was the reason to write it (expecially because I think it should be placed in mqtt-leveldb store package and not here) anyway I would say that it would be more correct and deterministic to put the publish 3 inside publish 2 callback, maybe that's the only missing piece here

MaximoLiberata:
> I didn't know about that, are you 100% sure of this? I dunno if this should be considered a bug or not on nodejs test suite, I would expext after to be called always after a test even when it fails as it is used to cleanup resources. I should check how others work (like jest, mocha, ava etc..)

Yes, but with a trick that I could find. The after method works in normal cases, but in async operations will never called. The following script is a replication of the issue and the same issue that present this [test  ](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717)

```
const fs = require('node:fs')
const { describe, it } = require('node:test')
const { assert } = require('chai') // Used in abstract_client.ts


describe("test", () => {

    it ("should 1=1", (t, done) => {
        t.after(async () => {
            console.log('\n----- AFTER MATHOD FIRED -----\n')

            await new Promise((resolve) => {
                fs.rm(storePath, { recursive: true }, () => {
                    resolve()
                })
            })
        })

        const storePath = fs.mkdtempSync('testing_store_')
        fs.appendFile(storePath + '/test.txt', 'Hello world', () => {
            // Remove this line to pass the test and the temp folder will be removed.
            // It fails, then the temp folder will not be removed
            assert.equal(2, 1)
            done()
        })

    })

})
``` 

> I would say that it would be more correct and deterministic to put the publish 3 inside publish 2 callback, maybe that's the only missing piece here

I think the same, but if there are more tests with a similar logic in their execution like this [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) we take the risk of if one test fail we will have a correlational issue.

robertsLando:
> Yes, but with a trick that I could find. The after method works in normal cases, but in async operations it will never called. The following script is a replication of the issue and the same issue that present this [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717)

I think it could be worth try to open an issue on NodeJS repo. cc @mcollina is this something expected?

> I think the same, but if there are more tests with a similar logic in their execution like this [test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717) we take the risk of if one test fail we will have a correlational issue.

I know but I don't see so many alternatives right now. I think our best bet is to fix one by one as we see errors like what you are doing now

mcollina:
This might be a bug in mocha. I don't have much time to investigate minus saying... don't mix promises and callbacks ;).

robertsLando:
@mcollina We dropped mocha almost one here ago [here](https://github.com/mqttjs/MQTT.js/pull/1697) do you remember? 

We are using nodejs test module now that's why I was asking you if you are aware of issues with it related to `after` inside `async` tests

test/abstract_client.ts
```diff
@@ -47,6 +47,55 @@ const fakeTimersOptions = {
 	shouldClearNativeTimers: true,
 }
 
+class CleanBeforeMethod {
```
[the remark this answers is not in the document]
MaximoLiberata: I moved it into helpers, please check it again.

robertsLando:
@MaximoLiberata Is it just an impression or tests are failing more frequently now then before?

MaximoLiberata:
[reference test](https://github.com/mqttjs/MQTT.js/blob/2da3b3401db1bb880801c7264a2a51dc08d50f62/test/abstract_client.ts#L717)
> Also I think the third publish should be put inside the payload2 publish callback to be 100% sure that the two are sent consecutively

I moved publish3 into publish2, now the tests always pass and the tests that I modified in this PR will not generate correlational issues if one of them fail. The rest tests might generate correlational issues.

MaximoLiberata:
> @MaximoLiberata Is it just an impression or tests are failing more frequently now then before?

Could you give more detail about that? they should not fail

robertsLando:
See https://github.com/mqttjs/MQTT.js/actions/runs/10182861315/job/28166435859?pr=1911

MaximoLiberata:
> See https://github.com/mqttjs/MQTT.js/actions/runs/10182861315/job/28166435859?pr=1911

I didn't clean references of the objects  `client` and `server` after each test.

MaximoLiberata:
Now I pass the references of the objects `client` and `server` using `cleanMethod.setClient(client)` and `cleanMethod.setServer(server)` respectively. These references will be used by the method `cleanMethod.closeClientAndServer()` to close their connection.

But I have a question, those reference can't be used before instance `client` or `server` because their value start in `null` or `undefined` and the `CleanMethod` class will not track the changes of them, so if there is an internal error in their initialization the `CleanMethod` class will not have reference of their new values and will not close those connections.

Can `class` track external change (not inside class by its method) of their attributes with other implementation or not?

cjihrig:
@MaximoLiberata you found a legit bug in the Node test runner - https://github.com/nodejs/node/pull/54151.

robertsLando:
> Can class track external change (not inside class by its method) of their attributes with other implementation or not?

Simply create a `setter` ?

Like 

```ts

set client(value) {
// your logic here
}

File: test/helpers/clean_method.ts [hunk 2]
```diff
@@ -0,0 +1,311 @@
...
+ *         const client = mqtt.connect('mqtt://localhost')
+ *
+ *         cleanMethod.setClient(client)
+ *         done()
+ *     })
+ *
+ * })
+ * ```
+ */
+class CleanMethod {
```
robertsLando, line 101: Could we name this `TeardownHelper`?

test/helpers/clean_method.ts
```diff
@@ -0,0 +1,311 @@
...
+		this.#methods = new Map()
+	}
+
+	/**
+	 * @description
+	 * Set the client to be used as default to close.
+	 */
+	setClient(client: MqttClient | null) {
+		this.#client = client
+	}
```
robertsLando, line 127: could make sense to make this throw in case there was already a client set? Just to be sure we don't call this multiple times

test/helpers/clean_method.ts
```diff
@@ -0,0 +1,311 @@
...
+
+	/**
+	 * @description
+	 * Set the server to be used as default to close.
+	 */
+	setServer(server: ServerBuilderInstance | null) {
+		this.#server = server
+	}
+
```
robertsLando, line 136: Same

test/helpers/clean_method.ts
robertsLando: I would also refactor this to simply add the closeClient method to methods instead

robertsLando (changes requested):

robertsLando:
> @MaximoLiberata you found a legit bug in the Node test runner - [nodejs/node#54151](https://github.com/nodejs/node/pull/54151).

Thanks @cjihrig, and thanks to @mcollina as I think he has asked you to look at this :)

robertsLando:
@MaximoLiberata I'm doing some changes, hold a bit

robertsLando:
@MaximoLiberata Done, check them out then I think we can merge this :)

robertsLando:
arggg: https://github.com/mqttjs/MQTT.js/actions/runs/10200588659/job/28220438706?pr=1911#step:6:337

MaximoLiberata:
@robertsLando I just adjust the class' documentation and thank for checking and improve the code. Also, Thank @mcollina and @cjihrig for reporting the bug and make a PR to fix it.

MaximoLiberata:
> arggg: https://github.com/mqttjs/MQTT.js/actions/runs/10200588659/job/28220438706?pr=1911#step:6:337

I think some opened connection never close (waiting for the callback) because the test run very fast and are we call `client.end` or `server.close` as soon as we can do it.

robertsLando:
> I think some opened connection never close (waiting for the callback) because the test run very fast and we are call client.end or server.close as soon as we can do it.

Yeah I know, seems now it get stucks on keeepalive tests. Anyway for sure now it's better then before, would you like me to merge this or do you want to try something else?

MaximoLiberata:
> Yeah I know, seems now it get stucks on keeepalive tests. Anyway for sure now it's better then before, would you like me to merge this or do you want to try something else?

I can try cover the functions `connect` and `serverBuilder` with `teardownHelper`. In this way we ensure that any client's instance can close. Example:

```
function connect(opts?: IClientOptions | string) {
	if (typeof opts === 'string') {
		opts = { host: opts }
	}
	opts = { ...config, ...opts } as IClientOptions
	const instance = mqtt.connect(opts)
	teardownHelper.addClient(instance)
	return instance
}
```

Also, I was trying to execute `beforeEach(beforeEachExec)` in the top of each `describe` hook and I found an issue with the execution of some tests according to run them very fast and a possible hang.

```
// This test is executed very fast before close the client connection
it('should send a subscribe message (offline)', function _test(t, done) {
	const client = connect()

	client.subscribe('test')

	server.once('client', (serverClient) => {
		serverClient.once('subscribe', () => {
			// Change to `client.end(done)` to fix it
			done()
		})
	})
})
```

MaximoLiberata:
This will modify the logic of the initial PR a little bit, so I don't know if adding it in this PR or another.

test/helpers/clean_method.ts
MaximoLiberata: LGTM

robertsLando:
@MaximoLiberata Let's merge this IMO and do the changes on another 👍🏼

Issue #1512, opened 2022-07-19
Unsuback user callback is invoked with no context


bretambrose:
The unsuback handler calls the user callback with null: https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/client.js#L1657-L1662

There doesn't seem to be a way to get any of the reason codes from the unsuback (let alone reason string and user properties).

Adding the packet to the callback invocation might not be a breaking change, but it might also be a surprise if the recipient was used to a null value for the packet.

robertsLando:
@bretambrose would you like to submit a PR?

PR #1922, opened 2024-08-02
bug: Invoke the unsubscribe ack callback with the Unsuback packet, not null

bretambrose:
https://github.com/mqttjs/MQTT.js/issues/1512 (I cannot re-open the issue)


robertsLando (approved):
LGTM

Issue #1510, opened 2022-07-18
Consider exposing Suback packet in Subscribe's completion callback


bretambrose:
Publish and Unsubscribe both use completion callbacks that take the full packet, but Subscribe's callback is the array of granted subscriptions, which keeps the Suback User Properties or Reason String from being exposed to the user.  

I'm not particularly confident that I know what is and isn't a breaking change in Javascript, so I thought I'd wait for some feedback before making a PR along the lines of:

```
export declare type ClientSubscribeCallback = (err: Error, granted: ISubscriptionGrant[]) => void
```
to
```
export declare type ClientSubscribeCallback = (err: Error, granted: ISubscriptionGrant[], packet?: Packet) => void
```

with associated logic update where the Suback is handled:

`callback(err, subs)` to `callback(err, subs, packet)` in subscribe()

robertsLando:
@bretambrose Consider opening a PR to fix this

PR #1923, opened 2024-08-02
feat: add `suback` packet to subscribe callback

bretambrose:
https://github.com/mqttjs/MQTT.js/issues/1510


robertsLando (commented):
Please add a test that checks the packet is received in the callback (you can add a check in an existing test)

bretambrose:
> Please add a test that checks the packet is received in the callback (you can add a check in an existing test)

Done

robertsLando (approved):

PR #1926, opened 2024-08-09
feat(electron-test): move third hosted broker to self hosted broker

MaximoLiberata:
electron tests may present connection issues when try to connect to third hosted broker. The tests will fail for external issues not related to MQTT.JS. A self hosted broker will mitigate these issues.


MaximoLiberata:
@axi92 Could you review this PR, too?

robertsLando (changes requested):
I higly suggest to use aedes-cli instead:

https://github.com/mqttjs/MQTT.js/blob/2f8e73975de1c043b1691a564e120625cf84eb3d/web-test-runner.config.mjs#L8-L16

MaximoLiberata:
@robertsLando I have a question, the `ServerLauncher` class/service is launched before each test starts. So far, the electron tests didn't fail, but I'm not sure if they will fail because duplicate server ports like node test issues.

Is there a way to close `aedes-cli` broker? I saw its the documentation, but I didn't find a method to close it

robertsLando:
@MaximoLiberata `start` returns servers, broker instances so yeah you can close them:

https://github.com/moscajs/aedes-cli/blob/master/lib/cli.js#L420

MaximoLiberata:
Thanks @robertsLando

electron-test/wdio.conf.ts
```diff
@@ -116,7 +119,12 @@ export const config: Options.Testrunner = {
     // Services take over a specific job you don't want to take care of. They enhance
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
-    services: ['electron'],
+    services: [
+        'electron',
+        [ServerLauncher, {
+            somsomeOption: true
```
robertsLando, line 125: I think this is a typo? 
```suggestion
            someOption: true
```

electron-test/tsconfig.json
```diff
@@ -19,7 +19,6 @@
         "allowImportingTsExtensions": true,
         "resolveJsonModule": true,
         "isolatedModules": true,
-        "strict": true,
```
robertsLando, line 22: @axi92 This affects only the electron test project not the MQTTjs project so not a big deal IMO

robertsLando (changes requested):

robertsLando:
@MaximoLiberata Sorry for the late review but I'm on holiday, I will be back on the end of August :)

electron-test/tsconfig.json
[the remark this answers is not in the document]
MaximoLiberata: Done. Please check it.

MaximoLiberata: Done. Please check it.

electron-test/wdio.conf.ts
```diff
@@ -116,7 +119,12 @@ export const config: Options.Testrunner = {
     // Services take over a specific job you don't want to take care of. They enhance
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
-    services: ['electron'],
+    services: [
+        'electron',
+        [ServerLauncher, {
+            someOption: true
```
axi92, line 125: I don't think we need the options, do we?
I guess you got it from that [example](https://webdriver.io/docs/customservices/#import-service-from-module)?

MaximoLiberata: Yes, I was following the instructions of the example. We can remove the option to reduce code or let it.

@axi92 Which option do you think is best? I'm according with both.

robertsLando: I agree this can be removed as it seems to be just a placeholder option put there as an example

robertsLando (changes requested):
Just came back from holidays, sorry for the late review! Everything looks good to me, just remove the useless option and I will merge this, thanks 🙏🏼

electron-test/wdio.conf.ts
MaximoLiberata: Done.

MaximoLiberata:
thanks @robertsLando and @axi92 for review 🙏🏼

robertsLando (approved):

PR #1927, opened 2024-08-15
fix: replace deprecated `url.parse` with `new URL`

### LINKED ISSUE #1569: When connecting `ws://[::1]:8883`, an error is thrown - Uncaught TypeError TypeError [ERR_INVALID_URL]: Invalid URL
Hi, I am getting an exception when connecting to IPv6 localhost address `::1`

I have found this was reported and fixed in https://github.com/mqttjs/MQTT.js/issues/1130 / #1147 (commit https://github.com/mqttjs/MQTT.js/commit/70a247c29e0b05ddd8755e7b9c8c41a4c25b431b)  however it seems to have been reverted and the current V4.3.7 is once more using the depreciated `url.parse` as you can see here: https://github.com/mqttjs/MQTT.js/blob/main/lib/connect/index.js#L64

Here is a minimal repro...
```js
        const userEnteredURL = 'ws://[::1]:8883'
        const client = mqtt.connect(userEnteredURL, this.brokerConfig)
```

Here is why I think the underlying issue exists: `mqtt.js` manually rebuilds the broker URL using hostname + port in
https://github.com/mqttjs/MQTT.js/blob/main/lib/connect/ws.js#L20 
```
let url = opts.protocol + '://' + opts.hostname + ':' + opts.port + opts.path
```
... but, `url.parse().hostname` drops the squarebrackets off `[::1]` which causes the above code to generate the url `http://::1:8883`.  Further downstream, the `ws` package uses `new URL(...)` with this re-assembled URL string & the `ERR_INVALID_URL` error occurs.

For anyone else reading, here is a minimal (hacky) workaround...
```js
        const userEnteredURL = 'ws://[::1]:8883'
        const tempNewURL= new URL(userEnteredURL)
        const brokerURL = url.parse(userEnteredURL)
        brokerURL.hostname = tempNewURL.hostname // force the hostname back to [::1]
        const client = mqtt.connect(brokerURL, this.brokerConfig)
```

---

### Additional info...

* Calling `mqtt.connect(new URL('http://[::1]:8883'))` fails because `connect` only accepts a `string` or `url.parse` object
* Calling `mqtt.connect(url.parse('http://[::1]:8883')` fails because unlike `new URL`, `url.parse` drops the square brackets off `hostname` and `mqtt.js uses `hostname` to reassemble the `url` 
* websocket code uses `new URL` so when `mqtt.js` passes the reassembled IP `http://::1:8883` it failes with error ERR_INVALID_URL

##### `url.parse('http://[::1]:8883')`
```
Url {
  protocol: 'ws:',
  slashes: true,
  auth: null,
  host: '[::1]:8883',
  port: '8883',
  hostname: '::1', //<< square brackets are dropped
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'ws://[::1]:8883/'
}
```

##### `new URL('http://[::1]:8883')`
```
URL {
  href: 'ws://[::1]:8883/',
  origin: 'ws://[::1]:8883',
  protocol: 'ws:',
  username: '',
  password: '',
  host: '[::1]:8883',
  hostname: '[::1]', //<< square brackets are kept
  port: '8883',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

Steve-Mcl:
closes #1569

Alters internal processing of connection URL. Removes `url.parse` and uses `new URL`  

This is done in a backwards compatible manor however I would recommend updating all types and function signatures to use the the `URL` object instead of a `string` (or as well as) but that may be more suited to a major bump.

NOTE: there are some TODOs in the code that need to be understood (and removed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ coverage
 test/typescript/.idea/*
 test/typescript/*.js
 test/typescript/*.map
+test-store*/*
 # VS Code stuff
 **/typings/**
 .vscode/
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -19,24 +19,6 @@ const debug = _debug('mqttjs')
 let protocols: Record<string, StreamBuilder> = null
 
 /**
- * Parse the auth attribute and merge username and password in the options object.
- *
- * @param {Object} [opts] option object
- */
-function parseAuthOptions(opts: IClientOptions) {
-	let matches: RegExpMatchArray | null
-	if (opts.auth) {
-		matches = opts.auth.match(/^(.+):(.+)$/)
-		if (matches) {
-			opts.username = matches[1]
-			opts.password = matches[2]
-		} else {
-			opts.username = opts.auth
-		}
-	}
-}
-
-/**
  * connect - connect to an MQTT broker.
  */
 function connect(brokerUrl: string): MqttClient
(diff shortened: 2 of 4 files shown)


Steve-Mcl:
@robertsLando is there something I need to do to permit tests to run?

robertsLando:
@Steve-Mcl sorry I was on vacation till yesterday. I have to approve your PR in order to allow test to run, done now. I will also do a quick review

robertsLando:
I know there have been another try to do this in the past: https://github.com/mqttjs/MQTT.js/pull/1147

Problem is that it has been reverted as it caused many compatibility issues. We need to track all possible issues related to this change.

jouniso:
Are there any plans to move this forward in the near future?

robertsLando:
@jouniso Would you take care of: https://github.com/mqttjs/MQTT.js/pull/1927#issuecomment-2314760795 ?

jouniso:
@robertsLando Unfortunately at the moment I have no time to contribute to this :(

PR #1965, opened 2025-01-31
fix: move protocol "clean" after merge of parsedOptions

mmoriani:
With this [commit](https://github.com/mqttjs/MQTT.js/commit/1004c78db7d6763f21c98fa3db2f12e688ca33ff) the removal of the “:”, for the protocol passed in connect options, was moved before the merge of parsedOptions.

```
parsedOptions.protocol = parsedOptions.protocol?.replace(
	/:$/,
	'',
) as MqttProtocol

opts = { ...parsedOptions, ...opts }

```

Previously, if someone passed as a protocol “mqtts:” (Protocol returned by the [URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol)) this was “cleaned” by removing the “:”.

By moving the replace after the options merge, a backward compatibility is maintained for those, like me, who find themselves having to update deps in codes where the protocol was passed in by taking the value from the URL interface.

I know that the documentation as protocol values does not accept “:”, but so far it has accepted them anyway and fixed if url is passed.

I also think that if the URL interface returns the protocol in this way, it may be convenient to pass it directly without having to modify it.

### DIFF
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -79,10 +79,7 @@ function connect(
 			throw new Error('Missing protocol')
 		}
 
-    opts.protocol = opts.protocol.replace(
-      /:$/,
-      '',
-    ) as MqttProtocol
+		opts.protocol = opts.protocol.replace(/:$/, '') as MqttProtocol
 	}
 
 	opts.unixSocket = opts.unixSocket || opts.protocol?.includes('+unix')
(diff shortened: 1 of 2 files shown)


robertsLando (changes requested):
Could you submit a quick test to cover this please?

robertsLando (approved):

PR #1966, opened 2025-02-05
feat: add intrinsic support for SOCKS proxies

DirtyHairy:
This PR adds support for establishing TCP and TLS connections through a SOCKS proxy. Proxy support is enabled with a new `socks` configuration parameter that takes a curl-style SOCKS URL, or by setting an environment variable (`MQTTJS_SOCKS_PROXY`).

Supported schemes (and protocols) are `socks5://` , `socks5h://` , `socks4://` and  `socks4a://` .

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -140,6 +140,10 @@ export interface IClientOptions extends ISecureClientOptions {
 	query?: Record<string, string>
 	/** Auth string in the format <username>:<password> */
 	auth?: string
+	/** Optional SOCKS proxy to use for TCP / TLS connections , i.e. socks5://localhost:1333, socks4://localhost:1333, socks5h://localhost:1333 . Default is socks5h. */
+	socks?: string
+	/** Timeout for establishing a socks connection */
+	socksTimout?: number
 	/** Custom ack handler */
 	customHandleAcks?: AckHandler
 	/** Broker port */
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -141,9 +141,9 @@ export interface IClientOptions extends ISecureClientOptions {
 	/** Auth string in the format <username>:<password> */
 	auth?: string
 	/** Optional SOCKS proxy to use for TCP / TLS connections , i.e. socks5://localhost:1333, socks4://localhost:1333, socks5h://localhost:1333 . Default is socks5h. */
-	socks?: string
+	socksProxy?: string
 	/** Timeout for establishing a socks connection */
-	socksTimout?: number
+	socksTimeout?: number
 	/** Custom ack handler */
 	customHandleAcks?: AckHandler
 	/** Broker port */
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -104,6 +104,18 @@ function connect(
 		opts.clientId = opts.query.clientId
 	}
 
+	if (isBrowser) {
+		opts.socks = undefined
+	}
+
+	if (
+		!isBrowser &&
+		opts.socks === undefined &&
+		typeof process !== 'undefined'
+	) {
+		opts.socks = process.env['MQTTJS_SOCKS_PROXY']
+	}
+
 	if (opts.cert && opts.key) {
 		if (opts.protocol) {
 			if (['mqtts', 'wss', 'wxs', 'alis'].indexOf(opts.protocol) === -1) {
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -106,7 +106,8 @@ function connect(
 
 	if (isBrowser || opts.unixSocket) {
 		opts.socksProxy = undefined
-	} else if(opts.socksProxy === undefined &&
+	} else if (
+		opts.socksProxy === undefined &&
 		typeof process !== 'undefined'
 	) {
 		opts.socksProxy = process.env['MQTTJS_SOCKS_PROXY']
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -106,12 +106,7 @@ function connect(
 
 	if (isBrowser || opts.unixSocket) {
 		opts.socksProxy = undefined
-	}
-
-	if (
-		!isBrowser &&
-		!opts.unixSocket &&
-		opts.socksProxy === undefined &&
+	} else if(opts.socksProxy === undefined &&
 		typeof process !== 'undefined'
 	) {
 		opts.socksProxy = process.env['MQTTJS_SOCKS_PROXY']
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -105,16 +105,16 @@ function connect(
 	}
 
 	if (isBrowser || opts.unixSocket) {
-		opts.socks = undefined
+		opts.socksProxy = undefined
 	}
 
 	if (
 		!isBrowser &&
 		!opts.unixSocket &&
-		opts.socks === undefined &&
+		opts.socksProxy === undefined &&
 		typeof process !== 'undefined'
 	) {
-		opts.socks = process.env['MQTTJS_SOCKS_PROXY']
+		opts.socksProxy = process.env['MQTTJS_SOCKS_PROXY']
 	}
 
 	if (opts.cert && opts.key) {
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -104,12 +104,13 @@ function connect(
 		opts.clientId = opts.query.clientId
 	}
 
-	if (isBrowser) {
+	if (isBrowser || opts.unixSocket) {
 		opts.socks = undefined
 	}
 
 	if (
 		!isBrowser &&
+		!opts.unixSocket &&
 		opts.socks === undefined &&
 		typeof process !== 'undefined'
 	) {
--- a/src/lib/connect/socks.ts
+++ b/src/lib/connect/socks.ts
@@ -65,27 +65,26 @@ class ProxyStream extends Duplex {
 		callback(error)
 	}
 
-	_onData = (chunk: any): void => {
+	private _onData = (chunk: any): void => {
 		assert(this._socket)
 
 		this._flowing = this.push(chunk)
-
 		if (!this._flowing) this._socket.pause()
 	}
 
-	_onEnd = (): void => {
+	private _onEnd = (): void => {
 		debug('proxy stream received EOF')
 
 		this.push(null)
 	}
 
-	_onClose = (): void => {
+	private _onClose = (): void => {
 		debug('proxy stream closed')
 
 		this.destroy()
 	}
 
-	_onError = (err: any): void => {
+	private _onError = (err: any): void => {
 		debug('proxy stream died with error %s', err)
 
 		this.destroy(err)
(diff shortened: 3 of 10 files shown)


src/lib/connect/index.ts
```diff
@@ -104,6 +104,19 @@ function connect(
...
+	}
+
+	if (
+		!isBrowser &&
+		!opts.unixSocket &&
+		opts.socks === undefined &&
+		typeof process !== 'undefined'
+	) {
+		opts.socks = process.env['MQTTJS_SOCKS_PROXY']
+	}
```
robertsLando, line 118: why? Could you remove the env var here?

src/lib/client.ts
```diff
@@ -140,6 +140,10 @@ export interface IClientOptions extends ISecureClientOptions {
 	query?: Record<string, string>
 	/** Auth string in the format <username>:<password> */
 	auth?: string
+	/** Optional SOCKS proxy to use for TCP / TLS connections , i.e. socks5://localhost:1333, socks4://localhost:1333, socks5h://localhost:1333 . Default is socks5h. */
+	socks?: string
```
robertsLando, line 144: name it socksProxy instead

src/lib/client.ts
```diff
@@ -140,6 +140,10 @@ export interface IClientOptions extends ISecureClientOptions {
...
+	/** Timeout for establishing a socks connection */
+	socksTimout?: number
```
robertsLando, line 146: Typo

```suggestion
	socksTimeout?: number
```

src/lib/connect/socks.ts
```diff
@@ -0,0 +1,155 @@
+import _debug from 'debug'
+import duplexify, { Duplexify } from 'duplexify'
+import { SocksClient, SocksProxy } from 'socks'
+import { lookup } from 'dns'
+import { SocksProxyType } from 'socks/typings/common/constants'
+import { IStream } from '../shared'
+import { promisify } from 'util'
```
robertsLando, line 7: I refactored the code a lot in order to drop duplexify, I suggest you to use BufferedStream util class instead If the goal is to buffer requests until connection

robertsLando (changes requested):

src/lib/connect/index.ts
DirtyHairy: I added the environment variable in order to provide an easy way to route unmodified applications that use mqtt.js through a proxy. In or case the culprit is Node-RED.

It would be definitely preferable if these had support for using a proxy themselves, but configuring a proxy is one of those networking options where it is not uncommon to offer configuration through the environment.

src/lib/connect/socks.ts
DirtyHairy: Fair enough. The reason why I used it is to be able to immediately return the stream even though the actual socket only becomes available when the proxy has connected (so, indeed, buffering 😏 ).

This merge request is based on an older version that I did against an older version of MQTT.js which had duplexify, so I went with what was there back then, and when I rewrote it I didn't think much about it and just pulled it back in.

That said, I can definitely drops duplexify for any other way to achieve the same means. I am not sure what you refer to by `BufferedStream` though --- there's `BufferedDuplex`, but that's very specific to web socket connections. I can roll a similar helper based on nodejs `stream.Duplex` though if that's what you prefer.

src/lib/connect/index.ts
```diff
@@ -104,6 +104,19 @@ function connect(
...
+
+	if (
+		!isBrowser &&
+		!opts.unixSocket &&
+		opts.socksProxy === undefined &&
+		typeof process !== 'undefined'
+	) {
+		opts.socksProxy = process.env['MQTTJS_SOCKS_PROXY']
+	}
```
robertsLando, line 118:
```suggestion
	if (isBrowser || opts.unixSocket) {
		opts.socksProxy = undefined
	} else if(opts.socksProxy === undefined &&
		typeof process !== 'undefined'
	) {
		opts.socksProxy = process.env['MQTTJS_SOCKS_PROXY']
	}
```

robertsLando (changes requested):

src/lib/connect/socks.ts
DirtyHairy: Done.

robertsLando:
@DirtyHairy Would it be possible to test part of this?

DirtyHairy:
Sure, done 😏

DirtyHairy:
Just checking in to see whether there is something still missing.

src/lib/connect/socks.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+function fatal<T>(e: T): T {
+	try {
+		if ((e as any).code === undefined) (e as any).code = 'SOCKS'
+		return e
+	} catch {
+		return e
+	}
+}
```
robertsLando, line 111: check `ErrorWithReasonCode` and do something like that instead:

```ts
export class SocksError extends Error {
	public code: string

	public constructor(message: string) {
		super(message)
		this.code = 'SOCKS'

		// We need to set the prototype explicitly
		Object.setPrototypeOf(this, SocksError.prototype)
		Object.getPrototypeOf(this).name = SocksError.name
	}
}
```

robertsLando (changes requested):
Tests are failing

robertsLando:
@DirtyHairy That's not the right approach to fix tests, I suggest you to ensure there are no leaked handlers that prevent test to exit

DirtyHairy:
@robertsLando I don't think the reason are leaking handlers, I think this is a node.js issue similar to https://github.com/nodejs/node/issues/54163 . I got errors because of  remaining work on the loop before, and these look differently.

Furthermore I can't reproduce this issue locally with either `v20.18.2` (which is what actions is running for 20) or `v23.7.0`, and for `v18.20.7` (again what the CI claims to be running for 18) I do get this exact error for other, existing suites, even on main (without any changes from me, just doing `npm install && npm run test:node`:

```
 ▶ keepalive
    ✔ should send ping at keepalive interval (2.114583ms)
    ✔ should not shift ping on publish (1.18025ms)
    ✖ should reschedule pings if publishing at a higher rate than keepalive and reschedulePings===true (1.017125ms)
      'test did not finish before its parent and was cancelled'

    ✖ should not reschedule pings if publishing at a higher rate than keepalive and reschedulePings===false
      'test did not finish before its parent and was cancelled'

    ✖ should shift ping on pingresp when reschedulePings===true
      'test did not finish before its parent and was cancelled'

    ✖ should shift ping on pingresp when reschedulePings===false
      'test did not finish before its parent and was cancelled'
```

DirtyHairy:
But, anyway, my commit does not fix the tests either 🤷‍♂️ I must admit I am slightly clueless here as I can't reproduce the issue.

DirtyHairy:
Also, not getting this error for any other run on the CI is a pretty strong argument for something being wrong with my tests 😏 I just have not clue what is going on.

robertsLando:
@DirtyHairy Seems you fixed it?

DirtyHairy:
Na, sorry, not yet. But I've got progress: I can reproduce the fail in a Linux VM (I am on a Mac) 😏 With a reproducer I'll get that fixed.

src/lib/connect/socks.ts
DirtyHairy: I am fine with making that change. However, I still need to keep the function to "decorate" errors from the SOCKS client with `.code` so that MQTT.js will treat them as fatal (I'd rather not wrap them in another `Error` as that obscures the original reason).

Should I make the change anyway?

robertsLando: > I'd rather not wrap them in another Error as that obscures the original reason

You are not wrapping them twice you just create a specific error instance for sock related errors and that is exactly what you are doing now but in a non-typed way

robertsLando:
@mcollina could you kindly give a check to this please?

src/lib/connect/socks.ts
robertsLando: About the first part cannot you do something like:

```ts
 constructor(messageOrError: string | Error) {
    if (messageOrError instanceof Error) {
      super(messageOrError.message);

      if (messageOrError.stack) {
        this.stack = messageOrError.stack;
      }
    } else {
      super(messageOrError);
    }
    
    Object.setPrototypeOf(this, SocksError.prototype);
    this.name = "SocksError";
  }
  ```
  
  > BTW: why are you manipulating the prototype directly? My understanding is that this used to be necessary for ES5 (as Error always returns a fresh object if called directly, independent of the scope), but for ES2015 the spec explicitly requires Error to be subclassable.
  
From what I know:
  
-  `Object.setPrototypeOf(this, SocksError.prototype)`: This used to be necessary in order to have `e instanceof Error` returning true, without that it used to return false
- `Object.getPrototypeOf(this).name = SocksError.name` without this the `name` property on the error thrown it's `Error` and not `SocksError`

Comment (DirtyHairy) [reply src/lib/connect/socks.ts:111]: > About the first part cannot you do something like:

Compared to that I personally prefer just setting `code` on the error. But, then again, this is not my project, so my personal preferences are not what counts here 😏 So, if you prefer that, I'll change the code.

> From what I know:

I think you can do without that by now, unless you want to support compiling down to ES5:

```
Welcome to Node.js v23.7.0.
Type ".help" for more information.
> class FooError extends Error {name = 'FooError'}
undefined
> foo = new FooError()
FooError
    at REPL2:1:7
    at ContextifyScript.runInThisContext (node:vm:137:12)
    at REPLServer.defaultEval (node:repl:597:22)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:926:10)
    at REPLServer.emit (node:events:519:35)
    at REPLServer.emit (node:domain:489:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
    at [_line] [as _line] (node:internal/readline/interface:887:18)
> foo.name
'FooError'
> foo instanceof FooError
true
> foo instanceof Error
true
```

But, same here, I can just go ahead and add the prototype mangling.

Comment (robertsLando) [reply src/lib/connect/socks.ts:111]: problem is that `foo instanceof Error` should be false?

Comment (DirtyHairy) [reply src/lib/connect/socks.ts:111]: > Anyway keep it like it is now no problem

Alright, I won't complain about that 😏

> problem is that foo instanceof Error is false without the second line?

The issue is `foo instanceof FooError` which is false for ES5 / "old-school" class definitions (unless you set up the prototype correctly after calling `Error`). If you extend `Error` ES2015 style with `class` / `extends` the prototype chain is set up correctly. My understanding is that the spec requires `Error` to check [`new.target`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target) to see whether it is called in a ES2015 constructing context and to behave accordingly (modify the target instead of returning a new object) in this case.

If you run the code in my NodeJS session through TS in ES5 mode `foo instanceof FooError` is false, and `foo instanceof Error` is true.

**Comment by web3botshop** [comment]
I can't use this
I tried:
`$env:DEBUG="mqttjs:socks"; node .\index.js `

but there is no debug output, and I can't find "mqttjs:socks" in dist

**Comment by DirtyHairy** [comment]
@wshshra What are you trying to do, and what happens, and what do you expect to happen? Can you show some code?

**Comment by wshshra** [comment]
> @wshshra您想做什么,会发生什么,您期望发生什么？您可以显示一些代码吗？

my code:
```
const options = {
      socksProxy: "socks5://example.com:port",
      ...otherOptions
    };
mqtt.connect(url, options)
```
the socks not work，then i want debug it, but there is not debug output

Finally I solved it like this
```
wsOptions: {
      agent,
 }
```

PR #1968, opened 2025-02-14
docs: npm download count link

cwtuan:
I’d like to suggest updating the NPM Downloads badge in the README to link to the download trends page on npm-compare.com, which provides a dynamic, three-year download trend chart for [mqtt](https://npm-compare.com/mqtt/#timeRange=THREE_YEARS).

While the other badges already link to the [NPM package page](https://www.npmjs.com/package/mqtt), this particular NPM Downloads badge will benefit from linking to npm-compare.com, where users can see a more informative, visual representation of the download trends over time.

By the way, i'm a maintainer of npm-compare.com, If you need any additional features or insights from npm-compare.com, I’d be more than happy to assist.

Thanks for considering my suggestion. I look forward to your feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
 [![Maintenance](https://img.shields.io/badge/Maintained%3F-yes-green.svg)](https://github.com/mqttjs/MQTT.js/graphs/commit-activity)
 [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://github.com/mqttjs/MQTT.js/pulls)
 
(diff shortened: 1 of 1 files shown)


robertsLando (approved):
LGTM

robertsLando:
@cwtuan I don't understand why here: https://npm-compare.com/aedes,mqtt,paho-mqtt

It mentions that mqttjs is more limited compared to paho and also that it doens't support MQTT v5 protocol. Do you know how can that be updated?

robertsLando:
> The mqtt package offers both client and broker functionality, making it a versatile choice for applications that need to implement both sides of the MQTT protocol. This dual capability allows for greater flexibility in how the library is used.

@cwtuan this is wrong as well, mqttjs cannot be used as mqtt broker

robertsLando:
I would simply use the description in official readme:

> MQTT.js is a client library for the [MQTT](http://mqtt.org/) protocol, written in JavaScript for node.js and the browser.

robertsLando:
@cwtuan Ok thanks

PR #1970, opened 2025-02-14
fix: remove undefined user-properties before sending Packet

SoulKa:
# Changes

- Remove `undefined` properties before sending packet

This fixes an error where the user sets a property to `undefined` which results in a `TypeError` (closes mqttjs/mqtt-packet#152).


robertsLando:
@SoulKa I have moved the issue to mqtt-packet repo. Please open a PR there

PR #1971, opened 2025-02-16
fix(electron-test): webdriver session not created

MaximoLiberata:
This PR fix the error of webdriver when attempt to start a new user session in browser.


MaximoLiberata:
Go to PR [1972](https://github.com/mqttjs/MQTT.js/pull/1972)

PR #1972, opened 2025-02-16
fix(electron-test): webdriver session not created

MaximoLiberata:
This PR fix the error of chromdriver when it attempts to start a new session in browser.


robertsLando (approved):
Thanks! LGTM 🙏🏼

PR #1986, opened 2025-04-17
fix(deps): removed unused 'reinterval' dependency and move two @type/ dependencies to devDependencies.

### LINKED ISSUE #1985: Remove @type/ dependencies and unused dependencies from the dependencies
Hey,

I'm opening this issue and will immediately open a PR resolving it.
I would just like to remove the `@type/` dependencies from `dependencies` as they needlessly increase the size of the build. There is also another unused dependency (`reinterval`) that can be removed.

SmashingQuasar:
# Description

This PR resolves #1985:
- Removes the unused `reinterval` dependency.
- Moves `@types/ws` to `devDependencies` from `dependencies`.
- Moves `@types/readable-stream` to `devDependencies` from `dependencies`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,6 @@
       "version": "5.11.0",
       "license": "MIT",
       "dependencies": {
-        "@types/readable-stream": "^4.0.18",
-        "@types/ws": "^8.5.14",
         "commist": "^3.2.0",
         "concat-stream": "^2.0.0",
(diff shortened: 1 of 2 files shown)


robertsLando (approved):
Thanks for this @SmashingQuasar! Too focus on the code that I never spotted this :)

SmashingQuasar:
Anytime. 👍 :)

Issue #1685, opened 2023-09-05
Connect fails silently if no username is provided but only a password


AuspeXeu:
Consider the following piece of code.
```
import Server, { AuthenticateError } from 'aedes';
import * as net from 'net';
import { connect } from 'mqtt';

const mqttServer = new Server({
  authenticate(_, __, password, callback) {
    if (password?.toString() !== 'MY GOOD SECURE PASSWORD') {
      const error: AuthenticateError = { ...new Error('Auth error'), returnCode: 4 };
      callback(error, null);
    } else {
      callback(null, true);
    }
  }
});

net
  .createServer(mqttServer.handle)
  .on('error', (err) => console.error(err))
  .listen(1883, '192.168.0.21', () => {
    console.log(`Aedes server listening on 192.168.0.21:1883`);
    const mqttClient = connect('mqtt://192.168.0.21', {
      // username: 'does-not-matter',
      password: 'MY GOOD SECURE PASSWORD',
      reconnectPeriod: 2 * 1000
    })
  })
```

In this example, the client will never connect and you will never find out why until you comment the username back in.

robertsLando:
Would you like to submit a PR to fix this?

AuspeXeu:
Bump

PR #1989, opened 2025-04-30
fix: Add warning when password is set without username

adam-abdelrahman0:
Ref #1685 

This change implements a warning if a password is provided without a username in an ```IClientOptions``` object in the ```connect()``` function.

- [index.ts](https://github.com/mqttjs/MQTT.js/blob/main/src/lib/connect/index.ts): add condition after ```parseAuthOptions()``` to warn if no username has been set.

### DIFF
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -99,6 +99,12 @@ function connect(
 	// merge in the auth options if supplied
 	parseAuthOptions(opts)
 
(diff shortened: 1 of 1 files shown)


robertsLando:
@adam-abdelrahman0 there is no specification that explains this, I mean someone could have users with empty username for whatever reason, cannot accept this PR

PR #1993, opened 2025-05-06
chore(1991): Added @typescript-eslint/consistent-type-imports rule and normalised all imports.

### LINKED ISSUE #1991: [Bug]: Build error: Could not find a declaration file for module 'ws' in 5.11.1 and beyond - works in 5.11.0
### MQTTjs Version

5.12.0

### Broker

N/A

### Environment

NodeJS

### Description

Working in a TypeScript project that's always been fine. Updated the package release from 5.11.0 to 5.12.0. Transpiling yields:

```
node_modules/mqtt/build/lib/client.d.ts:5:31 - error TS7016: Could not find a declaration file for module 'ws'. '/myprojectdir/node_modules/ws/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/ws` if it exists or add a new declaration (.d.ts) file containing `declare module 'ws';`

5 import { ClientOptions } from 'ws';
```

### Minimal Reproduction

Admittedly not minimal...but it makes the point. 😄 
 
1.. Download the source: https://github.com/hjdhjd/homebridge-plugin-utils/archive/refs/tags/v1.15.3.tar.gz
2. Update the MQTT package version to 5.11.0.
3. npm run build to verify it works.
4. Update the MQTT package version to 5.11.1 or 5.12.0
5. npm run clean ; npm run build
6. Error above occurs.

### Debug logs

N/A. See above.

SmashingQuasar:
# Description

This PR fixes #1991.

It does two separate things:
- Adds the `@typescript-eslint/consistent-type-imports` rule to force all imports used exclusively as types to be declared with the `import type` syntax. This will eliminate them at run time.
- Automatically converts all type imports as such by the `eslint --fix` feature.

This will solve the underlying problem as it will make `@types/readable-stream` and `@types/ws` true development dependencies instead of extraneous dependencies.


robertsLando (approved):
LGTM

PR #1995, opened 2025-05-09
feat: add `subscribeBatchSize` option to batch SUBSCRIBE packets

Sheile:
# Overview
AWS IoT Core limits each SUBSCRIBE packet to a maximum of 8 topics.

> A single SUBSCRIBE request has a quota of 8 subscriptions.
https://docs.aws.amazon.com/general/latest/gr/iot-core.html#message-broker-limits

This PR adds a new `subscribeBatchSize?: number` option that lets users control how many topics are sent in each packet. When the total number of topics to subscribe exceeds this value, the client will automatically split them into multiple SUBSCRIBE packets of the specified size.

# Usage
```typescript
const options = {
  ...
  subscribeBatchSize: 8,  // If you attempt to subscribe to more than 8 topics without this option, AWS IoT Core will reject the SUBSCRIBE packet.
}
const mqttClient = mqtt.connect('mqtts://xxxxxx-ats.iot.ap-northeast-1.amazonaws.com:8883', options)
const topics = ['/dummy1', '/dummy2', '/dummy3', '/dummy4', '/dummy5', '/dummy6', '/dummy7', '/dummy8', '/dummy9']
mqttClient.subscribe(topics, {}, (err) => { console.log(`subscribe is finished with ${err}`) })
```

# Discussion
The third argument of [`Client#subscribe`](https://github.com/mqttjs/MQTT.js/tree/main?tab=readme-ov-file#mqttclientsubscribetopictopic-arraytopic-object-options-callback) is a callback that receives three parameters:
1. `err`
2. `granted`
3. `packet` (undocumented)

Because this option may split your subscriptions into multiple SUBSCRIBE packets, the third parameter is handled as follows:
- On success: the callback is invoked with the last packet.
- On failure: the callback is invoked with the failed packet.

Any thoughts on whether this callback behavior is OK, or suggestions for improving it?


robertsLando:
@Sheile thanks for your PR! Giving that this is something specific to AWS iOT I'm wondering if it could make more sense to let users handle this theirself, I usually create a wrapper around mqtt client to handle every call to it, it's a design choice I use everywhere.

@mcollina what do you think?

Sheile:
@robertsLando 
Thanks for quick reply!

When calling `Client#subscribe`, User can split topics manually.
But, resubscribe in this client will send SUBSCRIBE packet with all topics automatically.

Therefore, User can't use this client with 9 or more topics in AWS IoT Core.

mcollina (approved):
Lgtm

I’m ok with it.

robertsLando (approved):

Issue #2006, opened 2025-07-03
[Bug]: typescript cannot compile since v5.11.1


DavideViolante:
### MQTTjs Version

5.11.1 and above (also 5.13.1)

### Broker

Aedes

### Environment

NodeJS

### Description

Simply upgrading from https://github.com/mqttjs/MQTT.js/compare/v5.11.0...v5.11.1 5.11.0 to 5.11.1 the build of my typescript project returns these errors:

```
> tsc

node_modules/mqtt/build/lib/client.d.ts:3:31 - error TS7016: Could not find a declaration file for module 'readable-stream'. '/home/dave/job/myproject/mqtt-broker/node_modules/mqtt/node_modules/readable-stream/lib/ours/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/readable-stream` if it exists or add a new declaration (.d.ts) file containing `declare module 'readable-stream';`

3 import { DuplexOptions } from 'readable-stream';
                                ~~~~~~~~~~~~~~~~~

node_modules/mqtt/build/lib/client.d.ts:5:31 - error TS7016: Could not find a declaration file for module 'ws'. '/home/dave/job/myproject/mqtt-broker/node_modules/ws/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/ws` if it exists or add a new declaration (.d.ts) file containing `declare module 'ws';`

5 import { ClientOptions } from 'ws';
                                ~~~~

node_modules/mqtt/build/lib/store.d.ts:1:26 - error TS7016: Could not find a declaration file for module 'readable-stream'. '/home/dave/job/myproject/mqtt-broker/node_modules/mqtt/node_modules/readable-stream/lib/ours/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/readable-stream` if it exists or add a new declaration (.d.ts) file containing `declare module 'readable-stream';`

1 import { Readable } from 'readable-stream';
                           ~~~~~~~~~~~~~~~~~
```

package.json
```json
...
"dependencies": {
    "@influxdata/influxdb-client": "^1.35.0",
    "aedes": "^0.51.3",
    "async-mutex": "^0.5.0",
    "axios": "^1.9.0",
    "date-fns": "~3.6.0",
    "date-fns-tz": "~3.2.0",
    "firebase": "^11.10.0",
    "firebase-admin": "^13.4.0",
    "log4js": "^6.9.1",
    "mqtt": "^5.11.1",
[...]

PR #2010, opened 2025-07-07
Strict TypeScript configuration

SmashingQuasar:
# Description

Following the discussion from #2006, this PR aims to implement a strict TypeScript configuration.


src/bin/pub.ts
```diff
@@ -4,51 +4,42 @@ import { Writable } from 'readable-stream'
...
 import minimist, { type ParsedArgs } from 'minimist'
 import split2 from 'split2'
-import { connect } from '../mqtt'
-import { type IClientOptions, type IClientPublishOptions } from 'src/lib/client'
+import { connect } from '../mqtt.js'
 import { pipeline } from 'stream'
 
 const helpMe = help({
-	dir: path.join(__dirname, '../../', 'help'),
+	dir: path.join(import.meta.dirname, '../../', 'help'),
```
SmashingQuasar, line 16: Not strictly necessary but enables migration to ESM.

src/bin/mqtt.ts
```diff
@@ -7,10 +7,12 @@
...
  */
 import path from 'path'
+// @ts-expect-error - There is no types available for this library.
 import Commist from 'commist'
+// @ts-expect-error - There is no types available for this library.
 import help from 'help-me'
-import publish from './pub'
-import subscribe from './sub'
+import publish from './pub.js'
+import subscribe from './sub.js'
```
SmashingQuasar, line 15: Best practice to avoid confusion between folders and files, there are also other more specific requirements from Node.js and browsers if using `.mjs` extension.

src/bin/pub.ts
```diff
@@ -101,65 +92,67 @@ export default function start(args: string[]) {
...
-		parsedArgs.will = {}
-		parsedArgs.will.topic = parsedArgs['will-topic']
-		parsedArgs.will.payload = parsedArgs['will-message']
-		parsedArgs.will.qos = parsedArgs['will-qos']
-		parsedArgs.will.retain = parsedArgs['will-retain']
+		parsedArgs['will'] = {}
+		parsedArgs['will'].topic = parsedArgs['will-topic']
+		parsedArgs['will'].payload = parsedArgs['will-message']
+		parsedArgs['will'].qos = parsedArgs['will-qos']
+		parsedArgs['will'].retain = parsedArgs['will-retain']
```
SmashingQuasar, line 130: All of these, and every access made that way can be easily resolved by using typeguards in a later PR.

src/lib/BufferedDuplex.ts
```diff
@@ -1,23 +1,37 @@
 import { Duplex, type Transform } from 'readable-stream'
 import { Buffer } from 'buffer'
-import { type IClientOptions } from './client'
+import { type IClientOptions } from './client.js'
+
+interface WritevChunksParameters {
+	chunk: unknown
+	encoding: string
+}
```
SmashingQuasar, line 8: I elected to create this very simple interface to limit cognitive load on function signature. It also simplified later typing.

src/lib/BufferedDuplex.ts
```diff
@@ -62,11 +80,15 @@ export class BufferedDuplex extends Duplex {
 		})
 	}
 
-	_read(size?: number): void {
+	public override _read(size?: number): void {
```
SmashingQuasar, line 83: The `override` keyword is mandatory here to indicate we are overriding the behaviour of a method that was already defined in the parent class. This improves maintainability.

src/lib/KeepaliveManager.ts
```diff
@@ -41,13 +42,15 @@ export default class KeepaliveManager {
...
 				: getTimer(variant)
-		this.setKeepalive(client.options.keepalive)
+		if (client.options.keepalive !== undefined) {
+			this.setKeepalive(client.options.keepalive)
+		}
 	}
 
 	private clear() {
-		if (this.timerId) {
+		if (this.timerId !== undefined) {
```
SmashingQuasar, line 51: Not strictly a TypeScript issue, but ESlint will complain if we use a strict configuration, as nullish comparisons can lead to problematic bugs.

src/lib/KeepaliveManager.ts
```diff
@@ -88,9 +91,11 @@ export default class KeepaliveManager {
 		this.counter = 0
 
 		// https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Figure_3.5_Keep
+		// @TODO: this._keepalive may be undefined. This operation will result in NaN.
 		const keepAliveTimeout = Math.ceil(this._keepalive * 1.5)
 
 		this._keepaliveTimeoutTimestamp = Date.now() + keepAliveTimeout
+		// @TODO: this._keepalive may be undefined. This operation will result in NaN.
 		this._intervalEvery = Math.ceil(this._keepalive / 2)
```
SmashingQuasar, line 99: A dangerous code that was revealed by using a strict TypeScript configuration. You have two very clear cases where those properties can unexpectedly end up as `NaN`.

src/lib/connect/ali.ts
```diff
@@ -68,17 +68,29 @@ function bindEventHandler() {
...
+				// @TODO: @robertsLando We need to handle this situation as if data is null, Buffer.from(null) will throw an Error
+				if (data === null) {
+					throw new Error('')
+				}
+
+				if (data instanceof ArrayBuffer) {
+					data = Buffer.from(data)
+				} else {
+					data = Buffer.from(data, 'utf8')
+				}
```
SmashingQuasar, line 93: Here again a strict TypeScript configuration reveals a potential error that will be uncaught and thus bubble back to our end users code. I left this `new Error('')` blank for now since I still need to understand better what to do in this situation. I'm not completely familiar with this code base.

SmashingQuasar (commented):
I added a few comments just to give more insight on what I did so far.

package.json
```diff
@@ -132,8 +132,13 @@
     "@esm-bundle/chai": "^4.3.4",
     "@release-it/conventional-changelog": "^7.0.2",
     "@types/chai": "^4.3.20",
+    "@types/concat-stream": "^2.0.3",
+    "@types/debug": "^4.1.12",
+    "@types/minimist": "^1.2.5",
```
robertsLando, line 137: I think we could drop minimis in favor of nodejs [parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig) but maybe we can do that in another PR.

src/lib/KeepaliveManager.ts
```diff
@@ -1,23 +1,24 @@
-import type MqttClient from './client'
-import getTimer, { type Timer } from './get-timer'
-import type { TimerVariant } from './shared'
+import type MqttClient from './client.js'
+import getTimer, { type Timer } from './get-timer.js'
+import type { TimerVariant } from './shared.js'
 
 export default class KeepaliveManager {
+	// @ts-expect-error - @TODO: @robertsLando: You have a situation where _keepalive can be undefined which will lead to cascading problems due to _keepalive being always used directly as a number.
```
robertsLando, line 6: could this throw if so?

src/lib/client.ts
```diff
@@ -584,7 +575,8 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			options.protocolVersion === 5 && options.customHandleAcks
 				? options.customHandleAcks
 				: (...args) => {
-						args[3](null, 0)
+						// @TODO: This would really need to be made easier to understand. It is difficult to know why the first parameter is Error | number.
+						args[3](0, 0)
```
robertsLando, line 579: I'm wondering if using 0 instead of null here may lead to some errors elsewhere? I would stick to undefined and fix the type here

src/lib/client.ts
```diff
@@ -722,6 +718,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * @return the auth packet to be returned to the broker
 	 * @api public
 	 */
+	// @ts-expect-error - packet is unused. @TODO: This method is seemingly never called.
```
robertsLando, line 721: never called? It's actually used AFAIK

src/lib/client.ts
```diff
@@ -734,6 +731,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * @param Function callback call when finished
 	 * @api public
 	 */
+	// @ts-expect-error - packet is unused. @TODO: This method is seemingly never called.
```
robertsLando, line 734: This is never used but can be overridden by user in order to handle messages with backpressure on stream (instead of using `message` event that doesn't provide backpressure)

src/lib/client.ts
```diff
@@ -852,13 +856,17 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			protocolId: this.options.protocolId,
 			protocolVersion: this.options.protocolVersion,
 			clean: this.options.clean,
-			clientId: this.options.clientId,
 			keepalive: this.options.keepalive,
 			username: this.options.username,
-			password: this.options.password as Buffer,
+			// @ts-expect-error - @TODO: @robertsLando this is a dangerous situation as this.options.password can never be a Buffer but was forced as a Buffer
+			password: this.options.password,
```
robertsLando, line 862: I know, there was a PR on mqtt-packet to fix this type don't remember right now if it has been merged or not but it was safe to force it to buffer especially because almost everyone passes it as string

src/lib/client.ts
```diff
@@ -852,13 +856,17 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
 			properties: this.options.properties,
 		}
 
+		if (this.options.clientId !== undefined) {
+			connectPacket.clientId = this.options.clientId
+		}
```
robertsLando, line 868: clientId should never be undefined, we generate a random one in case it is

src/lib/client.ts
```diff
@@ -1121,15 +1127,15 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		// force re-subscribe on reconnect. This is only true
 		// when provided `topicObject` is `this._resubscribeTopics`
 		let resubscribe = false
-		let topicsList = []
+		let topicsList: Array<string> = []
```
robertsLando, line 1130: Why not `string[]` ?

robertsLando (changes requested):

src/lib/KeepaliveManager.ts
DavideViolante: what if timerId is null?

DavideViolante:
I'd proceed more carefully on this and on a step by step basis. Maybe initially just adding strict to tsconfig and merging less changes, then adding stricter properties step by step in different PRs. Also I'd change the "module" in a different PR. Just my thoughts.

src/lib/client.ts
SmashingQuasar: Good question. This section of the code is treacherous. I think the best course of action would be to refactor it so it is simpler to understand. I am trying to limit the changes I am making to the bare minimum to have a strict TypeScript configuration so the review is easier. I'll see what I can do make this more stable.

src/lib/client.ts
SmashingQuasar: For some reason, VSCode failed to detect the call within `auth.ts`.

After having a look, here is the full method declaration:

```ts

	/**
	 * @param packet the packet received by the broker
	 * @return the auth packet to be returned to the broker
	 * @api public
	 */
	// @ts-expect-error - packet is unused. @TODO: This method is seemingly never called.
	public handleAuth(packet: IAuthPacket, callback: PacketCallback) {
		callback()
	}
```

The only place I find it used is here:

```ts

	client.handleAuth(
		packet,
		(err: ErrorWithReasonCode, packet2: IAuthPacket) => {
			if (err) {
				client.emit('error', err)
				return
			}

			if (rc === 24) {
				client.reconnecting = false
				client['_sendPacket'](packet2)
			} else {
				const error = new ErrorWithReasonCode(
					`Connection refused: ${ReasonCodes[rc]}`,
					rc,
				)
				client.emit('error', error)
			}
		},
	)
```

If we look at the code, we can see that both arguments `err` and `packet2` will both be `undefined` as the `handleAuth` implementation passes no argument to the `callback` argument invocation.

From what I can see, the code that is really being ran is:

```ts
			if (rc === 24) {
				client.reconnecting = false
				client['_sendPacket'](undefined) // packet2 is undefined so I replaced it by its constant value
			} else {
				const error = new ErrorWithReasonCode(
					`Connection refused: ${ReasonCodes[rc]}`,
					rc,
				)
				client.emit('error', error)
			}
```

I haven't had a look at `auth.ts` so I can't tell if this is intended.

src/lib/client.ts
SmashingQuasar: I agree, but in that case, the corresponding interface `IClientOptions` should reflect that and forbid the property from being `undefined`.

src/lib/client.ts
SmashingQuasar: That's a habit that I have, it's not even conscious at this point.

There is actually a good reason to this.
In TypeScript, `[]` can be used to defined two very different but related structures.
`string[]` is equivalent to `Array<string>`. This means an array of length 0 or more.
However, you can also use TypeScript tuples, which are specific arrays, for example:
`[string, ...Array<string>]` which means : "An array of strings with a minimum length of 1".

I've seen many developers quickly get confused when getting this kind of typing (I was even slightly confused while typing it):

```ts
const parsedValue: [[string, ...string[]], [string, ...string[]]];
```

I usually write:

```ts
const parsedValue: [[string, ...Array<string>], [string, ...Array<string]];
```
This helps with readability and isolates `[]` use to tuples only to reduce cognitive load.
Sub-typed structures also have a common structure in TypeScript:
```ts
const bar: Record<string, unknown> = {};

class Foo<T> {
   public setSomething(value: string): void {}
}

const myFoo: Foo<string> = new Foo<string>();

type FooSetSomething = Parameters<Foo<string>['setSomething']>;

const myMap: Map<string, string> = new Map<string, string>();
```

I find it more consistent to respect this syntax for every structure instead of creating an exception for arrays, especially if you end up with this kind of type:

```ts
const multiTypesArray: (number|string|string[])[] = [];
```

This is just an opinion. If you want to switch to `string[]` just let me know, it's not a problem.

SmashingQuasar:
> I'd proceed more carefully on this and on a step by step basis. Maybe initially just adding strict to tsconfig and merging less changes, then adding stricter properties step by step in different PRs. Also I'd change the "module" in a different PR. Just my thoughts.

We could do it that way if that makes people more comfortable. Let me know how we shall proceed and I will adapt the process accordingly.

src/lib/connect/ali.ts
SmashingQuasar: No worries, every code base has problems, even if you have 100% coverage, 100% stryker score and 0 SonarQube issue you have problems that are not clearly identified. I'm just highlighting the potential benefits of a stricter configuration.
This is actually really positive as we can improve this part since we know it has a potential issue here. 👍

SmashingQuasar:
I just pushed a commit with many new issues fixed. I haven't cleaned up the previously discussed one for now, I'll do a complete clean up at the end to avoid having to go back and forth. 225 errors left to resolve.

src/lib/connect/index.ts
```diff
@@ -142,30 +160,38 @@ function connect(
 	}
 
 	// only loads the protocols once
-	if (!protocols) {
+	// @TODO: This is very heavy on the runtime, especially when working with CJS, as you may be living in a cache-busted environment which will result in significant IOPS because of the repeated require()
+	// Consider either migrating to ESM, which is immutable and thus immune to cache busting, or simply requiring each file once.
+	// Note: Those situations are very rare, but considering how widespread MQTT is within the IoT world, it might be worth considering
+	if (protocols === undefined) {
```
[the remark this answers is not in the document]
robertsLando: considering that this happens only on client creation I don't think it's a real bottle neck for performances BTW I'm open for improvements :)

src/lib/connect/index.ts
```diff
@@ -47,30 +53,41 @@ function connect(
 	opts?: IClientOptions,
 ): MqttClient {
 	debug('connecting to an MQTT broker...')
-	if (typeof brokerUrl === 'object' && !opts) {
+	// @robertsLando: This is a dangerous evaluation of brokerUrl as typeof null === 'object' will return true
+	if (typeof brokerUrl === 'object' && opts === undefined) {
```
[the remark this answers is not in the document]
robertsLando: Yep I know that and I agree about having more strict types for this reason

src/lib/client.ts
```diff
@@ -1904,12 +1912,13 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
 		)
-		let storePacket = packet
+		let storePacket: mqttPacket.Packet = packet
 		let err: Error | undefined
 		if (storePacket.cmd === 'publish') {
 			// The original packet is for sending.
 			// The cloned storePacket is for storing to resend on reconnect.
 			// Topic Alias must not be used after disconnected.
+			// @ts-expect-error - This situation is really problematic. clone() returns a function, not an object, so this code needs to be looked at. Is this used and tested?
 			storePacket = clone(packet)
```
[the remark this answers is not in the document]
robertsLando: This is definetely an issue, I'm wondering why no test ever catched this

tsconfig.json
```diff
@@ -1,17 +1,49 @@
 {
+	"$schema": "https://json.schemastore.org/tsconfig",
 	"compilerOptions": {
-		"module": "commonjs",
+		"module": "NodeNext",
+		"moduleResolution": "NodeNext",
```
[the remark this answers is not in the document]
DavideViolante: Starting with the strictest config is not a smart idea imho, it will bring too many changes in a single PR. I'd proceed way more cautious since this library is literally used by **millions**.

SmashingQuasar:
Closing this PR to reopen it one rule at a time.

PR #2012, opened 2025-07-10
chore(types): migrate module resolution and module types

SmashingQuasar:
# Description

This PR slightly changes the TypeScript configuration as a first step to migrate towards a stricter TypeScript configuration.


src/lib/connect/ali.ts
```diff
@@ -75,11 +75,13 @@ function bindEventHandler() {
...
+				if (reader.result instanceof ArrayBuffer) {
+					proxy.push(Buffer.from(reader.result))
 
-				if (data instanceof ArrayBuffer) data = Buffer.from(data)
-				else data = Buffer.from(data, 'utf8')
-				proxy.push(data)
+					return
+				}
+
+				proxy.push(Buffer.from(reader.result, 'utf-8'))
```
SmashingQuasar, line 84: Since there was a typing issue here, I made a minor refactoring to make the code lighter.

src/lib/connect/tls.ts
```diff
@@ -10,16 +10,20 @@ const debug = _debug('mqttjs:tls')
...
-				}
-			: opts,
-	)
+	if (socksProxy !== undefined) {
+		const socket = openSocks(host, port, socksProxy, {
+			timeout: opts.socksTimeout,
+		})
+
+		// @ts-expect-error - There is a type overlap between readable-stream and stream which are used for browsers and node respectively.
+		// We would need to make it clearer what is aimed at browsers and what is aimed at Node.js to resolve this type issue.
```
SmashingQuasar, line 19: It would be very beneficial to make it clear which files are aimed at browsers and which files are aimed at Node.js within the project architecture. As it is today I find it too risky to change any interface since there are so many overlaps.

src/lib/shared.ts
```diff
@@ -9,7 +10,7 @@ export type GenericCallback<T> = (error?: Error, result?: T) => void
 
 export type VoidCallback = () => void
 
-export type IStream = Duplex & {
+export type IStream = (Duplex | NativeDuplex) & {
```
SmashingQuasar, line 13: This change actually represents the reality of how `IStream` is used. Since this is just a type, it won't affect runtime.
I added `@ts-expect-error` directives in other places so we don't break existing code.

tsconfig.json
```diff
@@ -1,17 +1,19 @@
 {
+	"$schema": "https://json.schemastore.org/tsconfig",
 	"compilerOptions": {
-		"module": "commonjs",
+		"skipLibCheck": true,
```
SmashingQuasar, line 4: Not added, simply moved up as it will be simpler for me to find which rule is still missing if I preserve the order from my reference.

tsconfig.json
```diff
@@ -1,17 +1,19 @@
 {
+	"$schema": "https://json.schemastore.org/tsconfig",
```
SmashingQuasar, line 2: No other effect than adding autocomplete and error highlighting in the IDE.

SmashingQuasar (commented):
This is a smaller version of my previous PR to migrate to a stricter TypeScript configuration. I added comments to help with the review as always.

@robertsLando could you let me know if there is a way to recognise the browser code and the Node.js code aside from looking at imported dependencies? It would probably make my life a lot easier.

robertsLando (changes requested):

src/lib/connect/ws.ts
```diff
@@ -138,18 +138,17 @@ function createBrowserWebSocket(client: MqttClient, opts: IClientOptions) {
...
 
 	const url = buildUrl(options, client)
 	const socket = createWebSocket(client, url, options)
-	const webSocketStream = Ws.createWebSocketStream(
-		socket,
-		options.wsOptions as DuplexOptions,
-	)
+	// @ts-expect-error - This is a type confusion because of the overlap between browser oriented code and Node.js oriented code.
+	// Here we are in a Node.js code since the Ws library explictly says it does not work in the browser.
+	const webSocketStream = Ws.createWebSocketStream(socket, options.wsOptions)
```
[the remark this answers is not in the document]
SmashingQuasar: So the TypeScript error is as follows:
image

If you look at the signature of `Ws.createWebSocketStream()`, you will see that it accepts: `internal.DuplexOptions` as second parameter, which is the `DuplexOptions` type from `node:stream`.
On the other hand, we are passing a `IClientOptions`, which has a `wsOptions` property with the following typing: `ClientOptions | ClientRequestArgs | DuplexOptions`.
The problem here is that the `DuplexOptions` from `IClientOptions` comes from `readable-stream`, while the `DuplexOptions` of `Ws.createWebSocketStream()` comes from `node:stream`. So there is a naming overlap between `node:stream:DuplexOptions` and `readable-stream:DuplexOptions`.

`ClientOptions.wsOptions` can be of three different types, we would need to create type guards to narrow down the type and ensure it is indeed a `readable-stream:DuplexOptions`. I can absolutely implement this, but this will affect runtime and reject any wrongly typed option, which will have possible effects on users. Since we want to have a cautious approach, I elected to use the safest course of action and not touch the code at runtime.

tsconfig.json
```diff
@@ -1,17 +1,19 @@
...
 	"compilerOptions": {
-		"module": "commonjs",
+		"skipLibCheck": true,
+		"module": "NodeNext",
+		"moduleResolution": "NodeNext",
+		"target": "ESNext",
```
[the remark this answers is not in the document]
SmashingQuasar: @robertsLando I just did that, the problem is that `@tsconfig/node20` is designed to work with Node.js 20, not with a mix of Node.js 20 and browsers API. For example the `FileReader` class whihc is used in the `ali.ts` file (which I assume is browser exclusive) can never be resolved

src/lib/connect/ws.ts
```diff
@@ -138,18 +138,17 @@ function createBrowserWebSocket(client: MqttClient, opts: IClientOptions) {
...
 	options.hostname = options.hostname || options.host || 'localhost'
 
 	const url = buildUrl(options, client)
 	const socket = createWebSocket(client, url, options)
-	const webSocketStream = Ws.createWebSocketStream(
-		socket,
-		options.wsOptions as DuplexOptions,
-	)
+	// @ts-expect-error - This is a type confusion because of the overlap between browser oriented code and Node.js oriented code.
+	// Here we are in a Node.js code since the Ws library explictly says it does not work in the browser.
```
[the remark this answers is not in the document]
SmashingQuasar: I can remove it for sure.
I just don't find it obvious at all that we are in a Node.js context only within this file, especially since we import `isBrowser` and have a condition on this on another section of the same file.

SmashingQuasar:
I removed the comment in question.
So for the test coverage, it complains that there are untested lines, but this is because when I change the `socksProxy` evaluation, it stopped considering this line as tested, because it was only partially tested before.
There does not seem to be a test for the `socksProxy` option.
I would like to add one so we preserve the coverage, however, I am unsure what the intent would be. Since this is supposed to test a proxy, it would require running a pseudo-proxy to test it. That's doable but what's your policy here?

SmashingQuasar:
> @SmashingQuasar We already have tests for socks with a mocked socks server here: https://github.com/mqttjs/MQTT.js/blob/main/test/node/socks.ts
> 
> Maybe there is just no test that inits a client with a sock proxy and I think that if that line isn't covered by test it was not even before. Would you like to add a simple test for that?

Hey @robertsLando , sorry for not answering earlier, the end of the week was quite busy.
Indeed I think it would be best to add a test for this case. I am just wondering if we shouldn't wait until we do the architectural changes, since they will most likely affect tests.
Since this line was not tested before, it would be silly to create a test, just to probably completely rewrite it not very long after.
What do you think?

robertsLando (approved):

PR #2013, opened 2025-07-17
chore(tsconfig): Added stricter TypeScript without direct effect on the codebase.

SmashingQuasar:
# Description

This PR adds the following TSConfig rules which this code base already complies with but are part of a strict TypeScript approach:

- [`alwaysStrict`](https://www.typescriptlang.org/tsconfig/#alwaysStrict): Ensures that your files are parsed in the ECMAScript strict mode, and emit “use strict” for each source file. **Changed from false to true.**
- [`allowUnusedLabels`](https://www.typescriptlang.org/tsconfig/#allowUnusedLabels): Since labels are rarely used, when an unused label is detected, it's most likely a typo where someone wanted to assign a value. **Added.**
- [`allowUnreachableCode`](https://www.typescriptlang.org/tsconfig/#allowUnreachableCode): Helpful to detected potentially unreachable code. This is better handled through ESlint but this rule does not hurt. **Added.**
- [`noErrorTruncation`](https://www.typescriptlang.org/tsconfig/#noErrorTruncation): This is useful only in development. When TypeScript displays an error, by default it truncates the type errors. This prevents this behaviour and shows the entire type error. **Added.**
- [`emitDecoratorMetadata`](https://www.typescriptlang.org/tsconfig/#emitDecoratorMetadata): Emitting metadata to decorators adds generally unnecessary code. **Changed from true to false.**
  - On a more personal note: Decorators are a bad practice. They are a clumsy attempt at implementing traits and mixins in JavaScript. Unfortunately, they are done through prototype hijacking and result in low quality computed code with non-null performance overhead. (~7% per decorator during my last benchmark a few years ago) They also make the intent of the code less clear and rely more on magic.
- [`experimentalDecorators`](https://www.typescriptlang.org/tsconfig/#experimentalDecorators): Disables using TypeScript decorators. Since decorators aren't even officially implemented within the TC39, TypeScript has done a custom implementation. Decorators might never see the light of day in JavaScript, and if they do, they will most likely to be implemented in the same way TypeScript did. Overall, another reason to not use decorators. **Changed from true to false.**
- [`forceConsistentCasingInFileNames`](https://www.typescriptlang.org/tsconfig/#forceConsistentCasingInFileNames): Very important rule. Linux-based operating systems use a case-sensitive filesystem, which means a file named "Foo.ts" and a file named "foo.ts" located in the same directory are not the same file. Some operating systems, notably Windows, do not use a case-sensitive filesystem. This means both files will be considered the same. So importing "Foo.ts" with `import { x } from "foo.ts"` will work on Windows, not on Linux. While I don't think anyone really develops directly on Windows, better safe than sorry. **Added.**
- [`checkJs`](https://www.typescriptlang.org/tsconfig/#checkJs): There is no point having this on since we are not supposed to have JavaScript files. **Added.**
- [`allowJs`](https://www.typescriptlang.org/tsconfig/#allowJs): Allows mixing Javascript and TypeScript files. Since we shouldn't have Javascript files, this is set to false. **Added.**
- [`isolatedModules`](https://www.typescriptlang.org/tsconfig/#isolatedModules): This is quite a long rule to explain, which will be better detailed in the official handbook. The reason that I am turning it off, is because we want to be able to use `const enum` and have them being replaced by their actual values to lower memory footprint at runtime. **Added.**
- [`preserveConstEnums`](https://www.typescriptlang.org/tsconfig/#preserveConstEnums): If this is set to true, then `const enum` are preserved in their object form. This defeats the point of using `const enum` and increases memory footprint. **Added.**
- [`newLine`](https://www.typescriptlang.org/tsconfig/#newLine): Forces TypeScript to emit files and use `LF` as a new line delimiter. Since `LF` is generally recognised as a new line delimited in modern systems but `CR` is not, it is better to configure it that way. **Added.**
- [`pretty`](https://www.typescriptlang.org/tsconfig/#pretty): Purely for development purposes. It makes the error outputted by TypeScript use colours and be more readable in general. **Added.**


SmashingQuasar:
@robertsLando I added many rules at the same time, since they did not trigger any error from TypeScript. In my opinion it was pointless to add them one by one in a dedicated PR since they are already functional.

PR #2023, opened 2025-09-02
fix: isWebWorkerEnv check for Deno environment

alex-controlx:
# Deno Worker Fix Explanation

## The Problem

The issue was that MQTT.js was incorrectly identifying Deno workers as browser environments, which caused problems when trying to use TCP connections. Here's what was happening:

1. **Original `isWebWorkerEnv()` logic**: The function was checking if the environment was a Web Worker by looking for `self?.constructor?.name?.includes('WorkerGlobalScope')`

2. **Deno Worker confusion**: In Deno workers, this condition was returning `true` because Deno workers do have a `WorkerGlobalScope`-like environment (`DedicatedWorkerGlobalScope` to be exact)

3. **Incorrect browser detection**: Since `isWebWorkerEnv()` returned `true`, the overall `isBrowser` check also returned `true` (line 37: `isBrowser = isStandardBrowserEnv() || isWebWorkerEnv() || isReactNativeEnv()`)

4. **TCP connection blocked**: When `isBrowser` was `true`, the MQTT connection logic would force WebSocket connections (`ws://` or `wss://`) and didn't allow TCP protocols (`mqtt://` or `mqtts://`)

## The Solution

By adding `typeof Deno === "undefined"` to the `isWebWorkerEnv()` function, you're ensuring that:

- **Deno workers are NOT considered web workers**: The function now returns `false` for Deno workers, even though they have a `WorkerGlobalScope`-like environment
- **Real web workers still work**: Browser-based web workers don't have `Deno` defined, so they still get detected correctly
- **TCP connections work in Deno**: Since `isBrowser` now returns `false` for Deno workers, the connection logic allows TCP connections (`mqtt://`, `mqtts://`) instead of forcing WebSocket connections

## Why This Matters

Deno workers need to be able to use TCP connections for MQTT because:
- They run in a server-like environment (not a browser)
- TCP connections are more efficient for MQTT than WebSocket connections
- The original browser detection was preventing legitimate TCP usage in Deno

The fix ensures that Deno workers are treated as server environments (like Node.js) rather than browser environments, allowing them to use the appropriate connection method for MQTT protocols.

### DIFF
--- a/src/lib/is-browser.ts
+++ b/src/lib/is-browser.ts
@@ -1,3 +1,8 @@
+// Global type declaration for Deno
+declare global {
+	const Deno: any
+}
+
 const isStandardBrowserEnv = () => {
 	// window is only defined when it is a browser
 	if (typeof window !== 'undefined') {
@@ -28,8 +33,8 @@ const isStandardBrowserEnv = () => {
 const isWebWorkerEnv = () =>
 	Boolean(
 		typeof self === 'object' &&
-    self?.constructor?.name?.includes('WorkerGlobalScope') &&
-    typeof Deno === "undefined",
+			self?.constructor?.name?.includes('WorkerGlobalScope') &&
+			typeof Deno === 'undefined',
 	)
 
 const isReactNativeEnv = () =>


robertsLando:
@alex-controlx You have to define Deno as global otherwise lint fails

alex-controlx:
used as example from 
https://github.com/mqttjs/MQTT.js/blob/7e2bfe461b3aaf07650eefcdd4ddeafc3968852c/src/lib/connect/wx.ts#L13

acedward:
This is great, I just was testing MQTT with deno, and saw this show up 🔥

alex-controlx:
MQTT.js works well in Deno, it was just a restriction on the usage of all protocols from Deno workers.

Issue #2038, opened 2026-02-18
Roadmap: MQTT.js v6 — Zero/minimal dependencies, tree-shakeable transports, reduced bundle size


robertsLando:
## Summary

This issue outlines the implementation roadmap for **MQTT.js v6**, a major release focused on reducing dependencies, improving tree-shakability, reducing browser bundle size, and supporting multiple JavaScript runtimes — while maintaining or improving current performance.

This builds on the discussion in #2002 and depends on [mqtt-packet v10](https://github.com/mqttjs/mqtt-packet/issues/159) (a zero-dependency, TypeScript-native rewrite).

## Current State (v5.x)

| Metric | Value |
|--------|-------|
| Production dependencies | ~75 packages (transitive) |
| Browser bundle (minified) | 368 KB |
| `sideEffects` in package.json | Not declared |
| Tree-shakeable transports | No (runtime `require()` with `isBrowser` check) |
| Runtime support | Node.js, Browser (via polyfills) |

### Dependency Breakdown

The heaviest chains today:

| Chain | Packages | Issue |
|-------|----------|-------|
| `mqtt-packet` → `bl` → `readable-stream` + `buffer` | ~12 | Core codec pulls stream polyfills |
| `readable-stream` → `buffer`, `events`, `process`, `abort-controller` | ~8 | Node stream polyfill for browser |
| `worker-timers` → `@babel/runtime`, `tslib`, `broker-factory`, `worker-factory` | ~10 | Timers in web workers |
| `socks` → `ip-address`, `smart-buffer` | ~4 | SOCKS proxy (Node-only) |
| `concat-stream`, `split2`, `commist`, `help-me`, `minimist` | ~8 | CLI-only deps shipped to all consumers |
| `number-allocator` → `js-sdsl` | ~2 | Full data structures library for simple ID allocation |

## Target State (v6)

| Metric | Target |
|--------|--------|
| Production dependencies | ~15-20 packages |
| Browser bundle (minified) | ~80-120 KB |
| Browser bundle format | **ESM only** (drop IIFE) |
| Tree-shakeable transports | Yes (separate packages in monorepo) |
| Runtime support | Node.js, Browser, Deno, Bun |
| Minimum Node.js version | **24+** |
| `sideEffects` | Declared `false` |
| Repository structure | **Monorepo** (npm workspaces + Nx) |

---

## Decisions (from discussion)

Based on feedback from @mcollina and @jacoscaz:

| Question | Decision | Rationale |
[...]

PR #2041, opened 2026-02-26
chore: replace number-allocator with inline Set-based implementation

roli-lpci:
## Summary

Replaces the `number-allocator` package (which depends on `js-sdsl` and `debug`) with an 80-line inline implementation using a `Set` and a low-water-mark cursor.

This addresses **item 1.1** of the v6 dependency cleanup roadmap (#2038).

- Removes 3 transitive dependencies (`number-allocator`, `js-sdsl`, `debug`)
- All existing tests pass (123 integration + 10 unit)
- 19 new dedicated unit tests for the inline `NumberAllocator`
- Browser-compatible (`Set` is ES2015)

## Changes

| File | Change |
|------|--------|
| `src/lib/number-allocator.ts` | **New** — inline Set-based allocator (~80 lines) |
| `src/lib/unique-message-id-provider.ts` | Import path change, removed unnecessary `as boolean` cast |
| `src/lib/topic-alias-send.ts` | Import path change |
| `package.json` | Removed `number-allocator` dependency |
| `package-lock.json` | Regenerated |
| `test/node/number-allocator.ts` | **New** — 19 dedicated unit tests |

## Implementation

The inline `NumberAllocator` tracks used numbers in a `Set<number>` and maintains a `lowWaterMark` cursor that avoids full scans in the common sequential-allocation case. For MQTT's 16-bit ID range (1–65535), this is more than sufficient.

API is identical to the original: `alloc()`, `use()`, `free()`, `clear()`, `firstVacant()`.

## Test plan

- [ ] Existing message-id-provider tests pass (10/10)
- [ ] New NumberAllocator unit tests pass (19/19)
- [ ] Integration tests with UniqueMessageIdProvider pass (123/123)
- [ ] TypeScript compiles clean
- [ ] Browser bundle builds successfully


robertsLando (approved):
LGTM

package.json
```diff
@@ -98,7 +98,6 @@
     "lru-cache": "^10.4.3",
     "minimist": "^1.2.8",
     "mqtt-packet": "^9.0.2",
-    "number-allocator": "^1.0.14",
     "readable-stream": "^4.7.0",
     "rfdc": "^1.4.1",
     "socks": "^2.8.6",
```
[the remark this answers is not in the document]
robertsLando: Ignore

src/lib/number-allocator.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * Mark a specific number as occupied.
+	 * @returns true if the number was vacant (now occupied), false if already occupied or out of range.
+	 */
+	use(num: number): boolean {
+		if (num < this.min || num > this.max || this.used.has(num)) {
+			return false
+		}
+		this.used.add(num)
```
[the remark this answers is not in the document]
robertsLando: I think this is a legit change

robertsLando:
@roli-lpci could you please fix conflicts ?

roli-lpci:
@robertsLando conflicts resolved in 88cf300 — merged current main and regenerated package-lock.json (number-allocator and js-sdsl dropped). Allocator and message-id-provider tests pass locally (29/29). Ready when you are.

src/lib/number-allocator.ts
```diff
@@ -0,0 +1,86 @@
...
+		this.max = max
+		this.used = new Set()
+		this.lowWaterMark = min
+	}
+
+	/**
+	 * Get the first vacant number without allocating it.
+	 * @returns The first vacant number, or null if all are occupied.
+	 */
+	firstVacant(): number | null {
```
robertsLando, line 30: **[Minor] · Performance**

`firstVacant()` is O(max) on the `TopicAliasSend` path. That consumer only ever calls `use()` + `firstVacant()` (never `alloc()`/`free()`), so `lowWaterMark` stays pinned at `min` forever and every `getLruAlias()` re-scans from 1. Once the alias table is saturated — the hot case for `getLruAlias`, which then falls through to LRU eviction — each publish-to-new-topic does a full failing 1→max scan, a regression from the library's O(1).

**Why:** with a broker-negotiated `topicAliasMaximum` in the thousands and a client publishing many distinct topics, this is a per-publish cost on a hot path.

**Fix:** early-return `null` when `used.size === max - min + 1` — cheap, and it collapses exactly the saturated-scan case (also kills the `alloc()` worst case below).

src/lib/number-allocator.ts
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+		return null
+	}
+
+	/**
+	 * Allocate the first vacant number. It becomes occupied.
+	 * @returns The first vacant number, or null if all are occupied.
+	 */
+	alloc(): number | null {
```
robertsLando, line 46: **[FYI] · Performance**

Worst-case `alloc()` is a full-range scan (up to 65535 `Set.has`) after a `free(low)` + `use(low)` sequence that leaves `lowWaterMark` low while the numbers above it stay occupied. It's one-time (the next `alloc` re-advances the mark) and sub-millisecond, so acceptable for MQTT — noting only because the original was O(log n).

**Why:** bounded and rare; not a real concern at MQTT's scale.

**Fix:** the same `used.size === max - min + 1` guard suggested on `firstVacant()` also removes this case.

src/lib/number-allocator.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	private min: number
+
+	private max: number
+
+	// Optimization: tracks the lowest number that *might* be vacant,
+	// avoiding full scans in the common sequential-allocation case.
+	private lowWaterMark: number
+
+	constructor(min: number, max: number) {
```
robertsLando, line 19: **[FYI] · Design**

The constructor validates neither `min <= max` nor integer inputs. Not a regression — the original library didn't either, and both call sites pass safe values (`1, 65535` and a guarded `1, max`).

**Why:** flagging only for completeness; no action needed for this PR.

**Fix:** optional — a cheap `if (min > max) throw ...` guard would make the class more robust as a standalone utility if it ever gets reused.

robertsLando (commented):
Replacing `number-allocator` with a self-contained ~86-line inline module is a sound call: it drops two transitive deps (`js-sdsl`, `debug`) and the reimplementation is **behaviorally equivalent** to the original's public API (`firstVacant`/`alloc`/`use`/`free`/`clear`) for all in-range operations — I diffed it against the library source. Nice touch dropping the now-native `as boolean` cast in `unique-message-id-provider`.

**Correctness: clean.** The `lowWaterMark` optimization holds the invariant "all numbers in `[min, lowWaterMark)` are occupied" across every method (`alloc` advances it, `free` pulls it back, `use` correctly leaves it untouched), so a real vacancy is never skipped and `null` is only returned when genuinely full. It's actually *safer* than the original at the edges — double-`free()` is a clean no-op here, vs. undefined interval-merge behavior in the library.

**Strengths:** thorough unit tests (exhaustion, peek-vs-consume, out-of-range, idempotent double-free, full 1–65535 recovery); the two consumers keep their existing regression coverage (`message-id-provider.ts`, `unique_message_id_provider_client.ts`, `client_mqtt5.ts`); no bundler `external` change needed since inline TS bundles automatically.

**Top risks (both non-blocking, both Minor/FYI perf):** 1) `firstVacant()` is O(max) on the topic-alias path; 2) worst-case `alloc()` full-range scan. A single `used.size` guard neutralizes both — see the two line comments. Neither blocks shipping.

**Tests (FYI):** the new suite covers `NumberAllocator` directly and thoroughly; no additional test needed to ship. A case mirroring `TopicAliasSend`'s `use()`-only + `firstVacant()` pattern would guard the perf note below, but it's optional.

@roli-lpci please sort out this minor findings then this is good to be merged

roli-lpci:
Implemented the requested saturated-range fast path in `d36a00b` for both `firstVacant()` and `alloc()`, with a focused `use()`-only regression asserting zero range scans. Checks: allocator tests 20/20, lint, TypeScript build, and browser build pass; the full Node run executed 668 assertions with 0 failures (its existing coverage-aggregation threshold failure reproduces on the prior `88cf300` head). Fresh exact-base/head review: PASS, 0 findings.

robertsLando (approved):

PR #2051, opened 2026-06-28
fix(connect/tls): don't overwrite an explicit servername (SNI)

rohankapoorcom:
## Problem

When connecting over TLS, `buildStream` unconditionally sets the SNI `servername` to the connect host:

```js
if (net.isIP(opts.host) === 0) {
    opts.servername = opts.host
}
```

This silently overwrites a `servername` the caller passed in. Since `servername` is also the hostname Node uses to verify the certificate, connecting to a host that differs from the cert's hostname fails even when the caller correctly set SNI:

```
Hostname/IP does not match certificate's altnames: Host: emqx-mqtt.emqx.svc.cluster.local. is not cert's CN: mqtt.dev.example.com
```

This is common in Kubernetes/ingress (and Docker-alias) setups: you connect to an internal address but the broker's cert is issued for a public/virtual domain. Per Node's `tls.connect` docs, `servername` is meant to support exactly this (it can differ from the TCP target).

## Fix

Only default `servername` to the host when the caller hasn't set one (IP hosts are still skipped per RFC 6066):

```js
if (net.isIP(opts.host) === 0 && !opts.servername) {
    opts.servername = opts.host
}
```

Also updated the `servername` JSDoc, which previously said *"Should be set to `host`"* — conflicting with Node semantics and the original intent of #1049.

## Behavior

- No `servername` + DNS host → SNI defaults to host (unchanged).
- Connect by IP → never used as SNI (unchanged).
- Explicit `servername` ≠ host → now honored; cert verified against `servername`.

## Tests

Added to `test/node/secure_client.ts`:
- explicit `servername` over a DNS host → SNI is the user value;
- explicit `servername` preserved when connecting by IP;
- cert verification succeeds against an explicit `servername` when the connect host doesn't match the cert (new fixture cert valid only for `mqtt.example.test`).

## References

- Downstream consumer adding `mqtt.server_name` → `options.servername`: https://github.com/Koenkk/zigbee2mqtt/pull/32417
- Same cert-hostname mismatch reported earlier: https://github.com/Koenkk/zigbee2mqtt/issues/19450
- Real-world workaround documenting this exact limitation in v5.15.1 (`MQTT_SERVERNAME` ignored; internal Docker alias + public-domain cert): https://github.com/ospp-org/csms-mqtt-bridge
- Prior art: https://github.com/mqttjs/MQTT.js/pull/1049 introduced the unconditional assignment; https://github.com/mqttjs/MQTT.js/pull/954 (reverted) had override tests

Made with [Cursor](https://cursor.com)


robertsLando (commented):
Reviewed the SNI fix — approach is correct: defaulting `servername` only when unset (and never for IPs) matches Node's `tls.connect` semantics where `servername` drives both SNI and cert verification. JSDoc correction is accurate, and the cert-mismatch test (`rejectUnauthorized: true` against a cert valid only for `mqtt.example.test`) genuinely exercises the fix. One minor nit inline; otherwise good to ship.

test/node/secure_client.ts
```diff
@@ -11,9 +11,14 @@ import getPorts from './helpers/port_list'
 const ports = getPorts(5)
 
 const port = ports.PORT
+const sniPort = ports.PORTAND400
```
robertsLando: **[Minor] · Tests**

`getPorts(i)` partitions the port space into 400-wide blocks (`PORT = 10000 + i*400`), and every other key stays within `PORT..PORT+327`. `PORTAND400` (= `PORT+400`) is the **base port of the next index's block** — `getPorts(5)` → `12400`, which is exactly `getPorts(6).PORT`.

**Why:** No collision today (index 6 is unused), but it's a silent footgun — the first person to add `getPorts(6)` gets an intermittent port collision that won't be obvious.

**Fix:** Pick a key inside this file's own block. `secure_client.ts` only consumes `ports.PORT`, so an in-block key like `PORTAND40` is free.

rohankapoorcom: Thanks for the explanation - I totally misread that implementation detail. Corrected.

robertsLando (approved):

PR #2052, opened 2026-08-01
fix: fix reconnection failures, PINGRESP blocking, and unstable PINGREQ interval

majinju:
## Summary

This PR fixes three critical issues in the MQTT.js client that cause connection drops, keepalive timeouts, and reconnection failures when the broker restarts or under heavy message load.

## Problem 1: Reconnection failures after broker restart

When the MQTT broker restarts or crashes, the client may get stuck in a "connecting" state and never automatically reconnect. This is caused by multiple bugs in the reconnection mechanism:

1. **`_cleanUp` condition error**: `!this.reconnecting` prevents `reconnectTimer` from being set during auto-reconnection. Since `stream.destroy()` is a no-op on a closed stream, the `close` event never fires, and `reconnectTimer` is permanently lost.

2. **`disconnecting` flag not reset on unexpected disconnect**: `end()` sets `disconnected=true`, but on unexpected disconnects, `disconnected` remains `false`, so `disconnecting` must be reset to allow reconnection.

3. **`connackTimer` timeout missing fallback**: `_cleanUp(true)` calls `stream.destroy()` which may not trigger `close` event if stream is already closed. Added explicit `_setupReconnect` fallback.

4. **`onKeepaliveTimeout` missing fallback**: Same issue as connackTimer timeout.

5. **CONNACK error codes don't trigger reconnection**: Added `reconnectOnConnackError: true` to default options.

6. **`connect` event delayed by outgoingStore**: `emit('connect')` was deferred until `outgoingStore` finished processing. If store processing stalled, the event was never emitted, causing clients (e.g., Node-RED) to stay in "connecting" state.

## Problem 2: PUBLISH handling blocking PINGRESP

The stream backpressure mechanism binds packet serial processing to the `writable._write` `done()` callback. When a PUBLISH message's `handleMessage` callback is slow or not called, `done()` is never invoked, causing the stream to pause reading. This prevents subsequent TCP packets (including PINGRESP) from being parsed, leading to keepalive timeout and connection drops.

**Fix**: Handle PINGRESP directly in `parser.on('packet')`, bypassing the serial queue. Decouple stream backpressure from packet processing by calling `done()` immediately after `parser.parse(buf)`.

## Problem 3: Unstable PINGREQ interval

`KeepaliveManager.reschedule()` resets the entire timer when PINGRESP is received. If PINGRESP arrives late, the next PINGREQ is also delayed, causing unstable heartbeat intervals (60s → 78s → 60s).

**Fix**: Added `_waitingForPingResp` flag. On PINGRESP receipt, only reset the flag instead of rescheduling the timer. Timer continues running independently, PINGREQ sent at fixed keepalive intervals.

### Problem 1 - Reconnection
- **Node-RED (mqtt.js 5.x)**: Broker restart detected in 10s, auto-reconnect succeeded in 85s
- **Before fix**: Client stuck in "connecting" forever

### Problem 2 - PINGRESP blocking
- **Mock broker test** (keepalive=10s, handleMessage blocks 15s):
  - Before fix: PINGRESP blocked, keepalive timeout at 15s, pingresps=1
  - After fix: PINGRESP received every 10s, no timeout in 70s, pingresps=6

### Problem 3 - PINGREQ interval
- **Mock broker test** (keepalive=60s, PINGRESP delayed 18s):
  - Before fix: PINGREQ intervals 60s → 78s → 60s (unstable)
  - After fix: PINGREQ intervals 60.0s → 60.0s → 60.0s (stable)

## Files Changed

- `src/lib/client.ts`: Reconnection fixes, PINGRESP direct handling, stream backpressure decoupling, connect event timing
- `src/lib/KeepaliveManager.ts`: New `_waitingForPingResp` flag and `resetWaitingState()` method

## Backward Compatibility

All changes are backward compatible:
- `reconnectOnConnackError` defaults to `true` (can be overridden)
- PINGRESP direct handling doesn't affect other packet types
- Timer behavior change only affects PINGREQ timing (more stable)

### DIFF
--- a/src/lib/KeepaliveManager.ts
+++ b/src/lib/KeepaliveManager.ts
@@ -19,6 +19,10 @@ export default class KeepaliveManager {
 
 	private _intervalEvery: number
 
+	// 修复：跟踪是否有未确认的 PINGREQ
+	// 避免 PINGRESP 延迟导致 reschedule，进而导致下一次 PINGREQ 延迟
+	private _waitingForPingResp: boolean
+
 	/** Timestamp of next keepalive timeout */
 	get keepaliveTimeoutTimestamp() {
 		return this._keepaliveTimeoutTimestamp
@@ -41,6 +45,7 @@ export default class KeepaliveManager {
 			'clear' in variant
 				? variant
 				: getTimer(variant)
+		this._waitingForPingResp = false
 		this.setKeepalive(client.options.keepalive)
 	}
 
@@ -81,6 +86,7 @@ export default class KeepaliveManager {
 
 		this.clear()
 		this.counter = 0
+		this._waitingForPingResp = false
 
 		// https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Figure_3.5_Keep
 		const keepAliveTimeout = Math.ceil(this._keepalive * 1.5)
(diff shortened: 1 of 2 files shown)


majinju:
还要优化下

PR #2054, opened 2026-08-30
fix: schedule browser process.nextTick with queueMicrotask

### LINKED ISSUE #2053: dist/mqtt.esm.js's bundled process.nextTick uses setTimeout(0), silently stalling cork()/uncork() writes when setTimeout is throttled (React Native background)
## Environment
- `mqtt` 5.15.2
- React Native (Android), custom `streamBuilder` backed by a native `okhttp3.WebSocket`
- Resolved via package's `exports["."]["react-native"]` condition → `dist/mqtt.esm.js`
- `mqtt.Client` constructed directly, bypassing `mqtt.connect()`

## Summary

When the app is backgrounded, MQTT writes silently stop reaching the transport's `_write()` - `writeToStream()` returns normally, but no error is ever thrown or emitted, and the bytes never arrive. Reproduces on both an already-connected stream (e.g. a routine `pingreq`) and freshly-created reconnect streams.

## Root cause

`mqtt-packet`'s `writeToStream.js` corks the stream before writing a packet, then schedules `uncork()` via `process.nextTick`. In `dist/mqtt.esm.js`, that resolves to a bundled `process/browser.js`-style shim whose queue-drain schedules via plain `setTimeout(fn, 0)`:

```js
function drainQueue(){
  if (!draining) {
    var timeout = setTimeout(cleanUpNextTick, 0)
    draining = true
    /* ...runs queued items... */
    draining = false; clearTimeout(timeout)
  }
}
function nextTick(fun){
  queue.push(new Item(fun, args))
  if (queue.length === 1 && !draining) setTimeout(drainQueue, 0)
}
```

RN throttles/suspends plain `setTimeout` while backgrounded, so `uncork()` never runs, corked bytes never reach `_write()`, and the client silently stops transmitting - `writeToStream()`'s synchronous return doesn't reflect whether the deferred flush ever happens. Confirmed specific to this shim, not general JS timer throttling (a control `setImmediate` driven by a native timer kept firing at single-digit-ms latency throughout the same stall).

## Reproduction

1. Connect with any transport whose `streamBuilder` is a plain `Duplex`.
2. Background the app (or otherwise suspend `setTimeout` while some other mechanism keeps triggering `mqtt.js` activity, e.g. keepalive `pingreq`).
3. `writeToStream` logs completing normally, but `_write()` is never invoked - `cork()` fires, `uncork()` never does.

## Suggested fix (verified working)

Schedule via `queueMicrotask` instead, falling back to `setTimeout` if unavailable:
```js
function schedule(fn) {
  typeof queueMicrotask === 'function' ? queueMicrotask(fn) : setTimeout(fn, 0)
}
```
at both call sites above.

Verified on-device via a local patch of this exact file: backgrounded for 2+ minutes, both idle and mid-call with an app-level keepalive layered on top (its own ~20s keepalive over the same MQTT connection) - `pingreq`/`pingresp` round-tripped on schedule (~100-140ms) throughout, no stalls, no timeouts.

Happy to share logcat captures or bundle excerpts if useful.

official-burak:
## Summary

The React Native and browser bundle polyfilled `process.nextTick` with `setTimeout(0)`. `mqtt-packet` corks the transport and uncorks on nextTick, so when React Native throttles timers in the background, `writeToStream()` returns normally while `_write()` never runs.

The browser build now resolves `process` to a small shim that drains via `queueMicrotask` (with a `setTimeout` fallback). Node keeps using the real `process.nextTick`.

Fixes #2053

## Test plan

- [x] `test/node/browser-process.ts`: nextTick is a microtask, forwards extra args, rejects a non-function, and uncorks a stream without `setTimeout`
- [x] `npm run lint`
- [x] `npm run unit-test:node` (698 pass, 0 fail)
- [ ] Maintainer: confirm a backgrounded RN client still sends pingreq while timers are throttled

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -18,6 +18,20 @@ const options = {
     globalName: 'mqtt',
     sourcemap: false, // this can be enabled while debugging, if we decide to keep this enabled we should also ship the `src` folder to npm
     plugins: [
+        {
+            name: 'browser-process',
+            setup(build) {
+                // mqtt-packet corks the stream and uncorks on process.nextTick.
+                // The default browser process shim drains that queue with
+                // setTimeout(0), which React Native throttles while
+                // backgrounded, so writes stall. Resolve process to a
+                // queueMicrotask-based shim instead (issue #2053).
+                const shim = require('path').resolve(__dirname, 'src/lib/browser-process.ts')
+                build.onResolve({ filter: /^(node:)?process$/ }, () => ({
+                    path: shim,
+                }))
+            }
+        },
         polyfillNode({
             polyfills: [
                 'readable-stream'
--- a/src/lib/browser-process.ts
+++ b/src/lib/browser-process.ts
@@ -1,0 +1,91 @@
+const pending: Array<{
+	fun: (...args: any[]) => void
+	args: any[]
+}> = []
+let draining = false
+
+function runNextTicks() {
+	draining = true
+	while (pending.length) {
+		const item = pending.shift()
+		item.fun(...item.args)
+	}
+	draining = false
+}
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
+		queueMicrotask(runNextTicks)
+	} else {
+		setTimeout(runNextTicks, 0)
+	}
+}
+
+export function nextTick(fun: (...args: any[]) => void, ...args: any[]): void {
+	if (typeof fun !== 'function') {
+		throw new TypeError('"callback" argument must be a function')
+	}
+	pending.push({ fun, args })
+	if (pending.length === 1 && !draining) {
+		scheduleDrain()
+	}
+}
+
+function noop() {}
+
+export const title = 'browser'
+export const browser = true
+export const env: Record<string, string | undefined> = {}
+export const argv: string[] = []
+export const version = ''
+export const versions: Record<string, string | undefined> = {}
+export const on = noop
+export const addListener = noop
+export const once = noop
+export const off = noop
+export const removeListener = noop
+export const removeAllListeners = noop
+export const emit = noop
+export const prependListener = noop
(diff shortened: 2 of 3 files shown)


robertsLando:
Thanks for the fix @official-burak! Closing in favour of #2056, which fixes the same issue (#2053) with the same approach but also resolves `process/` (trailing slash).

That matters here: `readable-stream` imports the shim as `require('process/')` in 7 modules including `internal/streams/writable.js`, which is the one doing the cork/uncork. The `/^(node:)?process$/` filter in this PR doesn't match those, so the stalling code path would keep the old `setTimeout(0)` shim.

Your work is fully credited in the discussion — sorry for the duplicate effort.

robertsLando:
Follow-up now that #2056 has been reviewed in depth — recording the evidence here so the reasoning is on the record.

I built both branches to compare rather than judging from the diff:

`readable-stream` imports the process shim as `require('process/')` — with a trailing slash — in 7 modules:

```
lib/internal/streams/writable.js:30
lib/internal/streams/readable.js:26
lib/internal/streams/destroy.js:5
lib/internal/streams/duplexify.js:3
lib/internal/streams/end-of-stream.js:8
lib/internal/streams/pipeline.js:3
lib/internal/streams/from.js:5
```

`writable.js` is the one doing the cork/uncork that stalls under React Native backgrounding. The filter here, `/^(node:)?process$/`, doesn't match `'process/'`, so those 7 modules would keep resolving to the old `setTimeout(0)` shim and the stall would remain. #2056 uses `/^(node:)?process\/?$/` and does match — in its build, all 7 sites bind the new shim.

Everything else was equivalent, so this was the deciding difference. Nothing wrong with the work — it's a real fix that just missed one specifier. Thanks again @official-burak, and sorry for the duplicated effort.

PR #2056, opened 2026-09-01
fix: schedule browser process.nextTick with queueMicrotask

### LINKED ISSUE #2053: dist/mqtt.esm.js's bundled process.nextTick uses setTimeout(0), silently stalling cork()/uncork() writes when setTimeout is throttled (React Native background)

sankalpsthakur:
Fixes https://github.com/mqttjs/MQTT.js/issues/2053.

The React Native and browser bundle polyfilled `process.nextTick` with `setTimeout(0)`. `mqtt-packet` corks the transport and uncorks on nextTick, so when React Native throttles timers in the background, `writeToStream()` returns normally while `_write()` never runs.

The browser build now resolves `process` (including readable-stream's `process/` import) to a small shim that drains via `queueMicrotask`, then `setImmediate`, then `setTimeout`. Node keeps using the real `process.nextTick`.

Review follow-up: removed the global nextTick mutation and made the shared fallback use microtasks when available. Eight focused tests, the TypeScript/browser build and targeted ESLint pass.

AI assistance: Cursor Agent assisted with the original code; Codex assisted with attribution cleanup, review fixes and tests.

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -2,6 +2,7 @@ const { build } = require('esbuild')
 const { polyfillNode } = require('esbuild-plugin-polyfill-node');
 const { rimraf } = require('rimraf')
 const fs = require('fs')
+const path = require('path')
 const { version } = require('./package.json');
 
 const outdir = 'dist'
@@ -18,6 +19,21 @@ const options = {
     globalName: 'mqtt',
     sourcemap: false, // this can be enabled while debugging, if we decide to keep this enabled we should also ship the `src` folder to npm
     plugins: [
+        {
+            name: 'browser-process',
+            setup(build) {
+                // mqtt-packet corks the stream and uncorks on process.nextTick.
+                // The default browser process shim drains that queue with
+                // setTimeout(0), which React Native throttles while
+                // backgrounded, so writes stall. Resolve process to a
+                // queueMicrotask-based shim instead (issue #2053).
+                const shim = path.resolve(__dirname, 'src/lib/browser-process.ts')
+                // readable-stream requires 'process/' (trailing slash).
+                build.onResolve({ filter: /^(node:)?process\/?$/ }, () => ({
+                    path: shim,
+                }))
+            }
+        },
         polyfillNode({
             polyfills: [
                 'readable-stream'
--- a/esbuild.js
+++ b/esbuild.js
@@ -18,6 +18,7 @@ const options = {
     platform: 'browser',
     globalName: 'mqtt',
     sourcemap: false, // this can be enabled while debugging, if we decide to keep this enabled we should also ship the `src` folder to npm
+    metafile: true,
     plugins: [
         {
             name: 'browser-process',
@@ -27,7 +28,7 @@ const options = {
                 // setTimeout(0), which React Native throttles while
                 // backgrounded, so writes stall. Resolve process to a
                 // queueMicrotask-based shim instead (issue #2053).
-                const shim = path.resolve(__dirname, 'src/lib/browser-process.ts')
+                const shim = path.resolve(__dirname, 'scripts/browser-process.ts')
                 // readable-stream requires 'process/' (trailing slash).
                 build.onResolve({ filter: /^(node:)?process\/?$/ }, () => ({
                     path: shim,
@@ -35,9 +36,8 @@ const options = {
             }
         },
         polyfillNode({
-            polyfills: [
-                'readable-stream'
-            ]
+            // Disable the default process polyfill so plugin order cannot replace our shim.
+            polyfills: { process: false, 'readable-stream': true }
         }),
         {
             name: 'resolve-package-json',
@@ -86,17 +86,17 @@ const options = {
 async function run() {
     const start = Date.now()
     await rimraf(outdir)
-    await build(options)
+    await buildWithProcessCheck(options)
 
     options.minify = true
     options.outfile = `${outdir}/mqtt.min.js`
-    await build(options)
+    await buildWithProcessCheck(options)
 
 
     options.outfile = `${outdir}/mqtt.esm.js`
     options.format = 'esm'
 
-    await build(options)
+    await buildWithProcessCheck(options)
 
     console.log(`Build time: ${Date.now() - start}ms`)
     console.log('Build output:')
@@ -109,7 +109,19 @@ async function run() {
     }
 }
 
-run().catch((e) => {
+async function buildWithProcessCheck(options) {
+    const result = await build(options)
+    const inputs = Object.keys(result.metafile.inputs)
+    if (!inputs.some(input => input.endsWith('scripts/browser-process.ts')) ||
+        inputs.some(input => /(?:process\/browser|node\/process)\.(?:js|mjs)$/.test(input))) {
+        throw new Error('Browser bundle did not resolve process exclusively to the microtask shim')
+    }
+    return result
+}
+
+if (require.main === module) run().catch((e) => {
     console.error(e)
     process.exit(1)
 })
+
+module.exports = { options, buildWithProcessCheck }
--- a/scripts/browser-process.ts
+++ b/scripts/browser-process.ts
@@ -1,33 +1,43 @@
+// Build-only process shim: microtasks keep MQTT writes moving when React Native
+// throttles background timers. Browser microtasks share the Promise job queue;
+// they cannot reproduce Node's separate nextTick queue ordering.
 const pending: Array<{
-	fun: (...args: any[]) => void
+	callback: (...args: any[]) => void
 	args: any[]
 }> = []
 let draining = false
 
 function runNextTicks() {
 	draining = true
-	while (pending.length) {
-		const item = pending.shift()
-		item.fun(...item.args)
+	try {
+		while (pending.length) {
+			const item = pending.shift()
+			item.callback(...item.args)
+		}
+	} finally {
+		draining = false
+		// A throwing user callback must not strand the remaining queue.
+		if (pending.length) scheduleDrain()
 	}
-	draining = false
 }
 
 function scheduleDrain() {
 	if (typeof queueMicrotask === 'function') {
 		queueMicrotask(runNextTicks)
-	} else if (typeof globalThis.setImmediate === 'function') {
-		globalThis.setImmediate(runNextTicks)
 	} else {
 		setTimeout(runNextTicks, 0)
 	}
 }
 
-export function nextTick(fun: (...args: any[]) => void, ...args: any[]): void {
-	if (typeof fun !== 'function') {
+export function nextTick(
+	callback: (...args: any[]) => void,
+	...args: any[]
+): void {
+	if (typeof callback !== 'function') {
 		throw new TypeError('"callback" argument must be a function')
 	}
-	pending.push({ fun, args })
+	pending.push({ callback, args })
+	// Schedule once for a new queue; nested ticks are consumed by the active drain.
 	if (pending.length === 1 && !draining) {
 		scheduleDrain()
 	}
@@ -39,6 +49,7 @@ export const title = 'browser'
 export const browser = true
 export const env: Record<string, string | undefined> = {}
 export const argv: string[] = []
+// An empty version deliberately selects the compatible process-nextick-args wrapper.
 export const version = ''
 export const versions: Record<string, string | undefined> = {}
 export const on = noop
--- a/scripts/test-browser-process-build.js
+++ b/scripts/test-browser-process-build.js
@@ -1,0 +1,24 @@
+const assert = require('node:assert/strict')
+const { runInNewContext } = require('node:vm')
+const { options, buildWithProcessCheck } = require('../esbuild.js')
+
+async function test() {
+    const result = await buildWithProcessCheck({
+        ...options, entryPoints: undefined, outfile: undefined, write: false,
+        globalName: 'probe',
+        stdin: {
+            contents: 'import {nextTick as a} from "process"; import {nextTick as b} from "process/"; export {a,b}',
+            resolveDir: process.cwd(),
+        },
+    })
+    const tasks = []
+    const context = { queueMicrotask: callback => tasks.push(callback) }
+    runInNewContext(result.outputFiles[0].text, context)
+    const seen = []
+    context.probe.a(() => seen.push(1))
+    context.probe.b(() => seen.push(2))
+    assert.equal(tasks.length, 1)
+    tasks.shift()()
+    assert.deepEqual(seen, [1, 2])
+}
+test().catch(error => { console.error(error); process.exitCode = 1 })
--- a/scripts/test-browser-process-build.js
+++ b/scripts/test-browser-process-build.js
@@ -1,5 +1,6 @@
 const assert = require('node:assert/strict')
 const { runInNewContext } = require('node:vm')
+const { build } = require('esbuild')
 const { options, buildWithProcessCheck } = require('../esbuild.js')
 
 async function test() {
@@ -20,5 +21,28 @@ async function test() {
     assert.equal(tasks.length, 1)
     tasks.shift()()
     assert.deepEqual(seen, [1, 2])
+    const shared = await build({
+        entryPoints: ['src/lib/shared.ts'], bundle: true, write: false,
+        platform: 'browser', format: 'iife', globalName: 'shared',
+    })
+    for (const microtasksAvailable of [true, false]) {
+        const jobs = []
+        const timers = []
+        const fallbackContext = {
+            process: {},
+            setTimeout: callback => timers.push(callback),
+            ...(microtasksAvailable ? { queueMicrotask: callback => jobs.push(callback) } : {}),
+        }
+        runInNewContext(shared.outputFiles[0].text, fallbackContext)
+        let ran = false
+        fallbackContext.shared.nextTick(() => { ran = true })
+        assert.equal(ran, false)
+        assert.equal(jobs.length, microtasksAvailable ? 1 : 0)
+        assert.equal(timers.length, microtasksAvailable ? 0 : 1)
+        ;(microtasksAvailable ? jobs : timers).shift()()
+        assert.equal(ran, true)
+        assert.equal(fallbackContext.process.nextTick, undefined)
+    }
+
 }
 test().catch(error => { console.error(error); process.exitCode = 1 })
--- a/src/lib/browser-process.ts
+++ b/src/lib/browser-process.ts
@@ -1,0 +1,93 @@
+const pending: Array<{
+	fun: (...args: any[]) => void
+	args: any[]
+}> = []
+let draining = false
+
+function runNextTicks() {
+	draining = true
+	while (pending.length) {
+		const item = pending.shift()
+		item.fun(...item.args)
+	}
+	draining = false
+}
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
+		queueMicrotask(runNextTicks)
+	} else if (typeof globalThis.setImmediate === 'function') {
+		globalThis.setImmediate(runNextTicks)
+	} else {
+		setTimeout(runNextTicks, 0)
+	}
+}
+
+export function nextTick(fun: (...args: any[]) => void, ...args: any[]): void {
+	if (typeof fun !== 'function') {
+		throw new TypeError('"callback" argument must be a function')
+	}
+	pending.push({ fun, args })
+	if (pending.length === 1 && !draining) {
+		scheduleDrain()
+	}
+}
+
+const noop = () => undefined
+
+export const title = 'browser'
+export const browser = true
+export const env: Record<string, string | undefined> = {}
+export const argv: string[] = []
+export const version = ''
+export const versions: Record<string, string | undefined> = {}
+export const on = noop
+export const addListener = noop
+export const once = noop
+export const off = noop
+export const removeListener = noop
+export const removeAllListeners = noop
+export const emit = noop
+export const prependListener = noop
+export const prependOnceListener = noop
+export function listeners() {
+	return []
+}
+export function binding() {
+	throw new Error('process.binding is not supported')
+}
+export function cwd() {
+	return '/'
+}
+export function chdir() {
+	throw new Error('process.chdir is not supported')
+}
+export function umask() {
+	return 0
+}
+
+const browserProcess = {
+	title,
+	browser,
+	env,
+	argv,
+	version,
+	versions,
+	nextTick,
+	on,
+	addListener,
+	once,
+	off,
+	removeListener,
+	removeAllListeners,
+	emit,
+	prependListener,
+	prependOnceListener,
+	listeners,
+	binding,
+	cwd,
+	chdir,
+	umask,
+}
+
+export default browserProcess
--- a/src/lib/connect/index.ts
+++ b/src/lib/connect/index.ts
@@ -9,11 +9,6 @@ import MqttClient, {
 import isBrowser from '../is-browser'
 import { type StreamBuilder } from '../shared'
 
-// Handling the process.nextTick is not a function error in react-native applications.
-if (typeof process?.nextTick !== 'function') {
-	process.nextTick = setImmediate
-}
-
 const debug = _debug('mqttjs')
 
 let protocols: Record<string, StreamBuilder> = null
--- a/src/lib/shared.ts
+++ b/src/lib/shared.ts
@@ -94,7 +94,11 @@ export const nextTick =
 	typeof process?.nextTick === 'function'
 		? process.nextTick
 		: (callback: () => void) => {
-				setTimeout(callback, 0)
+				if (typeof queueMicrotask === 'function') {
+					queueMicrotask(callback)
+				} else {
+					setTimeout(callback, 0)
+				}
 			}
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
--- a/test/node/browser-process.ts
+++ b/test/node/browser-process.ts
@@ -8,25 +8,15 @@ import { nextTick } from '../../scripts/browser-process'
 describe('browser process.nextTick', () => {
 	afterEach(() => sinon.restore())
 	it('runs the callback as a microtask, not via setTimeout(0)', async () => {
-		const origSetTimeout = setTimeout
-		let timeoutUsed = false
-		globalThis.setTimeout = ((...args: Parameters<typeof setTimeout>) => {
-			timeoutUsed = true
-			return origSetTimeout(...args)
-		}) as typeof setTimeout
-
-		try {
-			let ran = false
-			nextTick(() => {
-				ran = true
-			})
-			assert.equal(ran, false, 'must not run synchronously')
-			await Promise.resolve()
-			assert.equal(ran, true, 'must run after a microtask')
-			assert.equal(timeoutUsed, false, 'must not schedule setTimeout')
-		} finally {
-			globalThis.setTimeout = origSetTimeout
-		}
+		const timer = sinon.spy(globalThis, 'setTimeout')
+		let ran = false
+		nextTick(() => {
+			ran = true
+		})
+		assert.equal(ran, false, 'must not run synchronously')
+		await Promise.resolve()
+		assert.equal(ran, true, 'must run after a microtask')
+		assert.isFalse(timer.called, 'must not schedule setTimeout')
 	})
 
 	it('forwards extra arguments', async () => {
@@ -65,33 +55,23 @@ describe('browser process.nextTick', () => {
 	})
 
 	it('uncorks a stream without waiting for setTimeout', async () => {
-		const origSetTimeout = setTimeout
-		let timeoutUsed = false
-		globalThis.setTimeout = ((...args: Parameters<typeof setTimeout>) => {
-			timeoutUsed = true
-			return origSetTimeout(...args)
-		}) as typeof setTimeout
+		const timer = sinon.spy(globalThis, 'setTimeout')
+		const stream = new PassThrough()
+		const chunks: Buffer[] = []
+		stream.on('data', (chunk) => {
+			chunks.push(chunk)
+		})
+		stream.cork()
+		stream.write('ping')
+		assert.equal(chunks.length, 0, 'corked writes stay buffered')
 
-		try {
-			const stream = new PassThrough()
-			const chunks: Buffer[] = []
-			stream.on('data', (chunk) => {
-				chunks.push(chunk)
-			})
-			stream.cork()
-			stream.write('ping')
-			assert.equal(chunks.length, 0, 'corked writes stay buffered')
+		nextTick(() => {
+			stream.uncork()
+		})
+		await Promise.resolve()
 
-			nextTick(() => {
-				stream.uncork()
-			})
-			await Promise.resolve()
-
-			assert.equal(Buffer.concat(chunks).toString(), 'ping')
-			assert.equal(timeoutUsed, false)
-		} finally {
-			globalThis.setTimeout = origSetTimeout
-		}
+		assert.equal(Buffer.concat(chunks).toString(), 'ping')
+		assert.isFalse(timer.called)
 	})
 
 	it('falls back to a timer when microtasks are unavailable', () => {
--- a/test/node/browser-process.ts
+++ b/test/node/browser-process.ts
@@ -1,10 +1,12 @@
-import { describe, it } from 'node:test'
+import { describe, it, afterEach } from 'node:test'
 import { PassThrough } from 'node:stream'
+import { execFileSync } from 'node:child_process'
 import { assert } from 'chai'
-import { nextTick as nodeNextTick } from '../../src/lib/shared'
-import { nextTick } from '../../src/lib/browser-process'
+import sinon from 'sinon'
+import { nextTick } from '../../scripts/browser-process'
 
 describe('browser process.nextTick', () => {
+	afterEach(() => sinon.restore())
 	it('runs the callback as a microtask, not via setTimeout(0)', async () => {
 		const origSetTimeout = setTimeout
 		let timeoutUsed = false
@@ -91,10 +93,47 @@ describe('browser process.nextTick', () => {
 			globalThis.setTimeout = origSetTimeout
 		}
 	})
-})
 
-describe('Node process.nextTick', () => {
-	it('is what MQTT.js uses outside the browser bundle', () => {
-		assert.equal(nodeNextTick, process.nextTick)
+	it('falls back to a timer when microtasks are unavailable', () => {
+		sinon.stub(globalThis, 'queueMicrotask').value(undefined)
+		const timer = sinon.stub(globalThis, 'setTimeout')
+		let ran = false
+		nextTick(() => {
+			ran = true
+		})
+		assert.isFalse(ran)
+		assert.equal(timer.callCount, 1)
+		;(timer.firstCall.args[0] as () => void)()
+		assert.isTrue(ran)
+	})
+
+	it('resumes remaining and future callbacks after a throwing callback', () => {
+		const tasks: Array<() => void> = []
+		sinon.stub(globalThis, 'queueMicrotask').callsFake((callback) => {
+			tasks.push(callback)
+		})
+		const seen: string[] = []
+		nextTick(() => {
+			throw new Error('boom')
+		})
+		nextTick(() => {
+			seen.push('queued')
+		})
+		assert.throws(tasks.shift(), 'boom')
+		assert.lengthOf(tasks, 1)
+		tasks.shift()()
+		nextTick(() => {
+			seen.push('future')
+		})
+		tasks.shift()()
+		assert.deepEqual(seen, ['queued', 'future'])
+	})
+
+	it('resolves process and process/ to the microtask shim through the real build plugins', async () => {
+		// Run without esbuild-register's tsconfig baseUrl hook, which shadows
+		// the esbuild package with the repository's esbuild.js file.
+		execFileSync(process.execPath, [
+			'scripts/test-browser-process-build.js',
+		])
 	})
 })
--- a/test/node/browser-process.ts
+++ b/test/node/browser-process.ts
@@ -1,0 +1,100 @@
+import { describe, it } from 'node:test'
+import { PassThrough } from 'node:stream'
+import { assert } from 'chai'
+import { nextTick as nodeNextTick } from '../../src/lib/shared'
+import { nextTick } from '../../src/lib/browser-process'
+
+describe('browser process.nextTick', () => {
+	it('runs the callback as a microtask, not via setTimeout(0)', async () => {
+		const origSetTimeout = setTimeout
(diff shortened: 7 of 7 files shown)


robertsLando (commented):
## Deep review — `fix: schedule browser process.nextTick with queueMicrotask`

Thanks for this, and sorry for the duplicate effort with #2054 — yours is the one we kept, because it also resolves readable-stream's `process/` (trailing slash) specifier, which is where the cork/uncork actually happens.

**Verdict: needs work** — one blocker, otherwise a correct and well-targeted fix.

I reviewed against a worktree at the PR head with a full `npm run build`, so the notes below are build-verified rather than read off the diff.

**What's good**
- Root cause is right, and the fix lands in the artifact that matters (`exports["."]["react-native"]` → `dist/mqtt.esm.js`).
- The redirect genuinely wins: in `dist/mqtt.js`, `polyfills/process.js` compiles to `init_browser_process()`, and all 7 readable-stream `require('process/')` sites bind the shim.
- Bundle **shrinks**: `mqtt.js` 766.33 → 753.89 KB, `mqtt.esm.js` 329.84 → 324.82 KB.
- It ships tests, and all 6 pass.

**Top 3 risks**
1. A throwing callback permanently bricks the tick queue — a regression versus the shim being replaced.
2. The fix depends on undocumented esbuild plugin ordering; it reverts silently with a green build.
3. A competing, now-dead `nextTick` patch is left in place.

**Theme:** every blocker/major traces to one gap — the shim's *failure modes* are neither handled nor asserted: what happens when a callback throws, and what happens when the build stops resolving to the shim.

---

### Findings outside the diff

**[Major] `src/lib/connect/index.ts:12-14`** — the pre-existing RN patch

```js
if (typeof process?.nextTick !== 'function') {
  process.nextTick = setImmediate
}
```

is now dead: the shim always defines `nextTick`, so the guard never passes. It compiles to `browser_process_exports.nextTick = setImmediate` against a getter-only namespace object, which would throw `TypeError` in strict mode if the shim ever stopped exporting `nextTick`. Two competing fixes for one problem, with nothing cross-referencing them. Worth deleting as part of this PR.

**[Minor] `src/lib/shared.ts:93`** — the runtime's own `nextTick` fallback is still `setTimeout(callback, 0)`, the exact pattern this PR declares broken. Unreachable in the bundle, but the two siblings now disagree on what "nextTick" means.

### FYI

- `esbuild.js`'s existing `polyfills: ['readable-stream']` is an **array**, but the plugin reads it as an object map (`dist/index.cjs:74`: `polyfills[moduleName] ?? true`). That entry is already a silent no-op. Pre-existing, not introduced here — but it's why the object form is the right shape for the suggestion on `esbuild.js`.
- The filter `/^(node:)?process\/?$/` does **not** over-match: `process-nextick-args` and `process/browser` are excluded by the `$` anchor. Good call on the trailing slash.
- The old jspm `setTimeout`-based process chunk is still bundled but now unreferenced — dead weight, no correctness impact.
- A throw inside `queueMicrotask` is *not* swallowed (Node reports it as `uncaughtException`; the HTML spec requires hosts to report it). Error visibility is not degraded — the permanent queue brick is the real problem.

### Coverage

Ran: correctness, DRY/codebase-fit, design/API, tests, operability, readability. Not run: security and performance (no relevant surface in the diff).

src/lib/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+}> = []
+let draining = false
+
+function runNextTicks() {
+	draining = true
+	while (pending.length) {
+		const item = pending.shift()
+		item.fun(...item.args)
+	}
+	draining = false
```
robertsLando: **[Blocker] · Correctness**

`runNextTicks` has no `try/finally`, so one throwing callback leaves `draining = true` forever. The `pending.length === 1 && !draining` guard in `nextTick` then never schedules another drain, and the queue is permanently dead.

**Why:** reproduced against this branch — after a throw, no later `nextTick` ever runs:

```
caught: boom
SECOND nextTick ran after a throw? false
```

All 51 `nextTick` sites in the built bundle go dead, so the client permanently stops writing, reconnecting, and firing callbacks. The trigger is ordinary: `dist/mqtt.js` contains `process2.nextTick(() => callback.apply(stream, args))`, so a user `publish`/`subscribe` callback that throws bricks the client.

This is a **regression**, not an inherited flaw. The jspm shim being replaced arms `setTimeout(cleanUpNextTick, 0)` *before* draining and self-heals; Node's real `nextTick` also keeps draining after an `uncaughtException`.

**Fix:** always clear `draining`, and re-schedule a drain if `pending` is non-empty when the loop exits abnormally — either replicate the old watchdog, or drain one item per iteration inside a `try/finally`.

scripts/browser-process.ts
```diff
@@ -0,0 +1,93 @@
+const pending: Array<{
```
robertsLando, line 4: **[Major] · Design/API**

This is build-only tooling living in published library source, which makes it an accidental public API.

**Why:** verified — `tsconfig.build.json` excludes only `test`/`dist`/`build`/`example.ts`, so `tsc -p tsconfig.build.json` emits `build/lib/browser-process.{js,d.ts,js.map}`. `files` ships `build/`, and `exports` maps `"./*" → "./build/*.js"`. So `import 'mqtt/lib/browser-process'` becomes a resolvable, typed entry point with semver obligations — for a file nothing in the library imports. It also makes `esbuild.js` the only place reading raw `.ts` from `src/`, while every other build input is `build/*.js`.

**Fix:** move the shim outside `src/` (a build-scripts location), or exclude it from `tsconfig.build.json`.

scripts/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+	while (pending.length) {
+		const item = pending.shift()
+		item.fun(...item.args)
+	}
+	draining = false
+}
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
+		queueMicrotask(runNextTicks)
```
robertsLando, line 26: **[Minor] · Correctness**

Draining as a microtask inverts Node's ordering: a `nextTick` queued from inside a `.then()` now runs *after* other pending promise jobs, whereas Node drains the tick queue first.

**Why:** low risk for the cork/uncork path this targets, but it is a silent semantic divergence for readable-stream's `emitErrorNT`/`emitCloseNT` sequencing.

**Fix:** worth a line in a file header comment so the next reader knows it's deliberate.

src/lib/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
+		queueMicrotask(runNextTicks)
+	} else if (typeof globalThis.setImmediate === 'function') {
+		globalThis.setImmediate(runNextTicks)
```
robertsLando: **[Minor] · Correctness**

The `setImmediate` tier is unreachable.

**Why:** `queueMicrotask` is checked first and exists in every engine that also has `setImmediate` here (Node, RN/Hermes, jsdom). Only the `setTimeout` leg can ever fire, and nothing tests either.

**Fix:** drop the middle tier, or add a line saying which engine it's for.

src/lib/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+	} else {
+		setTimeout(runNextTicks, 0)
+	}
+}
+
+export function nextTick(fun: (...args: any[]) => void, ...args: any[]): void {
```
robertsLando: **[Nit] · Readability**

`fun` reads as carried over from `process/browser.js`. `callback` would match the surrounding TS and the `TypeError` message just below it. Optional.

scripts/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+		setTimeout(runNextTicks, 0)
+	}
+}
+
+export function nextTick(fun: (...args: any[]) => void, ...args: any[]): void {
+	if (typeof fun !== 'function') {
+		throw new TypeError('"callback" argument must be a function')
+	}
+	pending.push({ fun, args })
+	if (pending.length === 1 && !draining) {
```
robertsLando, line 41: **[Minor] · Readability**

This guard carries the whole scheduling invariant (schedule exactly once per drain cycle, and never re-schedule while draining) but is unexplained — and the file has no comments at all, so its reason to exist lives only in `esbuild.js`, far away.

**Fix:** a one-line WHY here, plus a short file header on why `queueMicrotask` rather than `setTimeout`.

scripts/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+		scheduleDrain()
+	}
+}
+
+const noop = () => undefined
+
+export const title = 'browser'
```
robertsLando, line 48: **[Minor] · Design/API**

The shim is narrower than the jspm polyfill it displaces: `emitWarning`, `exit`, `stdout`, `stderr`, `platform`, `arch`, `hrtime` and `pid` are gone. Since polyfillNode injects the global as `export * as process from "process"`, this namespace *is* `process` for every bundled dep.

**Why:** not a live bug — I grepped the built bundle and no dep reads them; the two `process.stdout`/`stderr` reads are `dest !== …` identity guards that behave identically against `undefined`. This is forward-looking only: a future dep's unguarded `process.platform` becomes a `TypeError` instead of degrading.

**Fix:** optional — carry over the inert members. `binding`/`chdir` throwing is correct and matches both jspm and `process/browser.js`.

scripts/browser-process.ts
```diff
@@ -0,0 +1,93 @@
...
+export const browser = true
+export const env: Record<string, string | undefined> = {}
+export const argv: string[] = []
+export const version = ''
```
robertsLando, line 53: **[Minor] · Design/API**

`version = ''` is falsy, so `process-nextick-args` takes its wrapper branch instead of `module.exports = process`.

**Why:** harmless — I traced it, the wrapper still calls this shim's `nextTick` — but it's an unintended path, and jspm's shim sets a plausible `'v16.8.0'` here.

**Fix:** a comment, or a plausible version string.

esbuild.js
```diff
@@ -18,6 +19,21 @@ const options = {
...
+                // mqtt-packet corks the stream and uncorks on process.nextTick.
+                // The default browser process shim drains that queue with
+                // setTimeout(0), which React Native throttles while
+                // backgrounded, so writes stall. Resolve process to a
+                // queueMicrotask-based shim instead (issue #2053).
+                const shim = path.resolve(__dirname, 'src/lib/browser-process.ts')
+                // readable-stream requires 'process/' (trailing slash).
+                build.onResolve({ filter: /^(node:)?process\/?$/ }, () => ({
+                    path: shim,
+                }))
```
robertsLando, line 35: **[Major] · Design/API + Operability**

The redirect works only because this plugin is registered *before* `polyfillNode` in the array, and nothing states or enforces that.

**Why:** a reorder, or a polyfillNode upgrade that changes registration semantics, silently reverts the RN fix — with a green build *and* green tests, since the added tests import the shim directly and never assert what the bundle resolves `process` to. That's the same class of silent stall this PR is fixing.

**Fix:** `polyfillNode` has a first-class opt-out. In `esbuild-plugin-polyfill-node/dist/index.cjs:74-76`:

```js
const polyfill = polyfills[moduleName] ?? true;
if (polyfill === false) {
  return;
}
```

So passing `polyfills: { process: false, 'readable-stream': true }` (object form) makes this redirect order-independent. Note the current `polyfills: ['readable-stream']` is an array where the plugin reads an object map, so that entry is already a silent no-op — switching to the object form fixes both.

A build-time assertion (fail the build if the resolved path isn't the local shim) would also catch drift.

test/node/browser-process.ts
```diff
@@ -0,0 +1,100 @@
+import { describe, it } from 'node:test'
+import { PassThrough } from 'node:stream'
+import { assert } from 'chai'
+import { nextTick as nodeNextTick } from '../../src/lib/shared'
+import { nextTick } from '../../src/lib/browser-process'
```
robertsLando: **[Major] · Tests**

Nothing tests the esbuild plugin — i.e. that `process` and `process/` actually resolve to the shim in a real build.

**Why:** that resolution *is* the fix. These tests import the source directly, so they would stay green through a complete revert of the plugin, and the regression would only show up as an RN-only stall in production.

**Fix:** assert against the built output (e.g. a marker unique to the shim in `dist/mqtt.js`, or an esbuild metafile check), or run this suite in `test/browser/` against the bundle.

test/node/browser-process.ts
```diff
@@ -0,0 +1,100 @@
...
+
+		try {
+			let ran = false
+			nextTick(() => {
+				ran = true
+			})
+			assert.equal(ran, false, 'must not run synchronously')
+			await Promise.resolve()
+			assert.equal(ran, true, 'must run after a microtask')
+			assert.equal(timeoutUsed, false, 'must not schedule setTimeout')
```
robertsLando: **[Major] · Tests**

`assert.equal(timeoutUsed, false)` passes trivially: Node has `queueMicrotask`, so the first branch always wins and this can never fail. Same at line 89.

**Why:** the assertion looks like it guards the fallback ladder but exercises nothing — a regression that fell through to `setTimeout` whenever `queueMicrotask` is unavailable would still pass.

**Fix:** stub out `queueMicrotask` to force the lower tiers, and assert there.

Also missing here: a callback that **throws**, and whether a later `nextTick` still runs afterwards. That's the blocker on `src/lib/browser-process.ts` and it's untested.

test/node/browser-process.ts
```diff
@@ -0,0 +1,100 @@
...
+		assert.deepEqual(order, [1, 2])
+	})
+
+	it('uncorks a stream without waiting for setTimeout', async () => {
+		const origSetTimeout = setTimeout
+		let timeoutUsed = false
+		globalThis.setTimeout = ((...args: Parameters<typeof setTimeout>) => {
+			timeoutUsed = true
+			return origSetTimeout(...args)
+		}) as typeof setTimeout
```
robertsLando: **[Minor] · DRY**

This `globalThis.setTimeout` save/stub/restore block is duplicated verbatim from lines 9-14.

**Why:** the repo already uses sinon for timer stubbing — see `test/node/keepaliveManager.ts:3,21` and the shared `fakeTimersOptions` in `test/node/abstract_client.ts:52-54`.

**Fix:** use sinon, or lift one local helper.

Separately: these tests share the module-level `pending`/`draining` singletons, so they're order-dependent. Safe today only because the `TypeError` test throws before pushing.

test/node/browser-process.ts
```diff
@@ -0,0 +1,100 @@
...
+			assert.equal(timeoutUsed, false)
+		} finally {
+			globalThis.setTimeout = origSetTimeout
+		}
+	})
+})
+
+describe('Node process.nextTick', () => {
+	it('is what MQTT.js uses outside the browser bundle', () => {
+		assert.equal(nodeNextTick, process.nextTick)
```
robertsLando: **[Minor] · Tests**

`assert.equal(nodeNextTick, process.nextTick)` restates the definition in `shared.ts` rather than testing behaviour — it fails only if that line is edited, which is what the diff would show anyway.

**Fix:** drop it, or assert something behavioural about the Node path.

src/lib/browser-process.ts
sankalpsthakur: Fixed in 10b3bae: finally clears draining and schedules remaining work. The regression covers both already-queued and future callbacks after a throw.

scripts/browser-process.ts
sankalpsthakur: Moved the shim to scripts/browser-process.ts. The TypeScript library build no longer publishes it (10b3bae).

scripts/browser-process.ts
sankalpsthakur: Documented the shared Promise/microtask queue and the difference from Node's separate nextTick queue (10b3bae).

src/lib/browser-process.ts
sankalpsthakur: Removed setImmediate. Added a test that disables queueMicrotask and exercises the timer fallback (10b3bae).

src/lib/browser-process.ts
sankalpsthakur: Renamed fun to callback (10b3bae).

scripts/browser-process.ts
sankalpsthakur: Added the scheduling invariant and a header explaining the background-timer motivation (10b3bae).

scripts/browser-process.ts
sankalpsthakur: Keeping the current minimal surface for now, since the bundled dependencies do not require those optional members. The build now verifies which process shim is included.

scripts/browser-process.ts
sankalpsthakur: Added a comment explaining the empty version and compatible process-nextick-args wrapper path (10b3bae).

esbuild.js
sankalpsthakur: Set polyfills to an object with process:false and added a metafile assertion to every browser build. Plugin order no longer selects the default process shim (10b3bae).

test/node/browser-process.ts
sankalpsthakur: Added a real esbuild probe for both process and process/, executing the built result in a VM. All three MQTT browser bundles also pass the metafile assertion (10b3bae).

test/node/browser-process.ts
sankalpsthakur: Added forced timer-fallback and throwing-callback recovery tests. All eight focused tests pass (10b3bae).

test/node/browser-process.ts
sankalpsthakur: Replaced both save/stub/restore blocks with Sinon spies and shared afterEach cleanup (48c5cb2). Each queue test drains its work, including the error-recovery test.

test/node/browser-process.ts
sankalpsthakur: Removed the definition-restating Node assertion (10b3bae).

sankalpsthakur:
@robertsLando Removed the global nextTick patch and fixed the shared microtask fallback in 6ccb529. Queue recovery after a thrown callback and build-plugin selection are covered. Eight focused tests, the TypeScript/browser build and targeted ESLint pass.

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

robertsLando (commented):
_(tooling artifact — disregard; superseded by the full review below)_

esbuild.js
```diff
@@ -93,7 +109,19 @@ async function run() {
     }
 }
 
-run().catch((e) => {
+async function buildWithProcessCheck(options) {
+    const result = await build(options)
+    const inputs = Object.keys(result.metafile.inputs)
+    if (!inputs.some(input => input.endsWith('scripts/browser-process.ts')) ||
+        inputs.some(input => /(?:process\/browser|node\/process)\.(?:js|mjs)$/.test(input))) {
```
robertsLando: **[Blocker] · Correctness / Security**

This guard does not detect the regression it exists to prevent. Two independent reproductions:

1. Put `polyfillNode` first so it wins bare `process`, and the build **passes**. `scripts/browser-process.ts` is still an input (it keeps `process/`, which polyfillNode's filter lacks a trailing-slash form for), so clause 1 is satisfied — but the global `process` becomes jspm's polyfill and its `setTimeout`-based `drainQueue` lands in the bundle. That is mqtt-packet's uncork back on setTimeout, i.e. the exact #2053 bug, with a green build.
2. Running the real build today: `@jspm/core/nodelibs/browser/chunk-DEMDiNwt.js` — a complete jspm `process` whose `nextTick` is `setTimeout(drainQueue, 0)`, pulled in transitively via the jspm `url`/`path` polyfills — is a metafile input and is present in `dist/mqtt.js`, `dist/mqtt.min.js` and `dist/mqtt.esm.js` (`grep -c 'is not supported by JSPM core'` → 2 in each).

Clause 2's regex never fires on either. The real input path is `@jspm/core/nodelibs/browser/process.js`; the regex also misses `esbuild-plugin-polyfill-node/polyfills/process.js`. It only appears to work today because readable-stream's `require('process/')` happens to drag in `process/browser.js` when the shim plugin is absent — drop that coincidence and a regression to the old scheduler ships silently, under an error message claiming the bundle "resolved process exclusively to the microtask shim".

**Fix:** invert it. Allowlist which inputs may provide `nextTick` — only `scripts/browser-process.ts` — and fail on any other module exporting one, rather than denylisting two hardcoded filenames. Asserting that the bundle contains no second `nextTick` implementation would also work. Either way the injected-global path (`polyfills/process.js` → shim) needs to be part of the assertion. `scripts/test-browser-process-build.js` currently locks the weak version in.

Minor, and moot if the above is done: the regex is unnamed and unexplained — a reader has to reverse-engineer what `process/browser` and `node/process` represent.

src/lib/connect/index.ts
```diff
@@ -9,11 +9,6 @@ import MqttClient, {
 import isBrowser from '../is-browser'
 import { type StreamBuilder } from '../shared'
 
-// Handling the process.nextTick is not a function error in react-native applications.
```
robertsLando, line 12: **[Blocker] · Design / Correctness**

Removing this guard is compensated only inside MQTT.js's own esbuild bundle. Checked the resolution on this head:

```
exports["."] = { "react-native": "./dist/mqtt.esm.js", "browser": {...}, "default": "./build/index.js" }
main: "./build/index.js"
// no top-level "react-native" field
```

Metro honours `exports` only with `unstable_enablePackageExports` on (RN ≥ 0.79). Older Metro, webpack 4, jest-resolve, and `mqtt/lib/*` subpath imports all land on `build/index.js`.

**Why:** there, mqtt-packet → `process-nextick-args` reads the **bare global** `process` (`!process.version` → wrapper branch → `process.nextTick(fn)`) — which is exactly the "process.nextTick is not a function" TypeError this line was added to fix. So those consumers get a *new* TypeError where there was none, **and** still don't get the microtask fix for #2053. `shared.ts`'s `nextTick` covers only MQTT.js's own call sites, never deps.

**Fix:** make the fix reachable on the `main`/`browser` path — e.g. a top-level `react-native` field pointing at the dist bundle, and/or keep a non-mutating guard that is inert in the bundled build (the injected namespace object is read-only, which is presumably why the assignment had to go). If the intent is to require RN ≥ 0.79, that needs to be stated.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
...
+	callback: (...args: any[]) => void
+	args: any[]
+}> = []
+let draining = false
+
+function runNextTicks() {
+	draining = true
+	try {
+		while (pending.length) {
+			const item = pending.shift()
```
robertsLando: **[Major] · Performance**

`pending.shift()` inside `while (pending.length)` is O(n) per removal (V8 re-indexes the backing array), making a single drain O(n²) in queued callbacks.

**Why:** `writeToStream.js` in mqtt-packet calls `nextTick(uncork, stream)` once per packet write, so any tight synchronous publish loop — bulk publish, retained-message replay, high-rate QoS0 fire-and-forget — enqueues N callbacks before the first microtask checkpoint runs. Draining them then costs O(N²) instead of O(N). The polyfill this replaces (browserify `process/browser.js`) explicitly avoids exactly this with a `queueIndex` cursor plus swapping `queue`/`currentQueue` references — no `shift()`/`splice()` anywhere. So this is an algorithmic regression versus the code being removed, on the hot path the PR names.

**Fix:** drain with an index cursor over the current array, swapping in a fresh array for new pushes.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
...
+const pending: Array<{
+	callback: (...args: any[]) => void
+	args: any[]
+}> = []
+let draining = false
+
+function runNextTicks() {
+	draining = true
+	try {
+		while (pending.length) {
```
robertsLando: **[Minor] · Security / Robustness**

`while (pending.length)` drains the entire queue inside a single microtask checkpoint, so a callback that re-queues itself blocks the main thread with no rendering and no event handling — tab freeze. The previous `setTimeout(0)` shim yielded to the task queue each pass.

No reachable path in MQTT.js or readable-stream today, and it matches both the browserify polyfill being replaced and real Node `nextTick` semantics (recursive nextTick starves the loop there too), so this is hardening rather than a live bug. **Fix:** bound the work per drain and reschedule the remainder.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
+		queueMicrotask(runNextTicks)
+	} else {
+		setTimeout(runNextTicks, 0)
+	}
+}
+
+export function nextTick(
```
robertsLando: **[Major] · DRY / Design**

This PR now ships two independent `nextTick` implementations in the same browser bundle, and they are not connected.

`src/lib/shared.ts:93` still resolves the **global** `process`. With `polyfills: { process: false }`, polyfill-node no longer injects a `process` global — only bare `process`/`process/` *imports* hit the shim. In React Native the host global `process.nextTick` exists and is the timer-based one this PR exists to avoid, so MQTT.js's own ticks (`src/lib/client.ts:811`, `src/lib/client.ts:1571`, and the `setImmediate` fallback at `src/lib/client.ts:47-54`, which is built on `shared.nextTick`) keep the throttled behaviour, while only mqtt-packet and readable-stream get microtasks.

The two also differ semantically — batched FIFO drain with arg forwarding and throw recovery here, one `queueMicrotask` per callback there — so ordering between the queues can diverge as either side changes.

**Fix:** one scheduler. Either have `shared.ts` import this shim, or move the shim into `src/lib` and point the esbuild alias at that single module, instead of branching on the host global.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
+// Build-only process shim: microtasks keep MQTT writes moving when React Native
```
robertsLando: **[Major] · Design / Codebase fit**

This is shipped runtime code living in the build-tooling directory. `scripts/` is new in this PR and also holds a build test, so the repo's "shipped code lives in `src/`, `scripts/` is tooling" split is being broken as it's established.

**Why:** the file is outside `tsconfig.build.json`'s program *and* `tsconfig.json`'s `include` (`src`, `test`, `example.ts`), so `npm run build:ts` never type-checks it — it's only reached via a test import. It's also not in `package.json`'s `files`, so no consumer bundling from `build/` or `src/` can reach it.

**Fix:** put it under `src/lib/` and alias to it from `esbuild.js`.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
...
+	if (pending.length === 1 && !draining) {
+		scheduleDrain()
+	}
+}
+
+const noop = () => undefined
+
+export const title = 'browser'
+export const browser = true
+export const env: Record<string, string | undefined> = {}
```
robertsLando: **[Major] · DRY**

Roughly 85 of this file's 104 lines re-implement `process@0.11.10` — `title`, `browser`, `env`, `argv`, `versions`, `on`/`emit`/`listeners`, `cwd`, `chdir`, `umask`, `binding` — which is already in the lockfile as a `readable-stream@4.7.0` dependency. Only `nextTick` needed to change.

**Why:** a hand-copied fork of an upstream shim drifts silently. Today only `nextTick`, `stdout`, `stderr` and `env` are read in the bundle (and lru-cache guards `emitWarning`), so nothing regresses — but any dep bump reaching `hrtime`, `emitWarning`, `platform` or `arch` gets `undefined` instead of a value, and the new metafile guard checks resolution, not surface. `versions: {}` also interacts with the electron detection at `src/lib/is-browser.ts:16`.

**Fix:** re-export the installed `process/browser` polyfill and override `nextTick` only, keeping the delta to the one behaviour being fixed.

scripts/browser-process.ts
```diff
@@ -0,0 +1,104 @@
...
+		}
+	} finally {
+		draining = false
+		// A throwing user callback must not strand the remaining queue.
+		if (pending.length) scheduleDrain()
+	}
+}
+
+function scheduleDrain() {
+	if (typeof queueMicrotask === 'function') {
```
robertsLando: **[Major] · Operability**

`scheduleDrain()` silently falls back to `setTimeout(runNextTicks, 0)` when `queueMicrotask` isn't present at runtime — the exact React-Native-throttled behaviour this PR exists to fix — with no signal at all.

**Why:** the build-time guard verifies which *file* got bundled, not whether the target engine actually has `queueMicrotask`. If a device or engine lacks it, the original bug (stalled writes) returns and is indistinguishable from "fixed" from the outside, so it could ship in a later release and nobody would know until users report stalls again.

Related: a throwing tick callback propagates out of the microtask with no mqttjs-side trace. Node's `process.nextTick` at least emits `uncaughtException`, a hook consumers can attach to; this queue has no equivalent, and in RN production builds crash-reporting may swallow the raw exception entirely.

**Fix:** emit a one-time debug on first fallback use, and give the drain a `debug()` call so a stuck or interrupted drain is visible through `DEBUG=mqttjs*` — the library's documented diagnostic path. The same duplicated fallback exists at `src/lib/shared.ts:97-101`.

esbuild.js
```diff
@@ -17,11 +18,26 @@ const options = {
...
+            name: 'browser-process',
+            setup(build) {
+                // mqtt-packet corks the stream and uncorks on process.nextTick.
+                // The default browser process shim drains that queue with
+                // setTimeout(0), which React Native throttles while
+                // backgrounded, so writes stall. Resolve process to a
+                // queueMicrotask-based shim instead (issue #2053).
+                const shim = path.resolve(__dirname, 'scripts/browser-process.ts')
+                // readable-stream requires 'process/' (trailing slash).
+                build.onResolve({ filter: /^(node:)?process\/?$/ }, () => ({
```
robertsLando: **[Major] · Correctness**

The filter `/^(node:)?process\/?$/` misses `process/browser` and `process/browser.js`, which bypass the shim entirely. Notably clause 2 of `buildWithProcessCheck` tries to blocklist that same path, so these were expected to occur.

**Fix:** widen the filter to the `process` subpath family.

src/lib/shared.ts
```diff
@@ -116,7 +116,11 @@ export const nextTick =
 	typeof process?.nextTick === 'function'
 		? process.nextTick
 		: (callback: () => void) => {
-				setTimeout(callback, 0)
+				if (typeof queueMicrotask === 'function') {
```
robertsLando: **[Major] · Design** + **[Minor] · Correctness**

Two things:

1. `nextTick` is public API — `src/mqtt.ts` does `export * from './lib/shared'`. This fallback flips macrotask to microtask ordering for external callers, with no changelog note. Worth calling out as a behaviour change, or keeping the export's semantics and fixing only the internal path.
2. The guard one line up is `typeof process?.nextTick`, which still *evaluates* `process`. With no `process` binding at all that throws `ReferenceError` (verified in node) before this `queueMicrotask` branch can run — so the new fallback is unreachable in the one environment it targets. `src/lib/connect/index.ts:125` already does this correctly with `typeof process !== 'undefined'`.

scripts/test-browser-process-build.js
```diff
@@ -0,0 +1,48 @@
...
+const { build } = require('esbuild')
+const { options, buildWithProcessCheck } = require('../esbuild.js')
+
+async function test() {
+    const result = await buildWithProcessCheck({
+        ...options, entryPoints: undefined, outfile: undefined, write: false,
+        globalName: 'probe',
+        stdin: {
+            contents: 'import {nextTick as a} from "process"; import {nextTick as b} from "process/"; export {a,b}',
+            resolveDir: process.cwd(),
```
robertsLando: **[Minor] · Robustness**

`resolveDir: process.cwd()` makes this resolve `process` against the invoking cwd, so the assertion's meaning silently changes with where it runs. The call site has the same issue — `test/node/browser-process.ts:116` passes the relative path `'scripts/test-browser-process-build.js'` with no `cwd` option, inheriting the runner's cwd.

That is the repo root under `npm test`, so CI is fine today, but both should anchor to `__dirname`.

scripts/test-browser-process-build.js
```diff
@@ -0,0 +1,48 @@
+const assert = require('node:assert/strict')
```
robertsLando, line 1: **[Minor] · DRY**

A bespoke assert + `runInNewContext` harness invoked via `execFileSync` sits beside the existing `node:test`/chai suite, and its second half re-tests the `shared.ts` microtask fallback already covered at `test/node/browser-process.ts:77-88`.

**Why:** a second test convention to maintain, plus duplicated coverage. **Fix:** fold into the `node:test` suite, or keep the escape hatch (the tsconfig-baseUrl reason in the comment is legitimate) but drop the duplicated `shared.ts` assertions.

robertsLando (commented):
## Review summary

**Verdict:** Needs work — the diagnosis is right and the shim itself is well built, but the fix doesn't reach a large class of consumers, and the guard meant to prevent regression doesn't detect it.

**Top 3 risks**
1. `buildWithProcessCheck` passes while the old `setTimeout`-based `nextTick` is still in every published bundle — reproduced two different ways.
2. Consumers who resolve `main` → `build/index.js` lose the old React Native fix *and* don't get the new one.
3. The drain is O(n²) on the exact hot path the PR exists to fix.

**Strengths.** The root cause is correctly identified and genuinely non-obvious: mqtt-packet corks the transport and uncorks on `nextTick`, so when React Native throttles background timers `writeToStream()` returns normally while `_write()` never runs. Dropping the global `process.nextTick = setImmediate` mutation is the right call — it was a cross-realm write to the host page's `process`. The queue mechanics check out: FIFO within a drain, nested ticks consumed by the active drain, no double-run (items are `shift()`ed before invocation), and the stall invariant holds including the throw-mid-drain reschedule — the `finally` runs only after the `while` has drained to empty, so no push can slip between `draining = false` and the re-check. `version = ''` correctly steers `process-nextick-args` to its arg-forwarding wrapper. Minified bundle drops 337,781 → 332,604 bytes (−5.2 KB). Eight focused tests, all passing.

**Theme.** Every safety net here checks a proxy rather than the property that matters: the guard checks file paths instead of "is there a second `nextTick` implementation in the bundle", and the tests check the shim instead of what consumers actually resolve to.

### CI wiring — no change needed

Checked, since it came up: the build check is already running on CI. `mqttjs-test.yml` (push + PR to main, Node 22.x/24.x) → `npm run test:node` → `node --test test/node/*.ts` picks up `test/node/browser-process.ts`, whose last case spawns `scripts/test-browser-process-build.js` via `execFileSync`. Ran it against this head: 8/8 pass, 244 ms for the file, 79 ms for the subprocess — no pressure on the 5-minute CI timeout. So the script is not orphaned and doesn't need its own step. What it *asserts* is the problem (see the `esbuild.js` thread).

### Verified clean

Recorded so it isn't re-litigated. Net supply-chain reduction: no new deps, and the npm `process` package plus `@jspm/core/nodelibs/browser/process.js` drop out of the bundle. `env`/`argv` are empty literals, there is no `define`, and esbuild does not auto-inline `NODE_ENV`, so no build-time env is embedded. The shim is IIFE-scoped and never touches `globalThis`; esbuild `inject` binds `process` per module, so nothing shadows a consumer's own `process`. The `onResolve` filter is correctly anchored and does not capture `process-nextick-args`. No prototype-pollution surface in the queue.

Ordering divergence from Node is real but inherent and already documented in the file: a tick queued from inside a promise continuation lands at the tail of the shared microtask queue, so promise jobs queued earlier run first.

### Coverage

Eight dimensions reviewed against a worktree at the PR head: correctness, security, DRY/codebase-fit, design/API/backcompat, performance, tests, operability, readability. No prior unresolved review threads.

PR #2048, opened 2026-04-08
feat(mqtt5): implement re-authentication and handle auth success

abdelrahmanSheref101:
## Description
* This feature was requested at issue #1795 
* This PR implements the **MQTT 5.0 Re-authentication** flow and enhances the handling of `AUTH` packets within the client. This allows for dynamic credential updates or extended authentication exchanges without disconnecting the session.

### Key Changes:
* **MqttClient API:** Added a public `.reauthenticate(options)` method to trigger the re-auth flow.
* **Packet Handling:** Updated `src/lib/handlers/auth.ts` to correctly process **Reason Code 0x00 (Success)** for both initial handshakes and mid-session re-authentication.
* **Event System:** Introduced a new `'reauth'` event that emits upon a successful authentication exchange, providing the server's response packet to the user.
* **Validation:** Added guards to ensure re-authentication is only attempted on MQTT 5.0 connections with a pre-established `authenticationMethod`.

---

## Test Plan
I have added a new suite of tests in `test/client_mqtt5.ts` to ensure stability:
* **[Success Path]**: Verified end-to-end `AUTH` exchange using a mock broker, ensuring the `'reauth'` event triggers correctly.
* **[Protocol Guard]**: Verified the client throws an error when attempting re-authentication on non-v5 (e.g., v3.1.1 or v4) connections.
* **[Constraint Check]**: Confirmed validation logic prevents calls if no `authenticationMethod` was defined during the initial connection.

**Local Results:**
* `npm test`: Passes (verified specifically for MQTT 5.0 client tests).

PR #2057, opened 2026-09-06
feat(mqtt5): add client-initiated re-authentication (reauthenticate)

### LINKED ISSUE #1795: [feat] Add support for MQTT 5 refresh auth
Specs: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Re-authentication

### Discussed in https://github.com/mqttjs/MQTT.js/discussions/1794

<div type='discussions-op-text'>

<sup>Originally posted by **sseiber** February 26, 2024</sup>
I'm using JWT Token authentication sucessfully by setting the mqtt client properties before connecting:
```
properties: {
    authenticationMethod: 'OAUTH2-JWT', // for Azure EventGrid
    authenticationData: <auth-token>
}
```
I need to be able to re-authenticate with the service when the token needs to be refreshed by sending an auth packet:
```
{
    cmd: 'auth',
    reasonCode: 0x19,
    properties: {
        authenticationMethod: 'OAUTH2-JWT',
        authenticationData: <auth-token>
    }
}
```
However, I don't see a way to send a specific packet. I'm following an example that uses MQTTnet to provide this functionality (see SendExtendedAuthenticationExchangeDataAsync):
https://github.com/dotnet/MQTTnet/blob/f82d382f23ec8388ed8a33c48b7ede08e2c18622/Source/MQTTnet/Client/MqttClient.cs#L296

Is there a method or standard pattern in mqtt.js for doing this?</div>

shirasu0206:
Closes #1795

Adds `client.reauthenticate(options, callback)` and `client.reauthenticateAsync(options)`, implementing MQTT 5.0 re-authentication (spec section 4.12.1) on an established connection, e.g. to hand a refreshed token to the broker without reconnecting.

### Behaviour

- sends an `AUTH` packet with reason code `0x19` (Re-authenticate) using the `authenticationMethod` of the CONNECT packet
- `AUTH 0x18` (Continue authentication) is forwarded to `client.handleAuth()` and the packet it provides is sent; an error from `handleAuth()` is reported to the `reauthenticate()` callback, so the callback never hangs
- `AUTH 0x00` (Success) completes the exchange: the callback is called first, then the new `'reauth'` event is emitted
- an `AUTH` packet whose `authenticationMethod` is not the negotiated one is rejected before `handleAuth()` sees it ([MQTT-4.12.0-5], code `0x8C`)
- a pending exchange is aborted after `reauthTimeout` ms (new option, default 30 s, `0` disables; the deadline is refreshed on each accepted `0x18`) and on teardown; the exchange is also capped at 10 rounds
- a failed re-authentication leaves the connection up - that is the application's call - and is recorded in `client.lastReauthError`
- **breaking:** an `AUTH` packet received after CONNACK while no re-authentication is in progress used to be forwarded to `client.handleAuth()`. Only the client may start a re-authentication, so it is now a protocol error (code `0x82`) that closes the connection. Anyone relying on broker-initiated mid-session enhanced authentication is affected; documented in the README.
- the pre-CONNACK enhanced authentication flow is unchanged apart from sharing the `handleAuth()` continuation with the re-auth path, which gives it the missing "no packet provided" guard and makes it reject a non-`0x18` code before calling the hook instead of after
- `client.log()` calls on every path of `reauthenticate()` and `handlers/auth.ts`

### Notes

- This builds on the review feedback given on #2048 (unsolicited AUTH, hanging callback on `handleAuth` errors, missing logging, listener registered after the event). That PR has been inactive since May, so I am proposing a fresh implementation; happy to defer to it or to rebase if its author returns.
- `mqtt-packet` only parses `AUTH` reason codes `0x00`/`0x18`/`0x19`, so a "refused" `AUTH` cannot reach the handler. The only code that can reach the `default` branch is `0x19`, which a broker must never send to a client, so it is reported as a protocol error.

### Tests

- 26 new tests in `test/node/client_mqtt5.ts`, including the pre-CONNACK enhanced authentication exchange, which had no coverage anywhere before this PR
- `npm run test:node` on Node 22: 944 tests, 939 passed, 0 failed, 5 skipped; coverage gate green
- `tsc --noEmit`, `eslint`, `prettier --check`: clean

README documents the new methods, the `reauthTimeout` option and the `'reauth'` event.

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -22,6 +22,7 @@ import { type ClientRequestArgs } from 'http'
 import * as validations from './validations'
 import Store, { type IStore } from './store'
 import handlePacket from './handlers'
+import { RC_REAUTHENTICATE } from './handlers/auth'
 import DefaultMessageIdProvider, {
 	type IMessageIdProvider,
 } from './default-message-id-provider'
@@ -52,6 +53,13 @@ const setImmediate =
 			callback(...args)
 		})
 	}) as typeof globalThis.setImmediate)
+
+/**
+ * Upper bound on the AUTH 0x18 challenges a broker may drive in one
+ * re-authentication exchange. Each one runs application `handleAuth()` work
+ * (typically a KDF), so an unbounded exchange is work the broker can force.
+ */
+const MAX_REAUTH_ROUNDS = 10
 
 const defaultConnectOptions: IClientOptions = {
 	keepalive: 60,
@@ -431,7 +439,7 @@ export type PacketCallback = (
 	packet?: Packet,
 ) => any
 export type CloseCallback = (error?: Error) => void
-export type AuthCallback = (err?: Error | null, packet?: IAuthPacket) => void
+export type AuthCallback = GenericCallback<IAuthPacket>
 /**
  * Properties of the AUTH packet sent by `reauthenticate()`. The
  * `authenticationMethod` is always the one used in the CONNECT packet.
@@ -530,6 +538,20 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	/** Callback of the pending re-authentication, if any */
 	private _reauthCallback: AuthCallback | null
 
+	/**
+	 * Why the last re-authentication failed, or null when the last one
+	 * succeeded or none has run. A failed re-authentication does not close the
+	 * connection, so this is what tells a healthy client from one still running
+	 * on credentials the broker has refused.
+	 */
+	public lastReauthError: Error | null
+
+	/** number of AUTH 0x18 rounds the broker has driven in the current exchange */
+	private _reauthRounds: number
+
+	/** the `_sendPacket` callback of the current exchange, parked on `drain` when the stream is backed up */
+	private _reauthSendCb: DoneCallback | null
+
 	private _storeProcessing: boolean
 
 	/** keep a reference of packets that have been successfully processed from outgoing store  */
@@ -567,6 +589,19 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			} else {
 				this.options[k] = options[k]
 			}
+		}
+
+		// A non-finite or negative timeout would silently disable the abort
+		// timer, turning every failed re-authentication into a permanent hang.
+		// Only an explicit 0 disables it.
+		const { reauthTimeout } = this.options
+		if (
+			reauthTimeout !== 0 &&
+			(typeof reauthTimeout !== 'number' ||
+				!Number.isFinite(reauthTimeout) ||
+				reauthTimeout < 0)
+		) {
+			this.options.reauthTimeout = defaultConnectOptions.reauthTimeout
 		}
 
 		this.log = this.options.log || _debug('mqttjs:client')
@@ -663,6 +698,9 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		this.reauthTimer = null
 		this._reauthPending = false
 		this._reauthCallback = null
+		this._reauthRounds = 0
+		this._reauthSendCb = null
+		this.lastReauthError = null
 		// Is processing store?
 		this._storeProcessing = false
 		// Packet Ids are put into the store during store processing
@@ -737,14 +775,9 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
 
-			if (this._reauthPending) {
-				this.log('close :: aborting pending re-authentication')
-				this._finishReauth(
-					new Error(
-						'reauthenticate: connection closed before the broker answered',
-					),
-				)
-			}
+			// Backstop only: `_cleanUp` already aborts on the teardown paths
+			// the client controls. This catches a close driven by the peer.
+			this._abortReauth('connection closed before the broker answered')
 
 			this._destroyKeepaliveManager()
 
@@ -1740,6 +1773,13 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * @api public
 	 * @example client.reauthenticate({ authenticationData: Buffer.from(newToken) }, (err) => {})
 	 */
+	public reauthenticate(callback?: AuthCallback): MqttClient
+
+	public reauthenticate(
+		options: ReauthenticateOptions,
+		callback?: AuthCallback,
+	): MqttClient
+
 	public reauthenticate(
 		options?: ReauthenticateOptions | AuthCallback,
 		callback?: AuthCallback,
@@ -1753,12 +1793,10 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 
 		const fail = (message: string) => {
 			this.log('reauthenticate :: %s', message)
-			const err = new Error(`reauthenticate: ${message}`)
-			if (callback) {
-				callback(err)
-			} else {
-				this.emit('error', err)
-			}
+			this._dispatchError(
+				new Error(`reauthenticate: ${message}`),
+				callback,
+			)
 			return this
 		}
 
@@ -1766,7 +1804,17 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			return fail('re-authentication is only supported in MQTT 5.0')
 		}
 
-		if (!this.connected || this.disconnecting) {
+		if (this._checkDisconnecting(callback)) {
+			return this
+		}
+
+		// `connected` is not enough: a graceful `end()` half-closes the socket,
+		// so a write here would never be answered.
+		if (
+			!this.connected ||
+			this.stream?.destroyed ||
+			(this.stream as any)?.writableEnded
+		) {
 			return fail('client is not connected')
 		}
 
@@ -1783,38 +1831,98 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			return fail('a re-authentication is already in progress')
 		}
 
+		let properties: IAuthPacket['properties']
+		try {
+			properties = this._buildReauthProperties(
+				options as ReauthenticateOptions,
+				authenticationMethod,
+			)
+		} catch (err) {
+			return fail(err.message)
+		}
+
 		const authPacket: IAuthPacket = {
 			cmd: 'auth',
-			reasonCode: 0x19, // Re-authenticate
-			properties: {
-				...(options as ReauthenticateOptions),
-				authenticationMethod,
-			},
+			reasonCode: RC_REAUTHENTICATE,
+			properties,
 		}
 
 		this._reauthPending = true
 		this._reauthCallback = callback || null
-
-		if (this.options.reauthTimeout > 0) {
-			this.reauthTimer = setTimeout(() => {
-				this.log('reauthenticate :: timeout hit, aborting')
-				this._finishReauth(
-					new Error(
-						'reauthenticate: timed out waiting for the broker AUTH response',
-					),
-				)
-			}, this.options.reauthTimeout)
-		}
+		this._reauthRounds = 0
+		this._armReauthTimer()
 
 		this.log('reauthenticate :: sending AUTH packet (reason code 0x19)')
-		this._sendPacket(authPacket, (err) => {
+		const sendCb: DoneCallback = (err) => {
 			if (err) {
 				this.log('reauthenticate :: failed to send AUTH packet')
 				this._finishReauth(err)
 			}
-		})
+		}
+		this._reauthSendCb = sendCb
+
+		try {
+			this._sendPacket(authPacket, sendCb)
+		} catch (err) {
+			// `_writePacket` serializes inline, so a malformed packet throws
+			// straight back at us. Without this the pending state would stay
+			// set for the life of the connection.
+			this.log('reauthenticate :: sending the AUTH packet threw')
+			this._finishReauth(err)
+		}
 
 		return this
+	}
+
+	/**
+	 * Build the AUTH properties for `reauthenticate()`. Only the three
+	 * documented properties are copied - an unknown or malformed one would make
+	 * mqtt-packet destroy the connection, with the raw token in the error
+	 * message - and `authenticationMethod` is always the negotiated one.
+	 */
+	private _buildReauthProperties(
+		options: ReauthenticateOptions | undefined,
+		authenticationMethod: string,
+	): IAuthPacket['properties'] {
+		const properties: IAuthPacket['properties'] = { authenticationMethod }
+
+		if (!options) {
+			return properties
+		}
+
+		const { authenticationData, reasonString, userProperties } = options
+
+		if (typeof authenticationData !== 'undefined') {
+			if (
+				!Buffer.isBuffer(authenticationData) &&
+				typeof authenticationData !== 'string'
+			) {
+				throw new Error(
+					'authenticationData must be a Buffer or a string',
+				)
+			}
+			properties.authenticationData = authenticationData as Buffer
+		}
+
+		if (typeof reasonString !== 'undefined') {
+			if (typeof reasonString !== 'string') {
+				throw new Error('reasonString must be a string')
+			}
+			properties.reasonString = reasonString
+		}
+
+		if (typeof userProperties !== 'undefined') {
+			if (
+				typeof userProperties !== 'object' ||
+				userProperties === null ||
+				Array.isArray(userProperties)
+			) {
+				throw new Error('userProperties must be an object')
+			}
+			properties.userProperties = userProperties
+		}
+
+		return properties
 	}
 
 	/**
@@ -1912,15 +2020,89 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		}
 	}
 
+	/**
+	 * Report an error to a caller supplied callback, or as an `'error'` event
+	 * when there is none. This is the error contract of `publish`, `subscribe`,
+	 * `unsubscribe` and `reauthenticate`.
+	 */
+	private _dispatchError(err: Error, callback?: GenericCallback<any>) {
+		if (callback && callback !== this.noop) {
+			callback(err)
+		} else {
+			this.emit('error', err)
+		}
+	}
+
 	private _checkDisconnecting(callback?: GenericCallback<any>) {
 		if (this.disconnecting) {
-			if (callback && callback !== this.noop) {
-				callback(new Error('client disconnecting'))
-			} else {
-				this.emit('error', new Error('client disconnecting'))
-			}
+			this._dispatchError(new Error('client disconnecting'), callback)
 		}
 		return this.disconnecting
+	}
+
+	/**
+	 * A protocol violation by the broker is fatal to the connection.
+	 *
+	 * This runs on the packet parsing path, where an `'error'` event with no
+	 * listener - or a listener that throws - would skip the parser's `done()`
+	 * and wedge it while `connected` stays true, so the event is deferred and
+	 * the teardown does not depend on the application having a listener.
+	 */
+	private _onProtocolError(err: ErrorWithReasonCode) {
+		this.log('_onProtocolError :: %s', err.message)
+		setImmediate(() => {
+			if (this.listenerCount('error') > 0) {
+				this.emit('error', err)
+			}
+		})
+		this._cleanUp(true)
+	}
+
+	/** (Re)arm the abort timer of the current re-authentication exchange. */
+	private _armReauthTimer() {
+		clearTimeout(this.reauthTimer)
+		this.reauthTimer = null
+
+		if (this.options.reauthTimeout > 0) {
+			this.reauthTimer = setTimeout(() => {
+				this.log('reauthenticate :: timeout hit, aborting')
+				this._finishReauth(
+					new Error(
+						'reauthenticate: timed out waiting for the broker AUTH response',
+					),
+				)
+			}, this.options.reauthTimeout)
+		}
+	}
+
+	/**
+	 * Account for one broker challenge (AUTH 0x18) of the current exchange.
+	 * Returns false when the exchange has been aborted for exceeding
+	 * `MAX_REAUTH_ROUNDS`; a legitimate round refreshes the deadline.
+	 */
+	private _startReauthRound(): boolean {
+		this._reauthRounds += 1
+
+		if (this._reauthRounds > MAX_REAUTH_ROUNDS) {
+			this.log('reauthenticate :: too many challenges, aborting')
+			this._finishReauth(
+				new Error(
+					`reauthenticate: the broker sent more than ${MAX_REAUTH_ROUNDS} authentication challenges`,
+				),
+			)
+			return false
+		}
+
+		this._armReauthTimer()
+		return true
+	}
+
+	/** Abort a re-authentication that can no longer complete. No-op when none is pending. */
+	private _abortReauth(reason: string) {
+		if (this._reauthPending) {
+			this.log('_abortReauth :: %s', reason)
+			this._finishReauth(new Error(`reauthenticate: ${reason}`))
+		}
 	}
 
 	/**
@@ -1981,10 +2163,6 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	}
 
 	/**
-	 * _cleanUp - clean up on connection end
-	 * @api private
-	 */
-	/**
 	 * _finishReauth - complete the pending re-authentication exchange
 	 *
 	 * Clears the timer and the pending state, then reports the outcome to the
@@ -1992,8 +2170,9 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * given). On success the `'reauth'` event is emitted.
 	 *
 	 * @param {Error | null} err - failure reason, or null on success
-	 * @param {IAuthPacket} [packet] - the final AUTH packet received, only
-	 * reported to the callback on success
+	 * @param {IAuthPacket} [packet] - the last AUTH packet received, reported to
+	 * the callback on success and on a broker side failure: its `reasonString`
+	 * and `userProperties` are the only record of why the broker refused
 	 */
 	private _finishReauth(err: Error | null, packet?: IAuthPacket) {
 		if (!this._reauthPending) {
@@ -2011,26 +2190,40 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		clearTimeout(this.reauthTimer)
 		this.reauthTimer = null
 
+		// The send callback is parked on 'drain' while the stream is backed up;
+		// leaving it there leaks one listener per aborted exchange.
+		if (this._reauthSendCb) {
+			this.stream?.removeListener('drain', this._reauthSendCb)
+			this._reauthSendCb = null
+		}
+
 		const callback = this._reauthCallback
 		this._reauthPending = false
 		this._reauthCallback = null
+		this._reauthRounds = 0
 
 		if (err) {
-			if (callback) {
-				callback(err)
-			} else {
-				this.emit('error', err)
-			}
+			this.lastReauthError = err
+			this._dispatchError(err, callback)
 			return
 		}
 
+		this.lastReauthError = null
 		if (callback) {
 			callback(null, packet)
 		}
 		this.emit('reauth', packet)
 	}
 
+	/**
+	 * _cleanUp - clean up on connection end
+	 * @api private
+	 */
 	private _cleanUp(forced: boolean, done?: DoneCallback, opts = {}) {
+		// Synchronous teardown entry point: a graceful end() only half-closes
+		// the socket, so waiting for 'close' can strand the exchange forever.
+		this._abortReauth('connection closed before the broker answered')
+
 		if (done) {
 			this.log('_cleanUp :: done callback provided for on stream close')
 			this.stream.on('close', done)
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -65,6 +65,7 @@ const defaultConnectOptions: IClientOptions = {
 	subscribeBatchSize: null,
 	writeCache: true,
 	timerVariant: 'auto',
+	reauthTimeout: 30 * 1000,
 }
 
 export type BaseMqttProtocol =
@@ -182,6 +183,12 @@ export interface IClientOptions extends ISecureClientOptions {
 	 * 30 * 1000 milliseconds, time to wait before a CONNACK is received
 	 */
 	connectTimeout?: number
+	/**
+	 * 30 * 1000 milliseconds, time to wait for the broker to complete a
+	 * re-authentication exchange started with `reauthenticate()`
+	 * (MQTT 5.0 only). Set to 0 to disable the timeout.
+	 */
+	reauthTimeout?: number
 
 	/**
 	 * a Store for the incoming packets
@@ -424,6 +431,15 @@ export type PacketCallback = (
 	packet?: Packet,
 ) => any
 export type CloseCallback = (error?: Error) => void
+export type AuthCallback = (err?: Error | null, packet?: IAuthPacket) => void
+/**
+ * Properties of the AUTH packet sent by `reauthenticate()`. The
+ * `authenticationMethod` is always the one used in the CONNECT packet.
+ */
+export type ReauthenticateOptions = Pick<
+	NonNullable<IAuthPacket['properties']>,
+	'authenticationData' | 'reasonString' | 'userProperties'
+>
 
 export interface MqttClientEventCallbacks {
 	connect: OnConnectCallback
@@ -437,6 +453,7 @@ export interface MqttClientEventCallbacks {
 	reconnect: VoidCallback
 	offline: VoidCallback
 	outgoingEmpty: VoidCallback
+	reauth: (packet: IAuthPacket) => void
 }
 
 /**
@@ -503,6 +520,15 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	private connackTimer: NodeJS.Timeout
 
 	private reconnectTimer: NodeJS.Timeout
+
+	/** Timer used to abort a re-authentication that receives no answer */
+	private reauthTimer: NodeJS.Timeout
+
+	/** True while a re-authentication exchange is in progress */
+	private _reauthPending: boolean
+
+	/** Callback of the pending re-authentication, if any */
+	private _reauthCallback: AuthCallback | null
 
 	private _storeProcessing: boolean
 
@@ -633,6 +659,10 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 		this.connackTimer = null
 		// Reconnect timer
 		this.reconnectTimer = null
+		// Re-authentication state
+		this.reauthTimer = null
+		this._reauthPending = false
+		this._reauthCallback = null
 		// Is processing store?
 		this._storeProcessing = false
 		// Packet Ids are put into the store during store processing
@@ -706,6 +736,15 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
+
+			if (this._reauthPending) {
+				this.log('close :: aborting pending re-authentication')
+				this._finishReauth(
+					new Error(
+						'reauthenticate: connection closed before the broker answered',
+					),
+				)
+			}
 
 			this._destroyKeepaliveManager()
 
@@ -1683,6 +1722,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	}
 
 	/**
+	 * reauthenticate - start an MQTT 5.0 re-authentication exchange
+	 * (MQTT 5.0 spec, section 4.12.1).
+	 *
+	 * Sends an AUTH packet with reason code 0x19 (Re-authenticate) using the
+	 * `authenticationMethod` of the CONNECT packet. The broker answers with
+	 * AUTH 0x18 (Continue authentication, forwarded to `handleAuth()`) or
+	 * AUTH 0x00 (Success), after which the `'reauth'` event is emitted.
+	 *
+	 * @param {Object} [options] - AUTH packet properties
+	 * @param {Buffer} [options.authenticationData] - Binary Data for the authentication method
+	 * @param {String} [options.reasonString] - Reason String
+	 * @param {Object} [options.userProperties] - User Properties
+	 * @param {Function} [callback] - fired when the exchange completes: `(err, packet)`.
+	 *   When omitted, failures are emitted as `'error'` events.
+	 * @returns {MqttClient} this - for chaining
+	 * @api public
+	 * @example client.reauthenticate({ authenticationData: Buffer.from(newToken) }, (err) => {})
+	 */
+	public reauthenticate(
+		options?: ReauthenticateOptions | AuthCallback,
+		callback?: AuthCallback,
+	): MqttClient {
+		this.log('reauthenticate :: (%s)', this.options.clientId)
+
+		if (typeof options === 'function') {
+			callback = options
+			options = undefined
+		}
+
+		const fail = (message: string) => {
+			this.log('reauthenticate :: %s', message)
+			const err = new Error(`reauthenticate: ${message}`)
+			if (callback) {
+				callback(err)
+			} else {
+				this.emit('error', err)
+			}
+			return this
+		}
+
+		if (this.options.protocolVersion !== 5) {
+			return fail('re-authentication is only supported in MQTT 5.0')
+		}
+
+		if (!this.connected || this.disconnecting) {
+			return fail('client is not connected')
+		}
+
+		const authenticationMethod =
+			this.options.properties?.authenticationMethod
+
+		if (!authenticationMethod) {
+			return fail(
+				'an authenticationMethod must be set in the CONNECT properties',
+			)
+		}
+
+		if (this._reauthPending) {
+			return fail('a re-authentication is already in progress')
+		}
+
+		const authPacket: IAuthPacket = {
+			cmd: 'auth',
+			reasonCode: 0x19, // Re-authenticate
+			properties: {
+				...(options as ReauthenticateOptions),
+				authenticationMethod,
+			},
+		}
+
+		this._reauthPending = true
+		this._reauthCallback = callback || null
+
+		if (this.options.reauthTimeout > 0) {
+			this.reauthTimer = setTimeout(() => {
+				this.log('reauthenticate :: timeout hit, aborting')
+				this._finishReauth(
+					new Error(
+						'reauthenticate: timed out waiting for the broker AUTH response',
+					),
+				)
+			}, this.options.reauthTimeout)
+		}
+
+		this.log('reauthenticate :: sending AUTH packet (reason code 0x19)')
+		this._sendPacket(authPacket, (err) => {
+			if (err) {
+				this.log('reauthenticate :: failed to send AUTH packet')
+				this._finishReauth(err)
+			}
+		})
+
+		return this
+	}
+
+	/**
+	 * reauthenticateAsync - promise wrapper around `reauthenticate()`
+	 *
+	 * @param {Object} [options] - see `reauthenticate()`
+	 * @returns {Promise<IAuthPacket>} resolves with the final AUTH packet of the broker
+	 * @api public
+	 */
+	public reauthenticateAsync(
+		options?: ReauthenticateOptions,
+	): Promise<IAuthPacket> {
+		return new Promise((resolve, reject) => {
+			this.reauthenticate(options, (err, packet) => {
+				if (err) {
+					reject(err)
+				} else {
+					resolve(packet)
+				}
+			})
+		})
+	}
+
+	/**
 	 * PRIVATE METHODS
 	 * =====================
 	 * */
@@ -1828,6 +1984,51 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * _cleanUp - clean up on connection end
 	 * @api private
 	 */
+	/**
+	 * _finishReauth - complete the pending re-authentication exchange
+	 *
+	 * Clears the timer and the pending state, then reports the outcome to the
+	 * `reauthenticate()` callback (or emits `'error'` when no callback was
+	 * given). On success the `'reauth'` event is emitted.
+	 *
+	 * @param {Error | null} err - failure reason, or null on success
+	 * @param {IAuthPacket} [packet] - the final AUTH packet received
+	 */
+	private _finishReauth(err: Error | null, packet?: IAuthPacket) {
+		if (!this._reauthPending) {
+			this.log(
+				'_finishReauth :: no re-authentication in progress, ignoring',
+			)
+			return
+		}
+
+		this.log(
+			'_finishReauth :: %s',
+			err ? `failed: ${err.message}` : 'success',
+		)
+
+		clearTimeout(this.reauthTimer)
+		this.reauthTimer = null
+
+		const callback = this._reauthCallback
+		this._reauthPending = false
+		this._reauthCallback = null
+
+		if (err) {
+			if (callback) {
+				callback(err)
+			} else {
+				this.emit('error', err)
+			}
+			return
+		}
+
+		if (callback) {
+			callback(null, packet)
+		}
+		this.emit('reauth', packet)
+	}
+
 	private _cleanUp(forced: boolean, done?: DoneCallback, opts = {}) {
 		if (done) {
 			this.log('_cleanUp :: done callback provided for on stream close')
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -1992,7 +1992,8 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * given). On success the `'reauth'` event is emitted.
 	 *
 	 * @param {Error | null} err - failure reason, or null on success
-	 * @param {IAuthPacket} [packet] - the final AUTH packet received
+	 * @param {IAuthPacket} [packet] - the final AUTH packet received, only
+	 * reported to the callback on success
 	 */
 	private _finishReauth(err: Error | null, packet?: IAuthPacket) {
 		if (!this._reauthPending) {
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -446,8 +446,14 @@ export type AuthCallback = GenericCallback<IAuthPacket>
  */
 export type ReauthenticateOptions = Pick<
 	NonNullable<IAuthPacket['properties']>,
-	'authenticationData' | 'reasonString' | 'userProperties'
->
+	'reasonString' | 'userProperties'
+> & {
+	/**
+	 * The new credential. `mqtt-packet` types this as a Buffer, but a string is
+	 * accepted too and is written as UTF-8.
+	 */
+	authenticationData?: Buffer | string
+}
 
 export interface MqttClientEventCallbacks {
 	connect: OnConnectCallback
--- a/src/lib/handlers/ack.ts
+++ b/src/lib/handlers/ack.ts
@@ -3,7 +3,7 @@
 import { type PacketHandler, ErrorWithReasonCode } from '../shared'
 
 export const ReasonCodes = {
-	0: '',
+	0: 'Success',
 	1: 'Unacceptable protocol version',
 	2: 'Identifier rejected',
 	3: 'Server unavailable',
@@ -11,6 +11,8 @@ export const ReasonCodes = {
 	5: 'Not authorized',
 	16: 'No matching subscribers',
 	17: 'No subscription existed',
+	24: 'Continue authentication',
+	25: 'Re-authenticate',
 	128: 'Unspecified error',
 	129: 'Malformed Packet',
 	130: 'Protocol Error',
@@ -47,6 +49,13 @@ export const ReasonCodes = {
 	161: 'Subscription Identifiers not supported',
 	162: 'Wildcard Subscriptions not supported',
 }
+
+/**
+ * Human readable text for a reason code, falling back to the numeric value so
+ * an unlisted code still names itself in the error message.
+ */
+export const reasonCodeText = (code: number): string =>
+	ReasonCodes[code] || `Unknown reason code 0x${Number(code).toString(16)}`
 
 const handleAck: PacketHandler = (client, packet) => {
 	/* eslint no-fallthrough: "off" */
--- a/src/lib/handlers/auth.ts
+++ b/src/lib/handlers/auth.ts
@@ -2,6 +2,10 @@ import { type IAuthPacket } from 'mqtt-packet'
 import { ErrorWithReasonCode, type PacketHandler } from '../shared'
 import { ReasonCodes } from './ack'
 
+const RC_SUCCESS = 0x00
+const RC_CONTINUE_AUTHENTICATION = 0x18
+const RC_PROTOCOL_ERROR = 0x82
+
 const handleAuth: PacketHandler = (
 	client,
 	packet: IAuthPacket & { returnCode: number },
@@ -9,6 +13,8 @@ const handleAuth: PacketHandler = (
 	const { options } = client
 	const version = options.protocolVersion
 	const rc = version === 5 ? packet.reasonCode : packet.returnCode
+
+	client.log('handleAuth :: received AUTH packet, reason code %d', rc)
 
 	if (version !== 5) {
 		const err = new ErrorWithReasonCode(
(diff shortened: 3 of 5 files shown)


src/lib/client.ts
```diff
@@ -707,6 +737,15 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
 
+			if (this._reauthPending) {
+				this.log('close :: aborting pending re-authentication')
```
robertsLando: **[Blocker] · Correctness**

The pending-reauth abort moved from `_cleanUp` (synchronous) into the `'close'` handler, which is only an echo of `stream.on('close')` (`client.ts:884-890`). A graceful `end()` calls `stream.end()`, which half-closes — `'close'` fires only when the peer FINs back.

**Why:** A broker that holds the socket open leaves `_reauthPending === true` forever. With `reauthTimeout: 0` (documented as "disable the timeout") the callback never fires, and every later `reauthenticate()` — including after `reconnect()`, which resets `disconnecting` but not the reauth state (`client.ts:1700-1713`) — fails permanently with `a re-authentication is already in progress`. #2048 did this inside `_cleanUp` and additionally guarded `stream.destroyed || stream.writableEnded`; both were dropped in the port.

**Fix:** Clear the reauth state at the teardown entry points and keep the `'close'` handler as a backstop — `_finishReauth` is already idempotent.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+		const authPacket: IAuthPacket = {
+			cmd: 'auth',
+			reasonCode: 0x19, // Re-authenticate
+			properties: {
+				...(options as ReauthenticateOptions),
+				authenticationMethod,
+			},
+		}
+
+		this._reauthPending = true
```
robertsLando, line 1856: **[Blocker] · Security + Correctness**

`_reauthPending = true` is set before the synchronous, fallible `_sendPacket`. `_writePacket` calls `mqttPacket.writeToStream(...)` inline (`client.ts:2219`), so a serialization throw — e.g. an `authenticationData` that is neither Buffer nor string, which is unvalidated at runtime — escapes `reauthenticate()` with the flag stuck true.

**Why:** Same permanent-stuck state as the `'close'` finding, but reachable from ordinary caller error rather than broker behaviour. Worse, while stranded the client accepts and answers *any* broker-initiated AUTH for the life of the connection — exactly the state the new protocol-error check at `auth.ts:57` exists to prevent.

**Fix:** Validate the options, and set/clear the pending state on a path a throw cannot bypass.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+			'error',
+			new ErrorWithReasonCode(
+				'Protocol error: received AUTH packet while no re-authentication is in progress',
+				RC_PROTOCOL_ERROR,
+			),
+		)
+		return
+	}
+
+	switch (rc) {
```
robertsLando, line 133: **[Blocker] · Security**

The received AUTH's `properties.authenticationMethod` is never compared to the negotiated one before dispatch.

**Why:** MQTT 5 requires every AUTH in an exchange to carry the CONNECT's method ([MQTT-4.12.0-5]), but the client does not enforce the broker's compliance. A hostile broker can answer a `SCRAM-SHA-256` re-auth with a `0x18` challenge labelled as a weaker method, and any `handleAuth()` that dispatches on that field is downgraded into computing a response under the attacker's chosen scheme.

**Fix:** Reject a mismatched or absent method before calling `handleAuth()`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+								'handleAuth :: failed to send next AUTH packet',
+							)
+							client['_finishReauth'](sendErr)
+						}
+					})
+				},
+			)
+			break
+
+		default:
```
robertsLando, line 154: **[Blocker] · Tests**

The `default:` branch — broker refuses re-authentication with a code other than 0x00/0x18 — has no test.

**Why:** Refusal is the *primary* failure mode of re-authentication and is completely unverified; a broken `default:` would still pass CI. It is genuinely reachable: mqtt-packet parses 0x19, so a broker echoing it lands here.

**Fix:** Have `buildReauthServer` answer the 0x19 with a refusal code and assert the callback receives an `ErrorWithReasonCode` carrying it.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+	}
+
+	// Re-authentication (MQTT 5.0 spec, section 4.12.1): the exchange can only
+	// be started by the client, so an AUTH packet received while no
+	// re-authentication is in progress is a protocol error.
+	if (!client['_reauthPending']) {
+		client.log(
+			'handleAuth :: unexpected AUTH packet while no re-authentication is in progress',
+		)
+		client.emit(
```
robertsLando: **[Major] · Security + Operability**

Unsolicited AUTH does an unconditional `emit('error')`. `TypedEventEmitter` is a plain `EventEmitter`, so with no `'error'` listener the process throws — and `handlers/index.ts:52-55` calls `handleAuth(client, packet)` *then* `done()`, so a synchronous throw skips `done()` and wedges the parse pipeline while `connected` stays `true`.

To be fair on provenance: I checked `main`, and the pre-PR code also emitted synchronously for `rc !== 24`, so the no-listener crash is **pre-existing, not introduced here**. What is new is user `reauthenticate`/`handleAuth` callback code running synchronously on this path, where a throw has the same effect.

**Why:** This PR adds the unsolicited-AUTH check as a hardening measure, and implementing it via a crashing emit undercuts that — one broker-controlled packet, no client action required.

**Fix:** Treat the protocol violation as fatal to the connection rather than depending on an app listener, and keep user callbacks off the path where a throw can skip `done()`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+				}
+			},
+		)
+		return
+	}
+
+	// Re-authentication (MQTT 5.0 spec, section 4.12.1): the exchange can only
+	// be started by the client, so an AUTH packet received while no
+	// re-authentication is in progress is a protocol error.
+	if (!client['_reauthPending']) {
```
robertsLando, line 99: **[Major] · Design + BackCompat**

Undocumented contract break: a post-CONNACK AUTH used to be forwarded to `client.handleAuth()` and the reply sent; it now errors whenever `_reauthPending` is false.

**Why:** Anyone using broker-initiated enhanced auth mid-session — or receiving a slow broker answer that arrives after `reauthTimeout` fired — silently starts getting `'error'` instead of their auth hook.

**Fix:** Call it out as a BREAKING note in the README/changelog, or keep the `handleAuth()` forward as a fallback and only error when the hook is untouched.

src/lib/client.ts
```diff
@@ -1828,6 +1984,52 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+
+		clearTimeout(this.reauthTimer)
+		this.reauthTimer = null
+
+		const callback = this._reauthCallback
+		this._reauthPending = false
+		this._reauthCallback = null
+
+		if (err) {
+			if (callback) {
```
robertsLando: **[Major] · Design + Operability**

`_finishReauth` accepts `packet` but discards it on the error path (`callback(err)`), so the packet that `auth.ts:115` passes is dead — while `AuthCallback` advertises `(err, packet)` and the README documents it.

**Why:** The broker's `reasonString`/`userProperties` are the only thing saying *why* re-auth was refused (expired token vs wrong method), and they never reach the caller or the logs.

**Fix:** Forward the packet on failure, or drop the argument at the call site so the contract is honest.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+							new ErrorWithReasonCode(
+								'Re-authentication failed: client.handleAuth() did not provide an AUTH packet to continue the exchange',
+								rc,
+							),
+						)
+						return
+					}
+
+					client.log('handleAuth :: sending next AUTH packet')
+					client['_sendPacket'](packet2, (sendErr) => {
```
robertsLando: **[Major] · Security + Correctness**

The timeout is armed once in `reauthenticate()` and never rearmed across `0x18` rounds, and nothing caps the number of round-trips.

**Why:** A legitimate multi-round SCRAM exchange must finish inside the single original window or it aborts mid-flight despite the broker responding. In the other direction, with `reauthTimeout: 0` there is no bound at all — a broker can loop challenges as fast as the socket allows, driving app-side KDF work on each one.

**Fix:** Reset the deadline on each accepted `0x18`, and cap iterations independently of the timer.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+
+		if (this._reauthPending) {
+			return fail('a re-authentication is already in progress')
+		}
+
+		const authPacket: IAuthPacket = {
+			cmd: 'auth',
+			reasonCode: 0x19, // Re-authenticate
+			properties: {
+				...(options as ReauthenticateOptions),
```
robertsLando: **[Major] · Security**

`{ ...options, authenticationMethod }` spreads unvalidated caller input straight onto the wire. (The pinning of `authenticationMethod` after the spread is correct — that part is right.)

**Why:** An unknown key makes mqtt-packet call `stream.destroy(new Error('Invalid property x: <value>'))`, dropping the live connection — and for a bad `authenticationData` that error message carries the raw token into whatever the app logs.

**Fix:** Allowlist the three documented properties instead of spreading.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+				authenticationMethod,
+			},
+		}
+
+		this._reauthPending = true
+		this._reauthCallback = callback || null
+
+		if (this.options.reauthTimeout > 0) {
```
robertsLando: **[Major] · Correctness**

`this.options.reauthTimeout > 0` silently disables the timer for `NaN`, negative, or non-numeric input, and nothing validates the option at construction.

**Why:** Combined with the two stranding blockers above, a typo'd option turns every failure into a permanent hang rather than a bounded one.

**Fix:** Normalize at construction; treat non-positive or non-finite as the default rather than as "disabled".

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+			this.reauthTimer = setTimeout(() => {
+				this.log('reauthenticate :: timeout hit, aborting')
+				this._finishReauth(
+					new Error(
```
robertsLando: **[Major] · Operability**

On timeout — and on broker refusal — the exchange is abandoned but the connection stays up on the old credentials, with nothing recording that re-auth failed.

**Why:** Spec 4.12.1 expects a failed re-authentication to end the connection. As written, a client running on an expired or revoked token is indistinguishable from a healthy one, and a late broker AUTH afterwards reports a protocol error — misdiagnosing the client's own timeout as broker misbehaviour.

**Fix:** Disconnect on failure or document the application's obligation, and keep a last-reauth-failed marker for the diagnostic.

src/lib/client.ts
```diff
@@ -424,6 +431,15 @@ export type PacketCallback = (
 	packet?: Packet,
 ) => any
 export type CloseCallback = (error?: Error) => void
+export type AuthCallback = (err?: Error | null, packet?: IAuthPacket) => void
```
robertsLando: **[Major] · Design**

`AuthCallback` types `err` as `Error`, but `handlers/auth.ts` hands back `ErrorWithReasonCode`. It also duplicates `GenericCallback<IAuthPacket>`, already exported from `src/lib/shared.ts:9`.

**Why:** TS callers must cast to read `.code` — i.e. to find out *why* re-auth was refused. Siblings `PacketCallback`/`OnErrorCallback` union both types. This is the same point raised on #2048's `IAuthPacketCallback` and it did not make the port.

**Fix:** Widen the union, or reuse the existing `GenericCallback<T>`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+					})
+				},
+			)
+			break
+
+		default:
+			// An AUTH packet can only carry 0x00, 0x18 or 0x19, so this is
+			// always 0x19: a broker must never ask a client to re-authenticate.
+			client.log(
+				'handleAuth :: unexpected AUTH reason code %d from broker',
```
robertsLando: **[Major] · Operability**

`ReasonCodes` (`handlers/ack.ts:5-48`) has no entry for 24 (0x18) or 25 (0x19), and maps `0: ''`.

**Why:** A broker echoing 0x19 produces `Re-authentication failed: undefined`; the pre-CONNACK path at line 44 produces `Connection refused: ` for rc=0. The error text is the operator's only artifact and it names neither the code nor the cause. Both were raised on #2048 and are unported.

**Fix:** Add the missing entries plus an unknown-code fallback that prints the numeric code.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
+			client.handleAuth(
+				packet,
+				(err: ErrorWithReasonCode, packet2: IAuthPacket) => {
+					if (err) {
+						client.log('handleAuth :: client.handleAuth() failed')
+						client['_finishReauth'](err)
+						return
+					}
+
+					if (!packet2) {
```
robertsLando: **[Major] · Tests**

The `!packet2` branch is untested — and it is exactly what fires with the *default* `handleAuth()` (`client.ts:770`, which calls `callback()` with no arguments).

**Why:** Any caller who invokes `reauthenticate()` without overriding `handleAuth` and whose broker answers 0x18 hits this in production, with zero coverage of the resulting error.

**Fix:** Add a test with `client.handleAuth` left at its default.

test/node/client_mqtt5.ts
```diff
@@ -1438,4 +1438,430 @@ describe('MQTT 5.0', () => {
 			})
 		},
 	)
+	describe('re-authentication', () => {
```
robertsLando, line 1441: **[Major] · Tests**

Coverage gaps in the new suite, beyond the two flagged inline:

- The pre-CONNACK enhanced-auth branch (`auth.ts:29-52`) has **zero** coverage anywhere in `test/` — `grep -r authenticationMethod test/` outside this block returns nothing — even though this PR restructured that exact function into an `if (!client.connected)` split.
- The `_sendPacket` write-error path, in both `reauthenticate()` (`client.ts:1810`) and the 0x18 continuation (`auth.ts:101`).
- `reauthTimeout: 0` (timer disabled).
- A reauth pending across a reconnect — every test here uses `reconnectPeriod: 0`, so the state reset is unverified.
- That `authenticationMethod` passed in `options` cannot override the CONNECT value — the one hardening property the design leans on.

**Why:** This is a credential path, and the pre-CONNACK gap in particular means a regression in the pre-existing enhanced-auth-during-CONNECT flow would go undetected by this PR or any future one.

**Fix:** Add the five cases. For the override check, pass `authenticationMethod: 'other'` and assert the wire packet still carries the CONNECT value.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+				this.emit('error', err)
+			}
+			return this
+		}
+
+		if (this.options.protocolVersion !== 5) {
+			return fail('re-authentication is only supported in MQTT 5.0')
+		}
+
+		if (!this.connected || this.disconnecting) {
```
robertsLando: **[Minor] · DRY**

Ad-hoc `!this.connected || this.disconnecting` where `_checkDisconnecting(callback)` already exists (`client.ts:1915`) and is used by `publish`/`subscribe`/`unsubscribe` (lines 1031/1189/1445), including the callback-vs-error-event dispatch.

**Why:** The same client state now yields two different user-facing messages (`reauthenticate: client is not connected` vs `client disconnecting`), and the shared helper stops being the single gate.

**Fix:** Route through `_checkDisconnecting` first and add only the `!this.connected` case on top.

robertsLando (commented):
## Summary

**Verdict: Needs work** — the approach is right, but `_reauthPending` can be stranded three different ways, and the new broker-facing path does not validate the peer's authentication method.

**Top 3 risks**
1. Pending-reauth state can stick `true` permanently, making every later `reauthenticate()` fail and (with `reauthTimeout: 0`) dropping the callback silently.
2. The received AUTH's `authenticationMethod` is never checked against the negotiated one.
3. The primary failure mode — broker refuses re-authentication — has no test.

**Strengths.** The four blockers raised on #2048 are genuinely fixed, not just claimed — unsolicited-AUTH rejection, `handleAuth`-error routing into `_finishReauth`, `_finishReauth` privacy, and log coverage all check out, and the tests assert real wire packets against a mock broker. The single-slot pending design is *forced*, not invented: `this.outgoing[]` is keyed by message id and AUTH carries none, so it cannot compose with the existing tracker; the shape used mirrors `connackTimer`, the closest existing analogue. The timer is armed once for the whole exchange, and `authenticationMethod` is correctly pinned *after* the options spread so a caller cannot override it. README and JSDoc coverage are well above the bar for this repo.

**Themes**
- **`_reauthPending`'s lifecycle is neither teardown- nor throw-safe.** Three independent strandings (the two blockers plus the reconnect gap) share one root cause: the flag is set before the fallible send and cleared only by an async event echo.
- **Reason codes now have four homes** — `RC_*` in `handlers/auth.ts`, `ReasonCodes` in `handlers/ack.ts`, `validReasonCodes` in `handlers/publish.ts`, and a bare `0x19` in `client.ts` — and `ReasonCodes` has no 0x18/0x19 entries, so error text degrades to `undefined`.
- **"callback else `emit('error')`" is written three times**, and `emit('error')` is used for both programmer errors and broker-triggered ones.

## Port fidelity vs #2048

Checked this against #2048 and its review, since the PR presents itself as a corrected port. The four blockers it claims to fix *are* fixed. But some of #2048's review backlog did not come across, and two of its hardening fixes were dropped:

- **Dropped:** the `stream.destroyed || stream.writableEnded` guard (added in #2048's `174af9c` specifically to avoid hanging timers on interrupted streams), and the `_cleanUp` cancellation — see the blocker on `client.ts:741`.
- **Still open from #2048's review:** the missing 0x18/0x19 `ReasonCodes` entries; `Connection refused: Success` on a pre-CONNACK rc=0; callback-before-`emit('reauth')` ordering (kept deliberately here, and now locked in by a test); the narrow error type on the callback; `0x19` as a bare literal; the unvalidated property spread; no `Buffer.isBuffer` check; the `reauthenticate: ` error prefix; no try/catch around the user callback; and the `'reauth'` vs `reauthenticate()` naming mismatch.
- **Behaviour changes from #2048 worth a deliberate call:** `authenticationData` went from required to optional (defensible — the spec makes it optional — but the README does not say so), and the default `reauthTimeout` went 15s → 30s.

## Also, not line-anchored

- **README** omits: only one exchange may be in flight; `authenticationMethod` is inherited from CONNECT and cannot be overridden; `.code` is available on the error; the chainable `MqttClient` return; and whether `authenticationData` is required.
- **`_writePacket`** logs `packet: %O` (`client.ts:2213`), which dumps `authenticationData`. Pre-existing for CONNECT passwords, but this puts it on a recurring token-rotation path.
- **Log correlation:** only the entry log carries `clientId`; siblings thread it through every line, so reauth logs cannot be correlated in a multi-client process.
- **`handleAuth` size:** the function now covers four scenarios, three callback levels deep in the 0x18 case, with two structurally identical `client.handleAuth(...)` call sites that route errors differently. Splitting the pre-CONNACK branch from the re-auth switch would make that difference visible.

**Nits** (all optional): `0x19` is a bare literal at `client.ts:1788` while `auth.ts` names its codes — and the `RC_*` constants are not exported, which is *why* it is inlined; errors are prefixed `reauthenticate: ` where no sibling method does; `ReauthenticateOptions` breaks the `IClientPublishOptions`/`IClientSubscribeOptions` family; three spellings of one concept (`'reauth'` / `reauthenticate()` / `reauthTimeout`) where every other pair matches (`reconnect()`/`'reconnect'`); `auth.ts:17` logs `reason code %d` while the feature and the spec use hex; `basePort + 4` is skipped with no comment (confirmed harmless — no collision, `PORTAND327+30..40` stays inside this file's own block per `test/node/helpers/port_list.ts` — but it reads like a dropped test); `test/node/client_mqtt5.ts:1755-1774` relies on an unstated ordering guarantee via `secondFailed`, and around 1520 assertions sit inside `process.nextTick`, where a failure throws uncaught instead of reaching `done`; the unsolicited-AUTH test uses `setTimeout(..., 50)` rather than driving off the server's own connack-sent callback.

## Coverage

Specialists run: Correctness, Security, Design/API/BackCompat, Operability, Tests, DRY, Readability. None came back clean. Not run: Performance — no files in its lane. Design pre-flight verdict: PROCEED.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+		callback?: AuthCallback,
+	): MqttClient {
+		this.log('reauthenticate :: (%s)', this.options.clientId)
+
+		if (typeof options === 'function') {
+			callback = options
+			options = undefined
+		}
+
+		const fail = (message: string) => {
```
robertsLando, line 1800: **[Minor] · DRY**

`fail()`'s "if callback → callback(err), else emit('error')" is byte-for-byte identical to the same block in `_finishReauth` (`client.ts:2018-2022`), and a third copy already lives in `_checkDisconnecting` (`client.ts:1917`).

**Why:** Three copies of one error-routing contract, already divergent: `_checkDisconnecting` carries a `callback !== this.noop` guard that neither new copy has.

**Fix:** Extract one private `_dispatchError(err, callback?)` and have all three use it.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +25,106 @@ const handleAuth: PacketHandler = (
...
-			if (rc === 24) {
-				client.reconnecting = false
-				client['_sendPacket'](packet2)
-			} else {
-				const error = new ErrorWithReasonCode(
-					`Connection refused: ${ReasonCodes[rc]}`,
+	if (!client.connected) {
+		// Enhanced authentication exchange during connection (before CONNACK)
+		client.log('handleAuth :: enhanced authentication during connect')
+		client.handleAuth(
```
robertsLando: **[Minor] · DRY**

Both this pre-CONNACK branch and the 0x18 re-auth branch (line 81) run `client.handleAuth(packet, (err, packet2) => ...)` and then `_sendPacket(packet2)`; only the error sink differs.

**Why:** Two copies of the enhanced-auth continuation that have *already* drifted — the re-auth branch gained the `!packet2` guard this one still lacks, so the same user hook has two contracts depending on when it is called.

**Fix:** One continuation helper parameterized by the error sink.

src/lib/client.ts
```diff
@@ -1828,6 +1984,52 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 	 * _cleanUp - clean up on connection end
 	 * @api private
 	 */
+	/**
```
robertsLando, line 2171: **[Minor] · Readability**

`_finishReauth` and its JSDoc were inserted between `_cleanUp`'s doc comment (ending line 1986) and `_cleanUp` itself (line 2032).

**Why:** `_cleanUp - clean up on connection end` now documents the wrong function, and `_cleanUp` — older and more consequential — is left with no doc comment.

**Fix:** Move `_finishReauth` and its doc above the `_cleanUp` doc block, or below `_cleanUp`.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+				this._finishReauth(
+					new Error(
+						'reauthenticate: timed out waiting for the broker AUTH response',
+					),
+				)
+			}, this.options.reauthTimeout)
+		}
+
+		this.log('reauthenticate :: sending AUTH packet (reason code 0x19)')
+		this._sendPacket(authPacket, (err) => {
```
robertsLando: **[Minor] · Correctness**

When `writeToStream` returns false (backpressure), `_writePacket` parks the callback on `stream.once('drain', cb)` (`client.ts:2225`), and `_finishReauth` never removes it.

**Why:** One leaked `'drain'` listener per aborted re-authentication, and a malformed AUTH surfaces only via the timeout rather than through this callback.

**Fix:** Remove the parked listener when the exchange finishes.

src/lib/client.ts
```diff
@@ -1682,6 +1721,123 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+	 * @param {String} [options.reasonString] - Reason String
+	 * @param {Object} [options.userProperties] - User Properties
+	 * @param {Function} [callback] - fired when the exchange completes: `(err, packet)`.
+	 *   When omitted, failures are emitted as `'error'` events.
+	 * @returns {MqttClient} this - for chaining
+	 * @api public
+	 * @example client.reauthenticate({ authenticationData: Buffer.from(newToken) }, (err) => {})
+	 */
+	public reauthenticate(
+		options?: ReauthenticateOptions | AuthCallback,
```
robertsLando, line 1790: **[Minor] · Design**

No declared overloads, unlike `publish`/`subscribe`/`unsubscribe`/`end`, which each list explicit ones.

**Why:** The `.d.ts` exposes the raw union, so `reauthenticate(cb, cb)` typechecks and silently discards the second callback — and `reauthenticateAsync(fn as any)` never settles, because the resolve callback is the one overwritten.

**Fix:** Two overload signatures, matching the siblings.

src/lib/client.ts
```diff
@@ -1828,6 +1984,52 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+				callback(err)
+			} else {
+				this.emit('error', err)
+			}
+			return
+		}
+
+		if (callback) {
+			callback(null, packet)
+		}
```
robertsLando, line 2220: **[Minor] · Security**

`'reauth'` fires on a broker-asserted `0x00` with no application verification hook — `handleAuth()` runs only for `0x18`.

**Why:** Mutual-authentication methods cannot validate the server's final message, so rotation and audit hooks treat a forged success as acceptance.

**Fix:** Surface the final packet to `handleAuth()`, or document the event as unverified.

src/lib/client.ts
shirasu0206: Moved the abort into `_cleanUp`, which runs synchronously on both teardown paths, and left the `'close'` handler as a backstop for a peer-driven close. `reauthenticate()` also refuses to start on a destroyed or half-closed stream now. Covered by a test that calls `end()` with `reauthTimeout: 0` against a broker that never answers and never FINs.

src/lib/client.ts
shirasu0206: The properties are validated and built before the pending state is set, and the send is wrapped so a serialization throw still goes through `_finishReauth`. `authenticationData` must be a Buffer or a string; the test asserts nothing reached the wire and that a later exchange still works.

src/lib/handlers/auth.ts
shirasu0206: The received method is compared to the negotiated one before the switch, and a mismatch fails the exchange with 0x8C without reaching `handleAuth()`. I had to make the mock brokers spec compliant for the existing tests to pass, which is some evidence the check bites.

src/lib/handlers/auth.ts
shirasu0206: That one landed in cc2255d, about ten seconds before the review: `should fail the re-authentication on an unexpected AUTH reason code` has the broker echo 0x19. A refusal code other than 0x00/0x18/0x19 cannot be put on the wire because mqtt-packet rejects it, so the reachable case is the 0x19 echo.

src/lib/handlers/auth.ts
shirasu0206: Unsolicited AUTH now goes through `_onProtocolError`: the stream is torn down synchronously and the `'error'` is emitted on the next tick, and only when there is a listener. New test connects with no `'error'` listener and asserts the connection closes.

src/lib/handlers/auth.ts
shirasu0206: Documented as a breaking change in the README rather than keeping the forward. A fallback keyed on whether `handleAuth` was reassigned would make the check conditional on a property comparison, which I did not want for a security check, but I will swap it if you prefer the compatible route.

src/lib/client.ts
shirasu0206: `_finishReauth` now forwards the packet on broker-side failures too, so `reasonString`/`userProperties` reach the callback and the README says so. The internal failures - timeout, close, `handleAuth` error - still pass none, because there is no packet.

src/lib/handlers/auth.ts
shirasu0206: The timer is rearmed on each accepted 0x18 and the exchange is capped at 10 rounds independently of it. A test drives an endless challenge loop with `reauthTimeout: 0` and asserts the cap ends it.

src/lib/client.ts
shirasu0206: Replaced with an allowlist of the three documented properties, each type checked. A test passes `authenticationMethod: 'other'` and asserts the wire packet still carries the CONNECT value.

src/lib/client.ts
shirasu0206: Normalized in the constructor: only an explicit 0 disables the timer, anything non-numeric, non-finite or negative falls back to the default.

src/lib/client.ts
shirasu0206: Kept the connection up - tearing down the connection from the library felt like the application's call - but documented that obligation in the README and added `client.lastReauthError` as the marker.

src/lib/client.ts
shirasu0206: Now `GenericCallback<IAuthPacket>` from `shared.ts`, which already unions the error type, so `.code` reads without a cast.

src/lib/handlers/auth.ts
shirasu0206: Added 24 and 25, changed `0` to `Success`, and added `reasonCodeText()` with a numeric fallback for unknown codes. The pre-CONNACK path uses it, so `Connection refused: Re-authenticate` instead of the empty string; there is a test for that string.

src/lib/handlers/auth.ts
shirasu0206: Added `should fail when the default handleAuth answers a 0x18 challenge`, which leaves `client.handleAuth` at its default.

test/node/client_mqtt5.ts
shirasu0206: All five added, plus two tests for the pre-CONNACK branch. `auth.ts` is at 90% statements / 79% branches now; what is left is the two `_sendPacket` write-error forwards, which I could not reach without stubbing `_writePacket`.

src/lib/client.ts
shirasu0206: Routed through `_checkDisconnecting` with only the `!this.connected` case on top. The disconnecting case now reports `client disconnecting` like the siblings, so I split the existing test in two to cover both states.

src/lib/client.ts
shirasu0206: Extracted `_dispatchError(err, callback)`. `fail()`, `_finishReauth` and `_checkDisconnecting` all use it, so the `!== this.noop` guard now applies to all three.

src/lib/handlers/auth.ts
shirasu0206: Extracted `continueAuth()`, parameterized by the error sink. The pre-CONNACK branch picks up the `!packet2` guard as a result, and it now rejects a non-0x18 code before calling `handleAuth()` rather than after - previously the hook ran and its answer was discarded.

src/lib/client.ts
shirasu0206: Moved `_finishReauth` below `_cleanUp`'s doc comment, so that comment documents `_cleanUp` again.

src/lib/client.ts
shirasu0206: The send callback is held on `_reauthSendCb` and removed from `'drain'` in `_finishReauth`.

src/lib/client.ts
shirasu0206: Added the two overloads, matching `publish`/`subscribe`.

src/lib/client.ts
shirasu0206: Documented the event in the README as the broker's assertion, unverifiable by the application. Routing the final 0x00 through `handleAuth()` would change what that hook means for callers who only expect challenges, so I left it, but I will add it if you want the verification point.

shirasu0206:
Pushed 78a84ad. All four blockers and every Major are addressed, plus most of the Minors and nits; replies are on the individual threads.

On the non-anchored points: the README now covers the single-exchange-in-flight rule, that `authenticationMethod` is inherited from CONNECT and cannot be overridden, `err.code`, the chainable return, and that `authenticationData` is optional. The `RC_*` constants are exported and `client.ts` uses `RC_REAUTHENTICATE` instead of the bare `0x19`. The AUTH logs are hex now, and `basePort + 4` is in use.

Not done, deliberately:

- `_writePacket` logging `packet: %O`. Redacting `authenticationData` there is a change to a shared debug line that also covers CONNECT passwords, so it belongs in its own PR rather than hidden in this one. Happy to open it.
- Threading `clientId` through the reauth logs, splitting `handleAuth` further than `continueAuth` already does, the `reauthenticate: ` error prefix, the `ReauthenticateOptions` name, and the `'reauth'`/`reauthenticate()`/`reauthTimeout` spelling. The last one I would rather leave: `'reauth'` matches `reauthTimeout`, and renaming either after the fact is worse than the mismatch. Say which of these you want and I will fold them in.
- The `process.nextTick` assertions and the `setTimeout(..., 50)` in the older tests. The new unsolicited-AUTH test drives off the server callback with no timer; I left the existing ones alone to keep this diff to the review.

Test run here on Node 22: 944 tests, 939 pass, 0 fail, 5 skipped, coverage gate green. `npm run lint`, `prettier --check` and `tsc -p tsconfig.build.json` are clean. The repo's own workflows still need a maintainer approval to run on this PR.

One thing worth a decision from you: the unsolicited-AUTH change is a behaviour break for anyone using broker-initiated enhanced auth mid-session. I documented it as breaking rather than adding a compatibility fallback, on the grounds that the fallback would make the check depend on whether `handleAuth` was reassigned. If you would rather ship this non-breaking, I will add the fallback instead.

shirasu0206:
Two more commits, no behaviour change to what you reviewed.

Codecov flagged 20 uncovered lines in the patch. Ten of them were reachable and now have tests: the `reasonString` and `userProperties` validation in `_buildReauthProperties`, the `try/catch` around the initial `_sendPacket` (mqtt-packet dereferences every user property value while it measures the packet, so `{ userProperties: { x: null } }` throws out of `_writePacket` before a byte is written - the test asserts the pending state is released and a later exchange still runs), and a `handleAuth()` failure during the pre-CONNACK exchange. The four that remain are defensive: `_sendPacket` only calls back with an error via `_applyTopicAlias`, which is publish-only, and the `_reauthPending` guard in `_finishReauth` is already enforced by every caller.

The other commit is a type fix I should have caught earlier: `ReauthenticateOptions` picked `authenticationData` from `IAuthPacket['properties']`, where it is `Buffer`, so the string form that `reauthenticate()` accepts and the README documents did not typecheck. `tsc -p tsconfig.json` covers the tests and was failing on my own test file; `npm run build` excludes `test/`, which is why CI stayed green.

948 tests, 0 failures, lint and prettier clean.

robertsLando:
Going with this one over #2048 — reasoning is in the note I left there. Thanks for offering to defer; no need.

Before the next round, a heads-up: `main` moved a lot since you branched. Six security fixes landed (#2060–#2065), and `b511e7b` (GHSA-fwrw-4mhv-wxvm) rewrote `src/lib/handlers/auth.ts` over much of the same ground. So the rebase is mostly *removing* your broker-facing hardening, not reconciling it.

**Already on `main` — drop your versions:**

- Unsolicited AUTH as a protocol error that tears the connection down. `rejectAuth()` handles it with `_cleanUp(true)` plus `pump.discardParsedPackets()`, so `_onProtocolError` isn't needed.
- The authentication-method check. Note main compares against `client['_authenticationMethod']` — the value actually sent in CONNECT — not `options.properties?.authenticationMethod`. Please use main's: a caller mutating `options` mid-exchange must not be able to move the goalposts.
- MQTT-4.12.0-6: an AUTH when no method was offered in CONNECT is rejected. No equivalent in your branch.
- The "`handleAuth()` provided no packet" guard.
- Rejecting reason codes other than `0x00` / `0x18`.

**Keep, rebased onto main's handler:**

- `reauthenticate()` / `reauthenticateAsync()`, `reauthTimeout`, the `'reauth'` event, `_finishReauth`, `lastReauthError`.
- The `0x19` send path and the `0x18` continuation, hooked into main's `handleAuth` callback rather than replacing it.
- `_startReauthRound()` round cap and the timer rearm on each accepted `0x18`.
- The `_sendPacket` error-callback wiring in `continueAuth`.
- The `ack.ts` additions (`reasonCodeText`, codes 24/25) — main still has `0: ''` and no helper.
- The tests, including the pre-CONNACK coverage.

`handleAuth` on main now takes a third `pump?: PacketPump` argument. The re-auth path is post-CONNACK so it mostly won't need it, but the signature has to line up.

Two smaller things:

- The breaking-change note in the description and README is moot now — main already made an unsolicited AUTH fatal. Please reword it to document current behaviour rather than introduce it.
- On the naming points I raised, your call stands; I'm not going to hold the PR on `'reauth'` vs `'reauthenticate'`. Go ahead and open the separate PR for redacting `authenticationData` in the `_writePacket` debug line — agreed it doesn't belong here.

I'll do a full pass once it's rebased and CI is green.

src/lib/client.ts
robertsLando: Still open as of this review. `_finishReauth` does receive `packet` now and both `auth.ts` failure sites pass it, but the error branch calls `_dispatchError(err, callback)` — and `_dispatchError` (`client.ts:2034`) has no packet parameter at all, it just calls `callback(err)`. So `packet` is dropped on every failure path, while README:827-829 and the `_finishReauth` JSDoc both promise it is delivered.

src/lib/handlers/auth.ts
```diff
@@ -1,6 +1,53 @@
...
+					new ErrorWithReasonCode(
+						'Authentication failed: client.handleAuth() did not provide an AUTH packet to continue the exchange',
+						rc,
+					),
+				)
+				return
+			}
+
+			client.log('handleAuth :: sending next AUTH packet')
+			client['_sendPacket'](packet2, (sendErr) => {
```
robertsLando, line 42: **[Blocker] · Correctness + Security**

`continueAuth`'s `handleAuth()` callback sends `packet2` without checking that the exchange is still alive.

**Why:** App authenticators do KDFs and token fetches. If `reauthTimeout` fires or `_abortReauth` runs first, `_reauthPending` is already false and the late send still goes out — and `_sendPacket` lets `cmd:'auth'` bypass the `!connected` guard straight into `_writePacket`, so it writes onto a destroyed socket or, post-reconnect, onto the *new* stream mid-handshake. The broker's reply then lands on the unsolicited-AUTH branch and `_onProtocolError` tears down a healthy connection. If a second `reauthenticate()` has started, the stale `onError` resolves the new exchange's callback with the old error.

**Fix:** Give each exchange an identity (a monotonic token captured at call time) and drop continuations — and their send callbacks — that no longer match.

src/lib/client.ts
```diff
@@ -1824,11 +2168,68 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+
+		const callback = this._reauthCallback
+		this._reauthPending = false
+		this._reauthCallback = null
+		this._reauthRounds = 0
+
+		if (err) {
+			this.lastReauthError = err
+			this._dispatchError(err, callback)
```
robertsLando, line 2213: **[Blocker] · Operability + Correctness**

The error path calls `_dispatchError`, which falls back to `emit('error')` when `reauthenticate()` had no callback — and it throws from three places that cannot absorb it.

**Why:** With no `'error'` listener, `emit` throws. `_cleanUp` aborts at its first statement so `stream.destroy()` never runs; the `close` handler skips `_destroyKeepaliveManager()` and `_setupReconnect()`; and `handlers/index.ts:52-56` calls `handleAuth(...)` then `done()` with no `try/finally`, while `work()` has no `try` either, so `completeParse` never fires and the parse pipeline wedges with `connected` still true. A plain `end()` throws and leaks the socket; the client stops reconnecting. The same hole swallows a *synchronous* throw from a user's `handleAuth()`, on both the pre-CONNACK and the re-auth path.

**Fix:** Route reauth outcomes through the same guarded, deferred dispatch `_onProtocolError` already uses twenty lines up (`listenerCount` + `setImmediate`), and/or make the handler dispatcher call `done()` in a `finally`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +71,102 @@ const handleAuth: PacketHandler = (
...
 	}
 
-	client.handleAuth(
-		packet,
-		(err: ErrorWithReasonCode, packet2: IAuthPacket) => {
-			if (err) {
-				client.emit('error', err)
-				return
-			}
+	if (!client.connected) {
```
robertsLando, line 74: **[Blocker] · Security**

The pre-CONNACK branch never compares `packet.properties.authenticationMethod` to the negotiated one — the [MQTT-4.12.0-5] check at lines 112-131 sits *below* the `_reauthPending` guard at line 99, so it only ever runs for re-auth.

**Why:** A hostile or MITM broker answers a `SCRAM-SHA-512` CONNECT with AUTH 0x18 carrying `authenticationMethod: "PLAIN"`. `client.handleAuth()` is application code that dispatches on that field, so the client hands over a plaintext credential — on the exchange that carries the *initial* secret. README:824-826 states the rejection applies "in either direction", so this is a documented guarantee that does not hold.

**Fix:** Hoist the comparison above the `client.connected` branch so both exchanges share it; on a pre-CONNACK mismatch, fail the connection with 0x8c instead of calling `handleAuth()`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +71,102 @@ const handleAuth: PacketHandler = (
...
 					rc,
-				)
-				client.emit('error', error)
+				),
+			)
+			return
+		}
+
+		client.reconnecting = false
+		continueAuth(client, packet, rc, (err) => {
```
robertsLando, line 90: **[Major] · Security**

No round cap on the pre-CONNACK exchange — `MAX_REAUTH_ROUNDS` is enforced only via `_startReauthRound()` in the re-auth branch.

**Why:** A broker can drive unbounded AUTH 0x18 rounds inside the `connectTimeout` window, each running application KDF work, and repeat it on every automatic reconnect. The PR's own rationale for the cap applies identically here.

**Fix:** Count rounds inside `continueAuth` so both branches are bounded.

src/lib/client.ts
```diff
@@ -1756,17 +2026,91 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
-			}
+			this._dispatchError(new Error('client disconnecting'), callback)
 		}
 		return this.disconnecting
 	}
 
+	/**
+	 * A protocol violation by the broker is fatal to the connection.
+	 *
+	 * This runs on the packet parsing path, where an `'error'` event with no
```
robertsLando, line 2052: **[Major] · Design + Operability + DRY**

`_onProtocolError` destroys via `_cleanUp(true)` — which falls through to `_setupReconnect()` — sends no DISCONNECT 0x82, and emits nothing when there is no `'error'` listener.

**Why:** A non-conforming broker produces a silent connect -> AUTH -> destroy -> reconnect loop whose only trace is `DEBUG=mqttjs:client`; `close` carries no reason. Without the DISCONNECT, "protocol error 0x82" is client-local fiction. The repo already has a protocol-violation convention at `handlers/index.ts:17-22` (emit, then `end({reasonCode, properties})`); this adds a second, divergent one.

**Fix:** Reuse the `end({reasonCode})` shape, make the failure observable after the fact, and decide deliberately whether a protocol error should re-arm reconnect.

src/lib/client.ts
```diff
@@ -1824,11 +2168,68 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+		this._reauthCallback = null
+		this._reauthRounds = 0
+
+		if (err) {
+			this.lastReauthError = err
+			this._dispatchError(err, callback)
+			return
+		}
+
+		this.lastReauthError = null
```
robertsLando, line 2217: **[Major] · Design + Security + Operability**

`lastReauthError` is cleared only in the constructor and on success.

**Why:** An exchange aborted by a disconnect leaves it set. After a clean reconnect that re-ran full CONNECT authentication the client is healthy but still reports a failure — the inverse of the documented "tells a healthy client from one running on refused credentials", and a false positive for anything monitoring it.

**Fix:** Clear it on a successful connect, or drop the public field and let the error flow only through the callback and `'error'`.

src/lib/handlers/auth.ts
```diff
@@ -1,6 +1,53 @@
...
+					),
+				)
+				return
+			}
+
+			client.log('handleAuth :: sending next AUTH packet')
+			client['_sendPacket'](packet2, (sendErr) => {
+				if (sendErr) {
+					client.log('handleAuth :: failed to send next AUTH packet')
+					onError(sendErr)
```
robertsLando, line 45: **[Major] · Correctness**

The continuation's `_sendPacket` callback is never recorded in `_reauthSendCb`.

**Why:** On a backpressured stream `_writePacket` parks it on `'drain'`, and `_finishReauth` only unhooks the *first* AUTH's callback — up to 10 listeners leak per exchange (masked by `setMaxListeners(1000)`) and can fire after the exchange has ended.

**Fix:** Track every in-flight send callback of the exchange and clear the list in `_finishReauth`.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +71,102 @@ const handleAuth: PacketHandler = (
...
+	// cannot steer client.handleAuth() onto a method the client never agreed to.
+	const negotiated = options.properties?.authenticationMethod
+	const received = packet.properties?.authenticationMethod
+	if (received !== negotiated) {
+		client.log(
+			'handleAuth :: authentication method mismatch, expected %s got %o',
+			negotiated,
+			received,
+		)
+		client['_finishReauth'](
```
robertsLando, line 123: **[Major] · Security**

A bad authentication method and an illegal broker-sent 0x19 are soft failures: the session keeps running on the old credentials with only `lastReauthError` set, while an unsolicited AUTH at line 99 is fatal.

**Why:** Inconsistent threat handling — the two soft cases are the ones that actually signal an active downgrade or spoofing attempt, yet the suspect connection stays up and keeps publishing.

**Fix:** Route both through `_onProtocolError` (spec: close with 0x8c / 0x82), or document why staying connected is the safe choice.

src/lib/handlers/auth.ts
```diff
@@ -19,26 +71,102 @@ const handleAuth: PacketHandler = (
...
+			),
+			packet,
+		)
+		return
+	}
+
+	switch (rc) {
+		case RC_SUCCESS:
+			client.log('handleAuth :: re-authentication succeeded')
+			client['_finishReauth'](null, packet)
```
robertsLando, line 136: **[Major] · Design**

The final `0x00` bypasses `handleAuth()`, so `'reauth'` and the success callback fire before the application can verify the server's final `authenticationData`.

**Why:** Mutual-authentication methods cannot reject a forged success. The README acknowledges the gap, which freezes it into the event contract.

**Fix:** Route `0x00` through `handleAuth()` or a verify hook, and signal success only if the application does not error.

src/lib/handlers/ack.ts
```diff
@@ -48,6 +50,13 @@ export const ReasonCodes = {
 	162: 'Wildcard Subscriptions not supported',
 }
 
+/**
+ * Human readable text for a reason code, falling back to the numeric value so
+ * an unlisted code still names itself in the error message.
+ */
+export const reasonCodeText = (code: number): string =>
```
robertsLando, line 57: **[Major] · DRY + Operability**

`reasonCodeText()` is added but wired into only the one new call site; `ack.ts:95`, `:117`, `:135` and `connack.ts:48` still index the table raw.

**Why:** Those sites still emit `Publish error: undefined` / `Connection refused: undefined` for codes missing from the table — the exact gap this helper exists to close — and the repo now has two ways to render a reason code.

**Fix:** Route every `ReasonCodes[rc]` interpolation through the helper in this PR.

src/lib/handlers/ack.ts
```diff
@@ -3,14 +3,16 @@
 import { type PacketHandler, ErrorWithReasonCode } from '../shared'
 
 export const ReasonCodes = {
-	0: '',
+	0: 'Success',
```
robertsLando, line 6: **[Minor] · Design + DRY**

`ReasonCodes[0]` `''` -> `'Success'` mutates a map that is re-exported publicly from `src/mqtt.ts:28`, to serve one new caller.

**Why:** No internal caller reaches `rc === 0` (connack guards `rc > 0`), so this is API-visible only — but a consumer doing `if (ReasonCodes[code])` as an is-this-an-error test flips behaviour, and unrelated messages can now read `Subscribe error: Success`.

**Fix:** Leave the shared table alone and let `reasonCodeText` distinguish "present but empty" from "absent".

src/lib/client.ts
```diff
@@ -543,6 +597,19 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			}
 		}
 
+		// A non-finite or negative timeout would silently disable the abort
+		// timer, turning every failed re-authentication into a permanent hang.
+		// Only an explicit 0 disables it.
+		const { reauthTimeout } = this.options
+		if (
+			reauthTimeout !== 0 &&
```
robertsLando, line 605: **[Minor] · Correctness**

Validation accepts any finite non-negative number, but `setTimeout` clamps values above 2^31-1 to 1ms.

**Why:** `reauthTimeout: 1e12` aborts the exchange almost immediately with a "timed out" error — the opposite of the stated intent, and the one case this validation block exists to prevent. The coercion to the default is also silent, with no log line.

**Fix:** Cap at `2147483647` (or reject) alongside the existing checks, and log when a supplied value is replaced.

src/lib/client.ts
```diff
@@ -707,6 +781,10 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			this.log('close :: clearing connackTimer')
 			clearTimeout(this.connackTimer)
 
+			// Backstop only: `_cleanUp` already aborts on the teardown paths
+			// the client controls. This catches a close driven by the peer.
+			this._abortReauth('connection closed before the broker answered')
```
robertsLando, line 786: **[Minor] · Operability**

The abort reason is byte-identical for a user-initiated `end()` and a peer-driven drop: "connection closed before the broker answered".

**Why:** At 3am those are different incidents, and the log cannot tell them apart. The abort error and the `_finishReauth` log also carry no round number, configured timeout, or authentication method, so "timed out waiting for the broker AUTH response" cannot distinguish "broker never replied" from "died after 3 challenges".

**Fix:** Pass the actual cause as the reason, and include `_reauthRounds` and the timeout in the error and the log.

src/lib/client.ts
```diff
@@ -1824,11 +2168,68 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+	}
+
 	/**
 	 * _cleanUp - clean up on connection end
 	 * @api private
 	 */
 	private _cleanUp(forced: boolean, done?: DoneCallback, opts = {}) {
+		// Synchronous teardown entry point: a graceful end() only half-closes
+		// the socket, so waiting for 'close' can strand the exchange forever.
+		this._abortReauth('connection closed before the broker answered')
```
robertsLando, line 2231: **[Minor] · Correctness**

`_reauthPending`, `_reauthRounds` and `lastReauthError` are reset only in the constructor, never in `connect()`.

**Why:** If any teardown path ever misses the abort, a stale `true` permanently blocks `reauthenticate()` with "already in progress" and misroutes a post-CONNACK AUTH away from `_onProtocolError`.

**Fix:** Reset the reauth block where `connect()` resets the other per-connection state.

robertsLando (commented):
## Deep review — MQTT 5 re-authentication

**Verdict: needs work.** Three blockers, all in the new exchange's *lifecycle* rather than its happy path. The feature itself is the right shape and belongs here.

**Strengths, first:** the test suite is the real thing — real server round-trips, no fake timers, assertions traced to the actual reason codes and messages, and it covers the pre-CONNACK enhanced-auth flow that had zero coverage before this PR. 21 of the 22 open review threads are genuinely fixed in code (I've resolved those). The comments that exist explain *why* rather than *what* (the sync abort in `_cleanUp`, the drain-listener removal, the serialization-throw recovery), and the spec citations check out.

### Top 3 risks

1. A slow `handleAuth()` that returns after its exchange ended still sends an AUTH; the broker's reply then trips the new unsolicited-AUTH kill switch and tears down a *healthy* connection.
2. Any error dispatch out of `_finishReauth` can throw — from inside `_cleanUp()`, from the `close` handler, and from the parser's packet loop. Socket never destroyed, reconnect never armed, parser wedged.
3. The pre-CONNACK exchange — where the *initial* credential lives — got neither of the two guards the re-auth exchange got, while the README promises the method check applies "in either direction".

### Themes

- **The exchange has no identity.** Nothing stamps an exchange, so late continuations, late send callbacks and late aborts all act on whatever exchange happens to be current.
- **`_finishReauth` dispatches errors from places that must not throw.** `_onProtocolError` already solves exactly this, twenty lines away; the reauth path doesn't reuse it.
- **Guards were added to the re-auth branch only.** Method validation and the round cap both sit below the `_reauthPending` check, so the pre-CONNACK branch is untouched.
- **A documented `(err, packet)` contract the code cannot honour.**

### Carry-over from the previous round

One thread is still genuinely open — `_finishReauth` does receive `packet` and both `auth.ts` failure sites pass it, but the error branch calls `_dispatchError(err, callback)`, and `_dispatchError` (`client.ts:2034`) has no packet parameter at all. `packet` is dropped on every failure path, while README:827-829 and the `_finishReauth` JSDoc both promise it is delivered. Replied on that thread rather than opening a duplicate here.

### Not line-anchorable

- **Release semantics of the breaking change.** Post-CONNACK AUTH goes from "forwarded to `handleAuth()`" to "connection destroyed", with no opt-out, in a 5.x minor — and overriding `handleAuth()` no longer helps, because the packet never reaches it. Spec-correctness isn't in question; the escape hatch is. Worth deciding deliberately: option-gated now with the new behaviour defaulted in 6.0, or hold it for 6.0. Related and not covered by the breaking-change note: the pre-CONNACK path also changed — `handleAuth()` is no longer invoked for non-0x18 codes.
- **No test makes `handleAuth` throw synchronously.** Every override in the new tests fails via the callback. Given risk 2, that is precisely the untested path that crashes the process or strands `_reauthPending` / `reauthTimer`.
- **`_writePacket` logs `packet: %O`**, so `properties.authenticationData` lands in debug output, and `packetsend` carries it too. Pre-existing, but `reauthenticate()` now routes token material through there routinely.
- **Callback/event ordering:** the callback fires before `emit('reauth')`, while `end()` emits `'end'` then calls back — worth picking one. And a callback-less failure surfaces on `'error'`, which existing applications treat as connection-fatal, though the connection deliberately stays up.
- **Structure:** `handleAuth` runs two structurally different jobs sharing only `continueAuth`. Splitting on `client.connected` into two named handlers would make both testable in isolation — and would have made blocker 3 visible.
- **Nits:** the "did not provide an AUTH packet" error carries `rc`, which on the reauth path is `0x18`, a non-error code; `RC_*` constants live in `handlers/auth.ts` and are imported by `client.ts`, inverting the layering (`shared.ts` is the natural home); naming drift (`ReauthenticateOptions`/`AuthCallback` vs `IClientPublishOptions`/`ClientSubscribeCallback`, and three spellings in one feature); `reauthenticate(undefined, cb)` doesn't type-check; `authenticationData as Buffer` casts a value the JSDoc says may be a string; only the first `reauthenticate ::` log line carries `clientId`.

### Coverage

Reviewed by seven parallel specialists — correctness, security, DRY/codebase fit, design & back-compat, tests, operability, readability — against a worktree at `6c8d83e`. Performance was not run: no files in its lane.

test/node/client_mqtt5.ts
```diff
@@ -1438,4 +1438,1039 @@ describe('MQTT 5.0', () => {
...
+		const authMethod = 'TEST-AUTH'
+		const basePort = ports.PORTAND327 + 30
```
robertsLando, line 1443: **[Minor] · DRY + Readability + Tests**

`const basePort = ports.PORTAND327 + 30` is the only port arithmetic in the suite, and the offsets are assigned out of file order (0,1,2,3,5,6,4,...).

**Why:** It claims ~30 ports `test/node/helpers/port_list.ts` does not know are taken (it stays under `PORTAND400`, so no collision today), and there is no name to grep for — reordering a test risks a silent reuse. Separately, nothing asserts `reauthTimer === null` after success, nothing exercises stream backpressure so the drain-listener removal never actually runs, and the round-cap test does not assert the exact AUTH count, so an off-by-one at the boundary would pass.

**Fix:** Add named entries to `getPorts`, and add the three assertions above.

PR #2058, opened 2026-09-08
fix: handle reserved topic names in subscription and topic alias maps

vpfkfl753:
Several topic names were mishandled because topic-keyed maps were plain objects, or because an internal flag was smuggled through one of them as a key. Each one silently breaks a subscription, a publish, or a connect.

### `__proto__` subscriptions are lost on reconnect

`_resubscribeTopics[topic] = opts` with `topic === '__proto__'` reassigns the map's prototype instead of adding an own key, so `Object.keys()` in `_resubscribe()` never sees it. The subscription maps are now created with a null prototype.

### A topic named `resubscribe` is lost on reconnect, permanently

`_resubscribe()` set `resubscribeTopic.resubscribe = true` to force a re-send, and `subscribe()` then did `delete topicObject.resubscribe`. When a real topic is named `resubscribe`, the flag overwrites its stored options and the `delete` removes the topic itself: `Object.keys()` returns `[]`, no SUBSCRIBE is sent, and the entry is gone from `_resubscribeTopics` for good.

The flag is now a parameter of a private `_subscribe()` rather than a map key. `subscribe()` is unchanged for callers — it just delegates.

### Topic aliases inherit `Object.prototype`

`TopicAliasSend.topicToAlias` was a plain object, so `getAliasByTopic('constructor')` returned the `Object` function and the `typeof alias !== 'undefined'` guard let it through:

```js
const t = new TopicAliasSend(3)
t.getAliasByTopic('constructor')  // -> [Function: Object]
t.put('__proto__', 1)             // -> true, but stores nothing
```

With MQTT 5, a broker advertising `topicAliasMaximum`, and `autoUseTopicAlias: true`, `client.publish('constructor', payload)` then blanked `packet.topic` and set `properties.topicAlias` to a function — mqtt-packet cannot serialize that, so the publish never reached the broker.

### `connect()` picks a stream builder off `Object.prototype`

Same class, found while reviewing the above. The protocol map is keyed by the scheme from the broker URL with no allowlist, so `if (!protocols[opts.protocol])` is truthy for inherited members and the unknown-protocol fallback is skipped:

```js
mqtt.connect('constructor://localhost:1883')
// TypeError: this.stream.pipe is not a function
```

URL schemes are lowercased, so `constructor` is the only one reachable that way — but an explicit `{protocol: 'toString'}` keeps its casing and fails the same way. Reachable from any app that takes a broker URL from config or service discovery.

### A malformed subscription map killed the connection

Pre-existing, and worth closing here because the `resubscribe` change above makes it newly reachable. A map entry without a usable qos reached mqtt-packet as `qos: undefined`, which destroys the stream:

```js
client.subscribe({ 'a/b': {} })   // connection dropped, cb err = "Connection closed"
client.subscribe({ 'a/b': 1 })    // same
```

Both now fail through the callback with `Invalid qos for topic a/b` and leave the connection up. `{ 'a/b': undefined }` still inherits the call's `opts`, as before.

### Tests

- `test/node/abstract_client.ts`: reconnect regression over `['__proto__', 'resubscribe']` × `[reconnect, resubscribe-disabled-then-re-enabled]`, for MQTT 3.1.1 and 5, plus unsubscribing the reserved topic and unsubscribing its neighbour. Listeners are detached on failure so a broken run doesn't leak onto the shared server.
- `test/node/topic-alias.ts` (new): unit coverage for both alias managers — reserved names, alias reuse, LRU ordering, the recv counter. Kept as unit tests on purpose: the integration version of the alias bug doesn't assert, it hangs, because the broken path produces no packet at all.
- `test/node/mqtt.ts`: protocol fallback for both the URL-scheme and the explicit-option paths.

Each test was confirmed to fail without its fix.

### Notes for maintainers

- `resubscribe?: boolean` was removed from the exported `ISubscriptionMap`. It was internal plumbing, never documented as a subscribe option, and it is exactly what made a topic named `resubscribe` unsubscribable. A caller who passed it now gets a clear `Invalid qos` error through the callback instead of a dropped connection, but it is a behaviour change worth a release note.
- `_resubscribe()` now calls the private `_subscribe()`, so a subclass override or monkey-patch of `client.subscribe` no longer runs on reconnect resubscribes. This is deliberate: `_resubscribeTopics` already holds final topic names, so re-running a prefixing wrapper would double-prefix.
- `TopicAliasRecv` keeps a plain-object map — its keys are broker aliases, range-checked in `put()` — and now says so in a comment. Its `length` is tracked incrementally instead of recomputed with `Object.keys()` on every aliased inbound PUBLISH.

Verified with `npm run lint`, `tsc --noEmit` on both `tsconfig.json` and `tsconfig.build.json`, and `npm run test:node` (975 tests, 970 passed, 5 skipped).


robertsLando:
Thanks for this — I confirmed the bug and the fix. I cherry-picked your test onto `main` and it fails there (`expected [ 'hello' ] to deeply equal [ '__proto__', 'hello' ]`), and passes on this branch. Good catch.

While reviewing I found two more instances of the same class of bug, so I pushed a commit to your branch (c7b9294) rather than open a separate PR — they're the same defect and it's easier to reason about them together. I also updated the PR title and description. Summary of what I added:

**1. A topic literally named `resubscribe` is lost on reconnect, permanently.**

`Object.create(null)` fixes the `__proto__` collision but not this one, on the very same map. `_resubscribe()` set `resubscribeTopic.resubscribe = true` to force a re-send, and `subscribe()` then did `delete topicObject.resubscribe`. If a real topic is named `resubscribe`, that flag overwrites its stored options and the `delete` removes the topic entry itself — `Object.keys()` returns `[]`, `subs.length === 0`, no SUBSCRIBE goes out, and the entry is gone from `_resubscribeTopics` for good.

Swapping `'__proto__'` for `'resubscribe'` in your test reproduces it: `expected [ 'hello' ] to deeply equal [ 'resubscribe', 'hello' ]`.

The fix is to stop smuggling the flag through the topic map: the body of `subscribe()` moved to a private `_subscribe()` that takes `resubscribe` as an argument, and `subscribe()` delegates to it. No change for callers. `resubscribe?: boolean` is gone from the exported `ISubscriptionMap` type — it was internal plumbing, never documented as a subscribe option.

**2. `TopicAliasSend.topicToAlias` has the same plain-object problem.**

```js
const t = new TopicAliasSend(3)
t.getAliasByTopic('constructor')  // -> [Function: Object]
t.getAliasByTopic('toString')     // -> [Function: toString]
t.put('__proto__', 1)             // -> true, but stores nothing
```

The `typeof alias !== 'undefined'` guard in `getAliasByTopic` passes for inherited members. With MQTT 5, a broker advertising `topicAliasMaximum`, and `autoUseTopicAlias: true`, `client.publish('constructor', payload)` hits the truthy branch in `_applyTopicAlias`, blanks `packet.topic` and sets `properties.topicAlias` to a function — mqtt-packet can't serialize that, so the publish never reaches the broker. `Object.create(null)` in the constructor and in `clear()`.

**Tests**

- Your reconnect regression now runs over `['__proto__', 'resubscribe']` × `[reconnect, reset-then-resubscribe]`, so your `resetSubscriptions` case is kept and applies to both topic names.
- New `test/node/topic-alias-send.ts` with unit coverage for the alias map. I deliberately did *not* write this one as an integration test: the broken path produces no packet at all, so the test doesn't assert, it hangs for the full timeout and then poisons neighbouring tests that reuse the same port.

Each test was confirmed to fail without its corresponding fix. `npm run lint`, `tsc -p tsconfig.build.json --noEmit` and `npm run test:node` are all clean (936 passed, 5 skipped).

Let me know if you'd rather I split the extra fixes into their own PR — happy to.

robertsLando:
Ran a deeper review over the branch and pushed the fixes in d93165d. Description updated. Summary of what came out of it:

**Two things the previous state got wrong**

1. **`tsc` was broken.** `events.once(client, 'connect')` doesn't typecheck — `TypedEventEmitter` declares no `addListener`, which `events.once` requires. `main` is at 0 errors, this branch was at 2 (TS2769). It stayed green because CI runs only `lint` + `test:node`, and `esbuild-register` strips types without checking them. Fixed by declaring `addListener` on the interface — `EventEmitter` really does have it via the mixin, the interface just under-declared it. Back to 0 on both `tsconfig.json` and `tsconfig.build.json`.

2. **Dropping `resubscribe` from `ISubscriptionMap` had a bad failure mode.** The removal is correct — it's the bug — but a caller who still passed it went from "works" to "connection dropped":

   ```
   subscribe({'a/b': {qos:1}, resubscribe:true})
   before → SUBSCRIBE [{"topic":"a/b","qos":1}], cb err = null
   after  → connection closed, no SUBSCRIBE, cb err = "Connection closed"
   ```

   The root cause is broader and pre-existing: any map entry without a usable qos reaches mqtt-packet as `qos: undefined`, which destroys the stream. `{'a/b': {}}` and `{'a/b': 1}` do it too, on `main` as well. There's now one guard covering all of them — they fail through the callback with `Invalid qos for topic a/b` and the connection stays up. `{'a/b': undefined}` still inherits the call's `opts`, unchanged.

**One more instance of the same class, outside the original diff**

`connect()`'s protocol map is keyed by the scheme from the broker URL with no allowlist, so `mqtt.connect('constructor://localhost:1883')` skipped the unknown-protocol fallback and called `Object.prototype.constructor` as a stream builder: `TypeError: this.stream.pipe is not a function`. Reachable from any app taking a broker URL from config or discovery. Same one-line fix. URL schemes get lowercased, so `constructor` is the only one reachable that way, but an explicit `{protocol: 'toString'}` keeps its casing and fails identically — both paths are covered now.

I folded it in here rather than opening a separate PR since it's the same bug class and the same fix; happy to split it out if you'd rather keep this PR narrow.

**Smaller cleanups**

- `nullProtoMap()` in `shared.ts` replaces the bare `Object.create(null)` calls — those return `any`, so every assignment site had silently lost its type checking.
- `resubscribe` moved next to the map it applies to in `_subscribe`, so neither call site pads positional `undefined`s.
- `TopicAliasRecv.length` is tracked incrementally instead of recomputed with `Object.keys()` on every aliased inbound PUBLISH. Its map stays a plain object — the keys are range-checked broker aliases — and now says so, so the asymmetry with the send side doesn't read as an oversight.

**Tests**

Reconnect cases now detach their server listeners on failure, and cover unsubscribing the reserved topic as well as unsubscribing its neighbour. Worth flagging: my first version of those unsubscribe tests passed even with the bug present, so they're inverted now — unsubscribe `hello`, require the reserved topic to survive — which does fail without the fix. `test/node/topic-alias-send.ts` is renamed to `topic-alias.ts` since it covers both managers, with alias reuse, LRU ordering and the recv counter added.

Every new test was checked to fail without its corresponding fix. 975 tests, 970 passed, 5 skipped; lint and `tsc` clean.

Two behaviour changes worth a release note, both called out in the description: the `resubscribe` flag removal, and `_resubscribe()` now calling the private `_subscribe()` — which means a monkey-patched `client.subscribe` no longer runs on reconnect resubscribes. That one's deliberate, since `_resubscribeTopics` already holds final topic names and re-running a prefixing wrapper would double-prefix, but it's a change either way.

vpfkfl753:
Thanks for catching the missing `addListener` declaration. Keeping the related map fixes together in this PR is fine with me.

I checked `d93165d` locally: both `tsconfig.json` and `tsconfig.build.json` pass `tsc --noEmit`, lint passes, and `npm run test:node` reports 970 passed and 5 skipped.

src/lib/client.ts
```diff
@@ -1540,6 +1550,20 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			return this
 		}
 
+		// a subscription map carries its options per topic, so a missing or
+		// malformed entry only surfaces here - without this the packet layer
+		// gets `qos: undefined` and destroys the connection
+		const invalidQos = subs.find(
```
robertsLando, line 1556: **[Major] · Correctness / Design / Security / DRY**

This check runs on `subs`, which `parseSub` has already deduped, so a malformed entry for a topic that is *already* subscribed never reaches it. Reproduced against a real broker:

```js
client.subscribe({'a/b': {qos: 1}})   // ok
client.subscribe({'a/b': {}})         // err = null, granted = []
```

The comparison that swallows it is `this._resubscribeTopics[topic].qos < subOptions.qos` → `1 < undefined` → false, so the entry is filtered out before `subs` is built.

**Why:** identical malformed input either errors or silently reports success, depending on prior subscription state *and* on config (`options.resubscribe` / `reconnectPeriod > 0` decide whether `_resubscribeTopics` is populated at all). The caller gets `callback(null, [])` for a subscription that was never sent.

**Fix:** validate where qos is read — inside `parseSub` on `subOptions.qos`, before the already-subscribed/qos-upgrade branch — or up front beside the existing `validations.validateTopics` call. `src/lib/validations.ts` already owns subscribe-argument validation and is called ~90 lines earlier in this same method, so putting it there also removes the second validation idiom.

src/lib/client.ts
```diff
@@ -1540,6 +1550,20 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+			(sub) => sub.qos !== 0 && sub.qos !== 1 && sub.qos !== 2,
+		)
+		if (invalidQos) {
+			setImmediate(
```
robertsLando, line 1560: **[Major] · Operability**

The error reaches only `callback`, which defaults to `this.noop` for the very common `client.subscribe(topicMap)` shape, and there is no `this.log(...)` at this site.

**Why:** previously a malformed qos destroyed the stream — loud and impossible to miss. Now, for a caller who passed no callback, the subscription silently never happens and it does not even appear under `DEBUG=mqttjs*`. That is the "loud failure became silent" case worth avoiding while fixing the crash.

**Fix:** log alongside the `setImmediate(callback, …)`. Separately, `Invalid qos for topic ${invalidQos.topic}` names the topic but not the offending value — for a large subscription map the caller cannot tell whether the qos was `undefined`, a string, or `3`. (Tests pin the current string, so that part needs a coordinated update.)

src/lib/topic-alias-recv.ts
```diff
@@ -11,8 +11,10 @@ export default class TopicAliasRecv {
 	public length: number
 
 	constructor(max: number) {
+		// plain object is safe here: keys are broker aliases, range-checked in put()
```
robertsLando, line 14: **[Major] · Security**

This comment documents an invariant that does not hold. mqtt-packet turns a **repeated** property into an array, so a broker sending `topicAlias` twice in one PUBLISH yields `alias = [1, 2]`, which passes the `put()` guard — `alias === 0` is false and `alias > this.max` is a NaN comparison, also false — and is stored under the string key `"1,2"`. Confirmed:

```js
const alias = [1, 2], max = 3
alias === 0 || alias > max   // false -> guard passes
const m = {}; m[alias] = 't'
Object.keys(m)               // [ '1,2' ]
```

With `properties.topicAliasMaximum: 3`, five such PUBLISHes give `recv.max = 3` but `recv.length = 5` and keys `['1,2','1,3','1,4','1,5','1,6']`.

**Why:** the advertised Topic Alias Maximum is bypassed entirely, so a malicious or compromised broker grows `aliasToTopic` without bound (topic strings as values) — memory-exhaustion DoS at the same trust boundary as GHSA-c8jq-r765-cq7g. The read path is safe (`alias > 0 && alias <= 0xffff` rejects the array), so this is unbounded growth, not topic confusion.

Not flagged as a Blocker because `put()` is unchanged by this PR — the bug is pre-existing. But this PR is an audit of exactly this bug class, and the new comment asserts safety that isn't there.

**Fix:** type-gate the alias at the boundary in `handlers/publish.ts` before `put()`, and make `put()` validate `Number.isInteger(alias)` rather than relying on relational operators. Same treatment for `TopicAliasSend.put`.

src/lib/topic-alias-recv.ts
```diff
@@ -25,8 +27,10 @@ export default class TopicAliasRecv {
 		if (alias === 0 || alias > this.max) {
 			return false
 		}
+		if (!(alias in this.aliasToTopic)) {
```
robertsLando, line 30: **[Minor] · Correctness**

`alias in this.aliasToTopic` walks `Object.prototype` on a prototyped `{}` — the exact pattern this PR removes elsewhere. The comment's "safe here" rests on an invariant only the current caller upholds.

**Why:** a non-numeric alias both skips the `length++` and makes `getTopicByAlias` return an inherited function.

**Fix:** `nullProtoMap<string>()`, or `Object.prototype.hasOwnProperty.call`.

src/lib/client.ts
```diff
@@ -400,8 +401,6 @@ export type ISubscriptionMap = {
 	 * object which has topic names as object keys and as value the options, like {'test1': {qos: 0}, 'test2': {qos: 2}}.
 	 */
 	[topic: string]: IClientSubscribeOptions
```
robertsLando, line 403: **[Major] · Design / BackCompat**

`resubscribe?: boolean` is dropped from the exported `ISubscriptionMap` (re-exported via `src/mqtt.ts`) under a `fix:` commit type.

**Why:** TS consumers passing `{'a': {qos: 0}, resubscribe: true}` now fail to compile. JS consumers change behaviour silently and worse: the key is no longer stripped, so it becomes a topic, hits the new qos check, and the **whole call** fails with `Invalid qos for topic resubscribe` — subscribing *nothing*, where before it subscribed everything with force-resubscribe. It is undocumented in the README and there are no type tests, so nothing catches the drop.

**Fix:** land it with an explicit breaking/behaviour note and a minor bump, or keep accepting-and-stripping the key for one release with a deprecation warning. Worth a release-note line either way — moving invalid qos from connection-destroy to callback error is itself an observable change for anyone whose `'error'` handler currently fires.

src/lib/shared.ts
```diff
@@ -52,6 +52,15 @@ export type PacketHandler = (
 
 export type TimerVariant = 'auto' | 'worker' | 'native'
 
+/**
+ * Empty map for keys that come from the user or the broker. The null prototype
+ * keeps `__proto__` an ordinary key instead of a prototype assignment, and stops
+ * lookups from inheriting `Object.prototype` members like `constructor`.
+ */
+export function nullProtoMap<V>(): Record<string, V> {
```
robertsLando, line 60: **[Minor] · Design**

Two things about making this an export:

1. `src/mqtt.ts` does `export * from './lib/shared'`, so `nullProtoMap` becomes public API — undocumented, and unremovable without a break. An internal prototype-pollution guard becomes something consumers can depend on.
2. The return type `Record<string, V>` claims an ordinary object, but the value has no `hasOwnProperty` and no `toString`: `map.hasOwnProperty(k)` and `\`${map}\`` both throw. The code already knows this internally (`Object.prototype.hasOwnProperty.call` at client.ts:1502) — the exported signature just doesn't say so.

**Fix:** keep it in a module that isn't re-exported, or mark it explicitly internal.

src/lib/topic-alias-send.ts
```diff
@@ -23,7 +24,7 @@ export default class TopicAliasSend {
 	constructor(max: number) {
 		if (max > 0) {
 			this.aliasToTopic = new LRUCache<number, string>({ max })
-			this.topicToAlias = {}
+			this.topicToAlias = nullProtoMap<number>()
```
robertsLando, line 27: **[Minor] · DRY / Codebase fit**

`topicToAlias` is a private, internal, string-keyed lookup sitting two lines below `new LRUCache(...)`, and the repo's existing idiom for keyed lookups is a real map (`src/lib/store.ts:69`, `this._inflights = new Map()`).

**Why:** `Map<string, number>` is `__proto__`-safe by construction — no helper, no comment, no per-call type argument. The helper mostly patches a container choice the codebase already solved elsewhere.

**Fix:** `Map` for the internal maps no caller sees as an object (`topicToAlias`, and `TopicAliasRecv.aliasToTopic`), and keep `nullProtoMap` for the two places where the object shape *is* part of a contract — `_resubscribeTopics` (typed `ISubscriptionMap`) and `connect()`'s `protocols`. That also makes `TopicAliasRecv.length` fall out of `.size` instead of being hand-maintained in two places, matching how the sibling class already derives it.

src/lib/client.ts
```diff
@@ -1430,6 +1429,22 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
 		callback?: ClientSubscribeCallback,
+	): MqttClient {
+		return this._subscribe(topicObject, false, opts, callback)
+	}
+
+	/**
+	 * Implementation of {@link subscribe}. `resubscribe` is internal: when true
+	 * the already-subscribed check is bypassed so reconnects re-send every topic.
+	 */
+	private _subscribe(
```
robertsLando, line 1440: **[Minor] · Design**

TS-`private` here removes resubscribe from any subclass override of `subscribe` — deliberate and correct per the PR body — but it also blocks subclasses from hooking the new seam at all.

**Fix:** `protected` keeps the extension point while still hiding the flag from ordinary callers.

test/node/abstract_client.ts
```diff
@@ -4807,6 +4808,114 @@ export default function abstractTest(server, config, ports) {
...
+						await client.endAsync()
+					}
+				},
+			)
+		}
+
+		it('should reject a subscription map entry without a valid qos', function _test(t, done) {
+			const client = connect()
+			client.on('connect', () => {
+				client.subscribe({ 'a/b': {} } as any, (err) => {
```
robertsLando, line 4909: **[Minor] · Tests**

Two gaps in this case:

1. The PR body names a second malformed shape, `{'a/b': 1}`, which has no test. Traced: `subOptions = (subOptions || opts)`, so a bare number is truthy and kept, then `subOptions.qos` reads `undefined` off the number primitive — same `invalidQos` branch today. Low risk now, but it is the exact case from the bug report, and a future change special-casing object vs. primitive (plausible — some clients treat a bare number as qos shorthand) would regress it silently.
2. Nothing covers the dedup path of the new qos check — only the fresh-topic path, which behaves correctly. The one branch that is wrong is the untested one: subscribe the topic first, then re-subscribe with a malformed entry.

The assertion that `client.connected` stays true is good and worth keeping.

test/node/abstract_client.ts
```diff
@@ -4807,6 +4808,114 @@ export default function abstractTest(server, config, ports) {
...
+		// resolves with the topics of the next SUBSCRIBE the server receives,
+		// detaching its listeners even when the caller never settles
+		const nextSubscribedTopics = (signal: AbortSignal) =>
+			new Promise<string[]>((resolve, reject) => {
+				const onClient = (serverClient) => {
+					serverClient.once('subscribe', (packet) => {
+						resolve(packet.subscriptions.map((sub) => sub.topic))
+					})
+				}
+				server.once('client', onClient)
```
robertsLando, line 4820: **[Minor] · Tests**

Listener detachment is verified for the assertion-failure path — `finally` runs `ac.abort()`, which is correct — but not for a hang. If `nextSubscribe` never settles (broker never re-sends SUBSCRIBE), node:test's `{timeout: 5000}` marks the test failed *without* cancelling the suspended async function, so `finally` never runs, `ac.abort()` never fires, and this `server.once('client', onClient)` stays attached to the shared `server` — which `teardownHelper` doesn't track, since it only knows about clients/servers from the local `connect()`/`serverBuilder()` wrappers.

**Why:** a surviving `.once` listener can steal a `'client'` event meant for a later, unrelated test — cross-test flakiness that is hard to trace back here. Doesn't affect the five bugs under test today (their regressions show up as content mismatches, not hangs).

**Fix:** wire the test's own `t`/signal into `ac` so a hang also triggers cleanup, or add `t.after(() => server.removeListener('client', onClient))` as a backstop.

src/lib/client.ts
```diff
@@ -713,7 +712,7 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
 			options.queueQoSZero === undefined ? true : options.queueQoSZero
 
 		// map of subscribed topics to support reconnection
-		this._resubscribeTopics = {}
+		this._resubscribeTopics = nullProtoMap<IClientSubscribeOptions>()
```
robertsLando, line 715: **[Nit] · DRY**

`nullProtoMap<IClientSubscribeOptions>()` appears three times (here, 2661, 2677) for a field already typed `ISubscriptionMap`. Optional: a small typed factory, or defaulting the generic, removes the repetition.

src/lib/client.ts
```diff
@@ -1430,6 +1429,22 @@ export default class MqttClient extends TypedEventEmitter<MqttClientEventCallbac
...
+		return this._subscribe(topicObject, false, opts, callback)
+	}
+
+	/**
+	 * Implementation of {@link subscribe}. `resubscribe` is internal: when true
+	 * the already-subscribed check is bypassed so reconnects re-send every topic.
+	 */
+	private _subscribe(
+		topicObject: string | string[] | ISubscriptionMap,
+		resubscribe: boolean,
```
robertsLando, line 1442: **[Nit] · Readability**

The internal flag sits second, while `subscribe`/`unsubscribe` keep `opts, callback` last. A trailing flag would read more consistently with the surrounding signatures, and the call sites (`this._subscribe(topicObject, false, ...)`) would need less signature lookup.

robertsLando (commented):
## Review summary

**Verdict:** Needs work — the approach is sound and the five bugs are real, but the new qos check has a reachable branch where it silently reports success, and that branch is untested.

**Top 3 risks**
1. The qos validation runs *after* `parseSub` dedup, so malformed input for an already-subscribed topic returns `callback(null, [])` with nothing subscribed.
2. `TopicAliasRecv`'s new "plain object is safe here" comment documents an invariant that doesn't hold.
3. `resubscribe?: boolean` is dropped from an exported type under a `fix:` commit type.

**Strengths.** Each of the five bugs is traced to a concrete failure rather than asserted. `nullProtoMap` is applied exactly where keys come from users or brokers and deliberately *not* where they're message-id-keyed — that restraint is correct, and I verified the remaining plain objects (`messageIdToTopic`, `outgoing`, `_packetIdsDuringStoreProcessing`, store) are all message-id-keyed. The `TopicAliasRecv.length` change is a genuine O(n)→O(1) win on the inbound-PUBLISH path. The tests trace real code paths and were confirmed to fail on revert.

**Theme.** Two findings share a root cause: validation placed downstream of a filter instead of at the trust boundary (qos), and a safety property asserted in a comment rather than enforced in code (`TopicAliasRecv`).

### Findings outside the diff

These are in files the PR touches but on lines it doesn't change, so they can't be line-anchored.

**Major · Security — `src/lib/connect/index.ts:232-241`.** `opts.servers[i].protocol` never passes the `!protocols[opts.protocol]` fallback at line 190; `wrapper` indexes the map directly. `connect({servers:[{host,port,protocol:'constructor'}]})` now throws `TypeError: protocols[opts.protocol] is not a function`. `servers` is exactly the config/service-discovery input in this PR's threat model, and `wrapper` runs on every reconnect from the `setInterval` in `_setupReconnect`, so a bad entry at `servers[n>0]` is an *uncaught* exception → process crash. Per-server protocols should go through the same resolution as the top-level one.

**Major · Security / Operability — `src/lib/connect/index.ts:190-208`.** Unrecognized schemes silently fall back to the first available protocol — plaintext `mqtt` in Node — with no `debug()` line naming the rejected scheme. The null-proto fix routes *more* config-influenced strings into this branch, so a typo'd `mqttss://user:pass@host` downgrades to cleartext TCP carrying the URL credentials, and the caller gets no signal. An explicit allowlist that throws on unknown schemes would be better than guessing; at minimum, never fall back to an insecure transport, and log the rejection.

### Confirmed sound

Checked and found correct, recorded so it isn't re-litigated: `TypedEmitter.addListener` is declaration-merging only, no runtime change. `TopicAliasRecv.length` stays in sync on every path (fresh put, overwrite, out-of-range reject, clear). `_resubscribe` → `_subscribe` preserves all side-effects — messageId, `messageIdToTopic`, `_resubscribeTopics` writes and `_storeProcessing` all live in the untouched `subscribeChunkedSubs`. Every `delete` / `Object.keys` / `hasOwnProperty.call` consumer is null-proto-safe. No perf cost from `Object.create(null)`: all reads are dynamic bracket accesses, and `topicToAlias` was already in dictionary mode due to `delete`.

The `test/node/mqtt.ts` comment is right as written — `toString`/`valueOf` lowercase to `tostring`/`valueof`, neither on `Object.prototype`, and `__proto__://` is an invalid URL, so `constructor` really is the only URL-reachable one.

### FYI, pre-existing

`TopicAliasSend.put(topic, a)` then `put(topic, b)` orphans alias `a` in `numberAllocator`/`aliasToTopic`; a later `put` recycling `a` deletes the *live* `topicToAlias[topic]` entry. The new tests cover only the reverse direction.

### Coverage

Eight dimensions reviewed against a worktree at the PR head: correctness, security, DRY/codebase-fit, design/API/backcompat, performance, tests, operability, readability. No prior unresolved review threads.

